PR #1, opened 2016-11-15
Make 'to' input fields of type template.

thijskh:
This triggers a more useful keyboard on mobile devices.


jornane:
Looks okay to me. I didn't know about `type="email"` or `multiple` as an attribute on `<input>`, but this does look useful for mobile devices.

PR #2, opened 2016-11-16
Prevent calling session_start() directly

jornane:
FileSender uses a third party authentication handler (often SimpleSamlPhp). It has its own session handler. When combining it with another session system, such as by calling `session_start()`, the sessions conflict.

In FileSender, `session_start()` is called, seeminly for legacy reasons. This has no direct negative effects, because it doesn't actually use this session. However, it does trigger a log message in SimpleSamlPhp. [SimpleSamlPhp 1.14.9 has a bug where this log message causes the application to crash](https://github.com/simplesamlphp/simplesamlphp/issues/517#issuecomment-261258092).

This pull request will remove `session_start()` everywhere it was used. I have tested this on my test instance for multiple weeks, it has so far not caused issues.

### DIFF
--- a/classes/auth/Auth.class.php
+++ b/classes/auth/Auth.class.php
@@ -109,6 +109,10 @@ class Auth {
                 self::$exception = $e;
                 throw $e;
             }
+
+            // If no session has been made at this point, we make one ourselves.
+            // Only types 'guest' and 'sp' are browsers.
+            if(in_array(self::$type, array('sp', 'guest'))) session_start();
             
             if(self::$attributes && array_key_exists('uid', self::$attributes)) {
                 $user_filter = Config::get('auth_user_filter');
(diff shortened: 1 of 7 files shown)


jornane:
[https://github.com/simplesamlphp/simplesamlphp/issues/517#issuecomment-261258092](Referring to Jaime's comment on the SimpleSamlPhp issue tracker). This commit will prevent FileSender from doing it's own session management, thus fixing the misconfiguration issue mentioned in that ticket.

jornane:
Can you tell me why this shouldn't be merged? As you can read on the SimpleSamlPhp link I referred, we should remove `session_start()` from our project. Let me know if you have a better solution.

Raz0rwire:
Terribly sorry, closed one too many. Apparently I had two PR's selected when I closed my own Saucelabs test PR.

PR #10, opened 2016-11-30
WIP: Break Travis-testing (don't accept this one)

jornane:
While I was fixing regressions in #9, I noticed that Travis was returning green checkmarks when it failed to start phpunit due to a bug in the preparation code.

This pull request consist of a commit that reverses my fix for the issue. The goal of this pull request is thus to make Travis return red.

**OF COURSE, THIS REQUEST SHOULD NOT BE MERGED BECAUSE IT RE-INTRODUCES A BUG.**


jornane:
Fails now, like it should. Goodie! :)

Let's close this one.

PR #12, opened 2016-12-01
Make FileSender compatible with PHP 5.3, for backwards compatibility

jornane:
The [] notation is prettier, but Red Hat 6 still uses PHP 5.3 which doesn't support it.
Since FileSender supports all alive versions of Debian and Red Hat, we must limit ourselves to syntaxes supported by old PHP versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 php:
-- 5.4
+- 5.3
 addons:
   sauce_connect:
     username: "$SAUCE_USERNAME"
--- a/classes/data/DBObject.class.php
+++ b/classes/data/DBObject.class.php
@@ -167,17 +167,6 @@ class DBObject {
     }
     
     /**
-     * Default creator
-     * 
-     * Creates empty object
-     * 
-     * @return object instance
-     */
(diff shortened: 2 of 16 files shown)


thijskh:
You should probably add 5.3 here then?
https://github.com/filesender/filesender/blob/master/.travis.yml#L3

Raz0rwire:
You should probably try, is probably the latest version of sausage (yes, sausage) that will run on PHP 5.3:
https://packagist.org/packages/sauce/sausage#0.15.2

jornane:
Can't get SauceLabs 0.15.2 to work with PHP 5.3. The mentioned version of sausage doesn't work on PHP 5.3 due to some strange `use` syntax that I'm not familiar with, but apparently is correct from PHP 5.4.
https://github.com/jlipps/sausage/blob/0.15.2/src/Sauce/Sausage/WebDriverTestCase.php#L6

I've tried downgrading SauceLabs further, from version 0.14.0 it works on PHP 5.3, but then it fails running the actual SauceLabs tests.

jornane:
Forked off the function signatures fix to pull request #14.

PR #14, opened 2016-12-05
Fix code consistency

jornane:
This pull request fixes some issues that I encountered while porting the code to PHP 5.3. The errors I found are not fatal, but may cause trouble anyway. 

The errors are all related to inconsistent function signatures between inheriting classes.


jornane:
Merged when #12 was.

PR #16, opened 2016-12-08
Selenium tests: don't die on error

DraughtGlobe:


Raz0rwire:
@jornane if you merge this PR, and after that run the build for PR #10 again it should fail as expected, turns out you shouldn't use die() in tests.

PR #19, opened 2017-01-11
Replace # in <a href with the actual download link.

### LINKED ISSUE #18: Allow getting cURL-able links
The Download button uses JavaScript to start the download. This is great for encrypted files (these must be decrypted by JavaScript) but not so great for normal files.

Users may want to open the link to the download page on their desktop, but then do the actual download on another server using curl or wget, to avoid having to download a large file to their desktop, only to send it somewhere else.

jornane:
This allows users to right click the Download button and copy the
download link to clipboard, allowing users to download the link in for
example cURL.  This should fix #18.


jornane:
It's not perfect, there are two things missing:
* There is no unit test
* The JavaScript code also constructs the download URL - this is not necessary

However, I do not have the time to fix these issues.

madsi1m:
I think pull 20 works better, shows a link you can copy and paste into external scripts

jornane:
I think it's counter-intuitive that «rightclick -> copy link» on the download button gives an URL, but not one you can paste into external scripts. #20 does not address this.

madsi1m:
yup, #20 just displays a url near the button, the button it self doesnt change

PR #20, opened 2017-01-17
direct link added to download page, php7 fix, User.class.php fix

madsi1m:


madsi1m:
sorry, this pull has two fixes in it:

1. Direct download url
2. session fix for php7+

PR #23, opened 2017-02-13
fix undefined hasOption function in Guest::close

Veredior:
Fix the error 500 when an user or an admin deletes a guest on the pages "Guests" and "Admin/Guests".


madsi1m:
should merge this

PR #24, opened 2017-02-13
fix guests and guests transfers lists requests

Veredior:
Currently on the guests page, the "Guests transfers" table is listing only the transfers when the guest option
"Valid for one upload only" is unchecked.
Consequently, the connected user can download the files only if he goes on the "My Transfers" page.

I changed the FROM_USER constant in order to get the guests only when the expire date is not passed.

However, as it breaks the results from the "Guests" table, i created a new constant and a new function.

I hope it's clear :)


Veredior:
Updating AVAILABLE constant doesn't seem to be necessary as this status is changed automatically when an invite is expired or when the guest has uploaded the files when "Valid for one upload only" is checked.

PR #30, opened 2017-02-24
Preventing spam of PHP notices in logs when we check the existence of a session

Veredior:
The actual check of the existence of a session is not enough as we get in our log files the following notice on all pages : 
`` A session had already been started - ignoring session_start() in /usr/share/filesender/filesender-2.0/classes/auth/Auth.class.php on line 115 ``.


madsi1m:
same fix in pull #20 (with out check for old php)

Issue #29, opened 2017-02-22
Creating Guest Voucher doesn't mail it out to recipient.


sndrsmnk:
Creating a Guest Voucher for an email address doesn't immediately email that voucher to the address provided, unlike the message that appears after clicking Submit would indicate.

The voucher does get mailed to the recipient when the 'Send Reminder'-button is used.

PHP7.0, git master.

madsi1m:
Hi, I had a problem like this where the guest email never sends.
I commented out this line from classes/data/Guest.class.php
if($this->getOption(GuestOptions::EMAIL_GUEST_CREATED))

The idea is that it always runs the next line which does the quickSend()

sndrsmnk:
Thanks for that tip. I hadn't looked into it yet.
Still a bug though :-)

Veredior:
Hi, it does work, just check in your config file in the ``$config['guest_options']`` param for this  : 
``// Notify the recipient when created a guest voucher
	'email_guest_created' => array(
		'available' => true,
		'advanced' => true,
		'default'  => true
	),``
It seems that this option is disabled by default...

sndrsmnk:
I'm quite new to FileSender, so forgive my bluntness but the `available` flag seems to only show/hide the toggle on the webinterface? It isn't entirely clear what the meaning of 'available' is.

Also, the `default` value for `email_guest_created` in `includes/ConfigDefaults.php` is `true`. I would expect it to be the default then, even though the option to switch it off in the webinterface is not `available`.

And also if my understanding of the `available` flag is wrong, i think this is a bug. ;)

The interface states a message is sent while it is not and there's no way to manually figure out the Voucher link though the webinterface.

Veredior:
Agree, some parameters don't have an impact on the UI which likely cause confusion. I will make a PR to fix this issue as this parameter should be enable by default...

madsi1m:
I will uncomment it in my install as well. Good find, should be on by default

jornane:
Seems like there is some confusion here. According to [guest_options](http://docs.filesender.org/v2.0/admin/configuration/#guestoptions), the **sender** of a voucher should get a mail when `email_guest_created` is set, and the **recipient** of a voucher should get a mail when `email_guest_created_receipt` is set.

[...]

PR #31, opened 2017-02-24
activate by default email_guest_created

Veredior:
Actually, no email is sent with the guest voucher after its creation, which is odd due to email_guest_created paramater values.

### DIFF
--- a/includes/ConfigDefaults.php
+++ b/includes/ConfigDefaults.php
@@ -219,7 +219,7 @@ $default = array(
             'default' => false
         ),
         'email_guest_created' => array(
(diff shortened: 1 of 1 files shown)


madsi1m:
good find. should be on by default.

jornane:
I'm not sure if this is the correct solution, as it doesn't seem to match the documentation of the configuration. See my comment on #29 for more information.

monkeyiq:
I am merging https://github.com/filesender/filesender/pull/34 as that seems to bring the default settings forward for things regardless of if the options are presented to the user or not.

PR #32, opened 2017-03-01
Make options label toggle checkboxes unless when user is a guest

Veredior:
Even if the options' checkboxes are disabled on the upload page when a guest access this page, he can interact with the options when clicking the associated label.

This pull request checks if the user is connected.


monkeyiq:
My plan is to merge https://github.com/filesender/filesender/pull/33 once it passes CI. I think using the html label/for to reference ID will be a cleaner solution. Hopefully it will also interact in a sane manner in browsers when the input element is disabled.

Issue #22, opened 2017-02-03
FilesenderRestClient.class.php help?


madsi1m:
Hi All,

We have a user who would like to use the API but is getting "EXCEPTION [0] Http error 500".

I looked in our logs and have this:

> 03-Feb-2017 00:42:17 Europe/London] PHP Fatal error:  Call to a member function exists() on null in /www/filesender-2.0/classes/rest/endpoints/RestEndpointTransfer.class.php on line 355
> [03-Feb-2017 00:42:17 Europe/London] PHP Stack trace:
> [03-Feb-2017 00:42:17 Europe/London] PHP   1. {main}() /www/filesender-2.0/www/rest.php:0
> [03-Feb-2017 00:42:17 Europe/London] PHP   2. RestServer::process() /www/filesender-2.0/www/rest.php:37
> [03-Feb-2017 00:42:17 Europe/London] PHP   3. call_user_func_array:{/www/filesender-2.0/classes/rest/RestServer.class.php:237}() /www/filesender-2.0/classes/rest/RestServer.class.php:237
> [03-Feb-2017 00:42:17 Europe/London] PHP   4. RestEndpointTransfer->post() /www/filesender-2.0/classes/rest/RestServer.class.php:237

The code the user has written (slightly changed for testing) is as follows:
```
<?php
require_once 'FilesenderRestClient.class.php';
try {
    $c = new FilesenderRestClient('https://cloudstor.aarnet.edu.au/sender/rest.php', 'user', 'username', 'apikey');

    print_r($c->getInfo());

    $c->sendFiles('someonereal@email','someonereal@email','/tmp/x.txt','someonereal@email',null,null,1486082270,null);
    print_r($c->getInfo());
} catch(Exception $e) {
    echo 'EXCEPTION ['.$e->getCode().'] '.$e->getMessage();
}

?>
```

And this is the output:
```
stdClass Object
(
    [url] => https://cloudstor.aarnet.edu.au/sender/
    [name] => CloudStor Sender
    [logon_url] => /simplesaml/module.php/core/as_login.php?AuthId=default-sp&ReturnTo=__target__
    [upload_chunk_size] => 12582912
)
EXCEPTION [0] Http error 500
```

Any ideas/pointers?

jornane:
When not setting any options, don't pass `null` but an empty array instead (`[]`).

Maybe the FilesenderRestClient class should be patched to warn about this, or to sanitize input so that this precondition is met, even when the user provides `null`.

[...]

PR #34, opened 2017-03-03
Use "default" config value for options that the user cannot set

jornane:
There is a bug where FileSender will assume options to have a value of `false` when they are not set to `availble`. This is even the case if `default` is set to `true`. This bug has been reported in #29.

This is problematic for options such as sending mail for guest vouchers, where an administrator wants the mail to be sent but wants to hide the checkbox from the user. 

Pull request #31 seeks to solve this problem by changing `available` to `true` for options where this problem is likely to cause issues. However, this does not solve the problem at hand, which is that FileSender returns the incorrect value for an option that a user is not allowed to change.

This pull request changes how FileSender handles options, which will solve #29. It may also solve some of the problems discussed in #22 (but not all) but I have not tested this.


classes/rest/endpoints/RestEndpointTransfer.class.php
```diff
@@ -350,12 +350,11 @@ public function post($id = null, $add = null) {
...
-                if(in_array($name, $allowed_options) && $data->options->exists($name))
-                    $value = $data->options->$name;
-                
-                $options[$name] = $value;
+                if(in_array($name, $allowed_options)
+                    && $data->options->exists($name))
+                {
+                    $options[$name] = $data->options->$name;
+                }
             }
```
monkeyiq: code below here uses $options['key'] for example using key=TransferOptions::GET_A_LINK. There are only a handful of these accesses but they will not see any default. While what is there might work ok now, it might be brittle to options which are not directly set (and want a default) in the future. 

Maybe we should set the options early, after the for loop;
```
$transfer->options = $options;
```
and then convert those direct $options['key'] lookups to use $transfer->getOption('key').

This update could be done in a subsequent PR, so it's not a blocker :)

monkeyiq:
Once I merged some other UI things I'll do a round of system testing to verify that things are working as one would expect with everything in there.

PR #38, opened 2017-03-17
php7 fix

madsi1m:


jornane:
Thanks for splitting your PR into multiple smaller ones! :1st_place_medal: 

I think that I prefer #30, which solves the same problem as this PR, but doesn't break the Travis build.

madsi1m:
i agree

PR #41, opened 2017-03-17
Weak Random Generator used in Generation of UIDs, replaced mt_rand

madsi1m:


madsi1m:
this is in response to fis001 from the security code audit from surfnet. I went with the openssl as it works in new and old php. I think it come with normal php, i didnt need to add any extra libs (on rhel)

Veredior:
This solution is far better even if there was a huge problem with this function as you can read [here](http://php.net/manual/fr/function.openssl-random-pseudo-bytes.php#118658) .
Tldr : php 5.3 is still insecure with this function.
Do we really need to keep compatibility with PHP 5.3 ?

jornane:
We need to support PHP 5.3 in order to support CentOS 6, which is supported until 2020. There are users running a FileSender 2.0 pre-release on it, and it is important that they can upgrade to FileSender 2.0 stable when that's released, so we don't have users hanging on pre-releases. Users not upgrading from a pre-release is a bigger security risk. 

After that, PHP 5.3 support can be dropped if that allows FileSender to be better.

tl;dr: FileSender == 2.0 will support PHP 5.3, FileSender > 2.0 may drop PHP 5.3, but should still support PHP 5.4.

jornane:
I've checked the affected code. It seems like the way FileSender generates random UUIDs is a bit strange:
* Get a **random** value `$len` between 16 and 32.
* Create a string `$rnd` and add a **random** number between 0 and 1000 to the string, while the string is shorter than `$len`.
* hmac hash  `$rnd` with sha1 and a pre-generated secret as key
* (optional) check if the UUID is unique

I've written the word **random** in bold two times. This pull request changes the implementation of this "random" operation from `mt_rand` to `openssl_random_pseudo_bytes`. This is not an ideal solution. Better would be to fix the `Utilities::generateRandomHexString()` method. For PHP 7 this is easy, as @madsi1m already mentioned. But we should also provide a better solution for PHP 5, because that's what Red Hat and Debian provide out of the box. This must be implemented through a version check in the code, which uses the best random generator for the installed PHP version.

madsi1m:
Updated, if your php has the "better" functions it will use it.

jornane:
This solution is better than what FileSender had before, but there's still room for improvement. I think a rewrite of the UUID-generator is in order, but @meijer told me that there is a reason why it is the way it is. So I'll await a response from him.

madsi1m:
not sure

PR #42, opened 2017-03-17
add spaces to urls in email messages the user type, so that mail clie…

madsi1m:
add spaces to urls in email messages the user type, so that mail clients dont detect it as a url and convert it to a cliackable url. (filesender links are not changed, only what user puts in message)


jornane:
Why shouldn't a user be allowed to put their own links in the message?

madsi1m:
Hi jornane, this is in response to fis013 from the security code audit from surfnet. The report suggests that even though html tags are ripped out of the message email clients still see the urls and create a clickable link.

PR #43, opened 2017-03-22
API fix, now follows documentation for 2.0

madsi1m:


madsi1m:
Updated.

RestEndpointTransfer.class.php was causing an exception.

FilesenderRestClient.class.php helper class now follows the doc in https://github.com/filesender/docs/blob/master/v2.0/rest/README.md

madsi1m:
Hi jorane,

Sorry wrong link, https://github.com/filesender/docs/blob/master/v2.0/rest/README.md#the-upload-process
Scroll to the bottom.

The api says to put {complete:true} in the body not /complete in the url.

The other file was causing an exception because ->exists() doesnt exist on an array which is what is passed.

jornane:
Refactored into 1ed00e0 and 4cc16f2 because this pull request address two issues.

jornane:
Pull request is accepted, but commits have been modified due to refactoring. Closing.

PR #50, opened 2017-04-11
fix cron not deleting transfers of the same day

Veredior:
We have users who can see transfers on "Currently available transfers" panel even if the transfer expired on the same day.
The cause is related to the EXPIRED request called by the cron which doesn't include the same day.

### DIFF
--- a/classes/data/Guest.class.php
+++ b/classes/data/Guest.class.php
@@ -104,7 +104,7 @@ class Guest extends DBObject {
      * Set selectors
      */
     const AVAILABLE = "status = 'available' ORDER BY created DESC";
-    const EXPIRED = "expires < :date ORDER BY expires ASC";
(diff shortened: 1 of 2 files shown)


jornane:
I'm not really sure what this does, but my assumption is that this changes the behaviour of FileSender to interpret the expire date as «unavailable from» instead of «unavailable after»

Does this sound reasonable?

Veredior:
The problem is the application has 2 behaviours for checking if a transfer is expired : 
 - the checks when a guest tries to access a transfer which expired on the same day.
- the cron deletes only the transfers before the day of execution.

I'm not sure if it's understandable.

monkeyiq:
Is this still an issue, if so I will schedule a look see and merging.

fransward:
Related to issue #357 ?
https://github.com/filesender/filesender/issues/357

PR #54, opened 2017-05-08
move to using a stronger random number generator

monkeyiq:
move from mt_rand() to a stronger random number generator


monkeyiq:
random_compat is added to use the php 7 random functions in a php 5x environment.

PR #59, opened 2017-05-12
FIS-007 fix for XSS injection in file name on transfer page

monkeyiq:
File names have no need to include things like the following

```
\"alert(1);
```

Most folks do the right thing, but they are not the ones we are concerned with here. Instead of trying to work out what to do with a blacklist of "you can't do a \" here" it is more secure to declare what characters are valid for a file name and reject the rest. Output is also sanitized just in case something got through before that is nasty.

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -725,7 +725,11 @@ class Transfer extends DBObject {
             
             if(count($matches)) return array_shift($matches);
         }
-        
+
+        if( !Utilities::isValidFileName( $name )) {
+            throw new TransferFileNameInvalidException( $name );
+        }
+
         // Create and save new recipient
         $file = File::create($this);
         $file->name = $name;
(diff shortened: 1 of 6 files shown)


jornane:
Our Norwegian users will complain about now using ÆØÅ anymore.

jornane:
It may be a good idea to make the server side code from this pull request behave similar to the JavaScript validation already in place.

https://github.com/filesender/filesender/blob/c324c890749ae03815b6d238cd4c5b6e949c2c6a/www/js/transfer.js#L165-L168

monkeyiq:
Yep, I will have a subsequent update to do client side validation too.

monkeyiq:
do we have other characters which will be missed by valid users?

monkeyiq:
I'll have a dig on this. likely Ü and other German chars and accents are valid

monkeyiq:
heh, so much for the whitelist being the smart solution :/

monkeyiq:
@jornane https://github.com/filesender/filesender/pull/62 should allow common things to be used again. It will be interesting to see how that works on asian languages.

PR #63, opened 2017-05-14
fis010 limit guest creation and reminder sending

monkeyiq:
The limits are fairly high by default so that most users will not ever hit them. But if a bot or bad person gets at things then they will not be able to go on forever.

### DIFF
--- a/classes/data/Guest.class.php
+++ b/classes/data/Guest.class.php
@@ -97,6 +97,13 @@ class Guest extends DBObject {
         ),
         'last_activity' => array(
             'type' => 'datetime'
+        ),
+        'reminder_count' => array(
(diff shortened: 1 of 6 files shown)


jornane:
Is there a way to disable this for installations where fis010 is not believed to be an issue? Something like setting the limit to -1. Or should I simply set it to `PHP_INT_MAX`?

monkeyiq:
Some really huge number might be the way to go. I can't see a person sending a million reminders for a transfer for any legitimate reason.

jornane:
A heads-up to the mailing list warning about the schema change would have been appreciated. Since there is no FileSender 2.0 stable release, the master version is being used in production at multiple places.

Issue #65, opened 2017-05-15
Issue with Commit 81e7b3c - Init.php autoloader path


jrhoades:
According to https://github.com/filesender/filesender/commit/81e7b3c4f468258094925b2249f860c3d4eae4e2#diff-81771d437d9a0ec666f0f38bd3906391

This line was added:
require_once(FILESENDER_BASE.'/vendor/autoload.php');
But this file & directory don't exist

monkeyiq:
could you please try running

```
composer update
```

in the root directory of your filesender install. ie, the same directory that has the composer.json file in it.
you should then have a vendor/autoload.php

jornane:
I have the same problem. When I try your fix, I get `-bash: composer: command not found`.

monkeyiq:
hmm, composer is packaged for Fedora, but it seems not for debian.

maybe I should look at committing the vendor (composer output) stuff or just the php7 random number overlay.

monkeyiq:
I will make a PR to do the later tomorrow. 

there is already a composer setup in FileSender but it seems that it was optional.

https://getcomposer.org/download/

jornane:
The Composer setup in FileSender before 81771d4 is only for unit tests and not needed or even recommended to use in production. FileSender always had all dependencies vendored in.

I see you want to use [paragonie/random_compat](https://github.com/paragonie/random_compat), note that they also have a composer-free installation guide.

meijer:
Ben, 

Up until now we've always tried to have as few dependencies as possible. Sometimes this has meant coding some things ourselves, rather than glueing in a big library that in turn would need to be security tested as well. Depends on the particular problem that was trying to be solved obviously.

In any case, when there's a change in dependencies, the following needs to be considered:
- document (in the commit most likely) why the dependency needs to be added
- modify the installation guide (https://docs.filesender.org/, documentation is now part of the source tree on github)
- inform the user community


Cheers,
Jan

monkeyiq:
The above PR should have fixed this issue. Sorry about that.

PR #66, opened 2017-05-15
Move random_compat to be internal the to project tree

monkeyiq:


monkeyiq:
This should resolve https://github.com/filesender/filesender/issues/65 in the right way

Issue #25, opened 2017-02-15
1071 Specified key was too long when updating UserPreferences key


thijskh:
Hi,

When running the database upgrade on current master, I get a MySQL error "1071 Specified key was too long; max key length is 767 bytes" (see below). This is correct since the update tries to enlarge the id column from varchar(128) to varchar(255). Since for encoding reasons this requires up to 4 bytes per character, the requested size is 1020 bytes which oversteps InnoDB's prefix length for a primary key.

I wonder what the reason is for extending this column and if it's the best idea to have a primary id be a very long varchar that's hard to index? Perhaps use a numeric auto_increment id instead?

```
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException exception uid:58a456c7e6cd7] database_access_failure
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException trace uid:58a456c7e6cd7] #0 /srv/filesender/classes/utils/DatabaseMysql.class.php(245): DBI::__callStatic('exec', Array)
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException trace uid:58a456c7e6cd7] #1 /srv/filesender/classes/utils/Database.class.php(123): DatabaseMysql::updateTableColumnFormat('UserPreferences', 'id', Array, Array)
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException trace uid:58a456c7e6cd7] #2 /srv/filesender/scripts/upgrade/database.php(97): Database::updateTableColumnFormat('UserPreferences', 'id', Array, Array)
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException trace uid:58a456c7e6cd7] #3 {main}
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException details uid:58a456c7e6cd7] "SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 767 bytes"
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException details uid:58a456c7e6cd7] Array
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException details uid:58a456c7e6cd7] (
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException details uid:58a456c7e6cd7]     [name] => exec
[2017-02-15 13:25:27] [upgrade:error] [DBIUsageException details uid:58a456c7e6cd7]     [args] => Array
[...]

PR #92, opened 2017-07-01
avoid hitting index size limit on mysql

monkeyiq:
If the user is on a mysql type database and has character encodings that take 4 bytes per character then using varchar(255) will fail to create an index. The simple fix is to only allow email addresses up to 190 bytes long.


monkeyiq:
This was reported here https://github.com/filesender/filesender/issues/25

Issue #36, opened 2017-03-14
Documentation in webinterface


jornane:
The FileSender interface has buttons «About» and «Help». These buttons show information that is hardcoded in the language files. This information is a bit sparse, in that «About» doesn't say so much about FileSender or the particular instance that is running, and «Help» doesn't actually provide help so much as an insight in the configuration settings.

We need a better «About» text and an actual user help page.

The current content is as follows:

### About
> TEST FileSender is an installation of FileSender (www.filesender.org), which is developed to the requirements of the higher education and research community.

### Help
> Login
> 
>   * You log in through one of the listed Identity Providers using your standard institutional account. If you do not see your institution in the list, or your login fails, please contact your local IT support
> 
> 
> Your browser's features
> 
>   * You can upload files of any size up to `max_transfer_size` per transfer.
>   * You can upload files of at most `max_legacy_file_size` each and up to `max_transfer_size` per transfer.
> 
> 
> Uploads of any size with HTML5
> 
>   * You'll be able to use this method if the  sign is displayed above
>   * To enable this functionnality simply use an up to date browser that supports HTML5, the latest version of the "language of the web".
>   * Up to date versions of Firefox and Chrome on Windows, Mac OS X and Linux are known to work.
>     
>   * You can resume an interrupted or cancelled upload. To resume an upload, simply send the exact same files again !
>      Make sure the files have the same names and sizes as before.
>      When your upload starts, you should notice the progress bar jump to where the upload was halted, and continue from there.
>     
> 
> 
> Uploads up to `max_legacy_file_size` per file without HTML5
> 
>   * FileSender will warn you should you try to upload a file that is too big for this method.
>   * Resuming uploads is not supported with this method.
> 
> 
> Downloads of any size
> 
>   * Any modern browser will do just fine, nothing special is required for downloads
> 
> 
> Configured service constraints
> 
>   * Maximum number of recipients : `max_transfer_recipients` email addresses separated by a comma or semi-colon
>   * Maximum number of files per transfer : `max_transfer_files`
>   * Maximum size per transfer : `max_transfer_size`
>   * Maximum file size per file for non-HTML5 browsers : `max_legacy_file_size`
>   * Transfer expiry days : `default_transfer_days_valid` (max. `max_transfer_days_valid`)
>   * Guest expiry days : `default_guest_days_valid` (max. `max_guest_days_valid`)
> 
> 
> Technical details
> 
>    
>   * `site_name` uses the FileSender software.
[...]

PR #93, opened 2017-07-02
Freshen up help, about, and splast text

monkeyiq:
The help text now has some lead content to help users as well as telling them of the system limits and if they can use HTML5 or not.

I also moved these to being pages to allow simpler browser back type navigation. Going to about or help is probably a destination in itself rather than a dialog to pop up.


monkeyiq:
This is in response to https://github.com/filesender/filesender/issues/36

PR #94, opened 2017-07-02
allow per site, per lang about, help, splash pages

monkeyiq:


monkeyiq:
https://github.com/filesender/filesender/issues/36

Issue #45, opened 2017-03-24
"My Transfers" page very slow


madsi1m:
Hi All,

When i click "My Transfers" or after an upload the pages does many sqls and takes minutes to load the transfers page.

Any pull requests or work done to solve this?

Cheers

jornane:
I've noticed the same thing. We need some profiling in order to find out where the problem is, but this may take a long time to debug.

My *****guess***** is that the my transfers page runs a lot of SQL queries, which have a little overhead each, counting up to a lot.

madsi1m:
i added some logging and it does many many sqls

jornane:
Probably some `for`-loop then. Can you provide some more information? Which tables are accessed oftest?

madsi1m:
ok when clicking "My Transfers" the first few queries (I didnt let it finish):
`
SELECT * FROM Transfers WHERE user_id = :user_id AND status='available' ORDER BY created DESC
SELECT * FROM TrackingEvents WHERE target_type='Recipient' AND target_id = :id ORDER BY created
SELECT * FROM TrackingEvents WHERE target_type='Recipient' AND target_id = :id ORDER BY created
SELECT * FROM AuditLogs WHERE target_type = :type AND target_id = :id ORDER BY created ASC, id ASC
SELECT * FROM AuditLogs WHERE target_type='File' AND target_id IN :ids
SELECT * FROM AuditLogs WHERE target_type='Recipient' AND target_id IN :ids
SELECT * FROM AuditLogs WHERE author_type = :type AND author_id = :id ORDER BY created ASC, id ASC
SELECT * FROM AuditLogs WHERE author_type = :type AND author_id = :id ORDER BY created ASC, id ASC
SELECT * FROM AuditLogs WHERE target_type = :type AND target_id = :id ORDER BY created ASC, id ASC
SELECT * FROM TrackingEvents WHERE target_type='Recipient' AND target_id = :id ORDER BY created
SELECT * FROM TrackingEvents WHERE target_type='Recipient' AND target_id = :id ORDER BY created
SELECT * FROM AuditLogs WHERE target_type = :type AND target_id = :id ORDER BY created ASC, id ASC
SELECT * FROM AuditLogs WHERE target_type='File' AND target_id IN :ids
SELECT * FROM AuditLogs WHERE target_type='Recipient' AND target_id IN :ids
SELECT * FROM AuditLogs WHERE author_type = :type AND author_id = :id ORDER BY created ASC, id ASC
[...]

PR #95, opened 2017-07-04
Support for secondary indexes in mysql and pgsql

monkeyiq:


monkeyiq:
https://github.com/filesender/filesender/issues/45

PR #96, opened 2017-07-04
Augment the 'key' security with a file fingerprint

monkeyiq:
This makes life a little harder for attackers as they would have to either control the creation of the transfer or force a person creating a transfer to use a known file prefix byte set. It is not as good as having saml last the whole session but it does help to close up some options for possible attackers.

This can be turned off in the config using
protect_upload_with_fingerprint => false


jornane:
I don't think I fully understand this pull request. What would an attacker be able to do if this is not merged?

(sidenote: don't forget to [update the documentation](https://github.com/monkeyiq/filesender/blob/2017/key-with-fingerprint/docs/v2.0/admin/configuration/index.md) if you add configuration directives ;)

monkeyiq:
The basic idea is that the key which is passed in CGI is enough to authenticate and "do things". So if somebody accidentally sent a URL with the key in it then others could use that to authenticate themselves against the transfer. 

In the old days they could delete things that way. I've already limited the scope of that so that once the transfer is complete they can't use just the key to delete things in the transfer. You always have to have current and valid saml authentication to delete a completed transfer now. 

This PR sends the fingerprint (hash of first bunch of bytes of file) to the server when the transfer is setup. Then to upload you also need to send the fingerprint of the file in an HTTP header. This should always be trivial if you are the person who made the transfer as you have all the files on hand. 

If on the other hand you get the magical key value and want to try to upload then you will have to know the finderprint used when setting up the transfer. In a future PR, if the server verifies the fingerprints against the uploaded data after upload then it can reject files that are not really matching their proposed fingerprint. So I would then need the key (which you might accidentally leak to me) and then find out the fingerprint and then be able to make a second image that will hash to the same fingerprint and try to upload that.

jornane:
I'm still a bit unclear on the exact threat model here.

> So **if somebody accidentally sent a URL with the key in it** then others could use that to authenticate themselves against the transfer.
_(emphasis mine)_

As I understand it, as soon as the file is completely uploaded, it cannot be changed anymore, even with the key. A SAML session is required. That means that an attacker must obtain the key _before_ the upload completes. FileSender won't sent a link until all files in a transfer are uploaded.

Can you describe how an uploader may be incentivised or tricked to send an URL with the key before the upload is complete? Judging from the UI as it is today, I couldn't find the key even if I wanted to. I can find it with the inspector during upload, but if an attacker has that level of access, the upload is compromised anyway.

PR #97, opened 2017-07-08
make strings 190 chars for possible sec indexing in mysql

monkeyiq:
Sometimes folks use four byte character encodings in mysql. it seems there are combinations out there that only allow 190(x4) bytes to be indexed as primary or secondary indexes. So if it doesn't matter if a string is 190 or 255 chars long, maybe we should make it 190 so that we can add a secondary index to the column later if desired without having to think about column truncation at that time.

I still have to run over these changes to reason able if they truncate something that will feel it.

### DIFF
--- a/classes/data/AuditLog.class.php
+++ b/classes/data/AuditLog.class.php
@@ -53,20 +53,20 @@ class AuditLog extends DBObject {
         ),
         'target_type' => array(
             'type' => 'string',
-            'size' => 255
+            'size' => 190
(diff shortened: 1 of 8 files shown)


monkeyiq:
this seems to have been lifted in the last year. I am not sure which version of mariadb made indexing 4 byte strings up to 768 characters (max key length is 3072 bytes) but on modern systems this doesn't seem to be an issue any more.

Issue #79, opened 2017-06-09
{invalid_file_name} error when uploading file containing spaces in the filename


fransward:
I'm experiencing errors when trying to send a file containing spaces in the filename.
This happens with any file extension.

```
Screen Shot 2017-05-26 at 12.54.12.png : 144.3 kB
{invalid_file_name}
```
When I rename the file by removing the spaces no error will occur anymore.

btw, the follow default setting is active in the config:
  $config['ban_extension'] = 'exe,bat';

bart-j:
This is due to the default $config['valid_filename_regex'] setting. One solution is to disable this regex with:

    $config['valid_filename_regex'] = '';

There is still another check in the Javascript library validating the filename, so I am not sure why there are two checks.

monkeyiq:
the above pr should fix this issue

PR #99, opened 2017-07-13
allow spaces in uploaded file names

monkeyiq:


monkeyiq:
fixes https://github.com/filesender/filesender/issues/79

PR #103, opened 2017-07-18
Crypto fix for terrasender

madsi1m:
This resolves an issue i see when testing encryption with parallel uploads when a file is larger than a single chunk + clean up


jornane:
I support moving the calculation of `$file->encrypted_size` from RestEndpointFile.class.php to Transfer.class.php. While you're making these improvements, could you add documentation on how it is calculated? Especially the role of `$lastChunkPadding` I find quite hard to wrap my head around.

Try to watch your whitespace changes; you commited changes to AuditLog.class.php and Recipient.class.php and reverted them later, but there are whitespace changes left. Also in Transfer.class.php you add a whitespace error.

Again, moving the logic of `$file->encrypted_size` to Transfer.class.php: good idea, keep it up! :+1:

www/download.php
```diff
@@ -241,7 +241,7 @@ function downloadSingleFile($transfer, $recipient, $file_id, $recently_downloade
         if (!$chunk_size)
             $chunk_size = 1024 * 1024;
 
-        if($file->encrypted_size){
+        if($file->transfer->options->encryption){
```
jornane: `$file->encrypted_size` is set even if `$file->transfer->options->encryption`. Maybe it would be a good idea to _not_ calculate `$file->encrypted_size` if no encryption is used. This would need to be changed in Transfer.class.php.

jornane: This change by itself is good, though. Checking that a flag is explicitly true is easier to read than checking if a variable is not falsey.

madsi1m: I prefer to always set file->encrypted_size because we always set file->size

madsi1m:
`$transfer->options['encryption'] == 1` seems to work and read better

monkeyiq:
This is merged in https://github.com/filesender/filesender/pull/134

PR #104, opened 2017-07-19
some more secondary indexes

monkeyiq:


monkeyiq:
The bulk of the indexes from https://github.com/filesender/filesender/issues/45#issuecomment-314610047

I have left of TType_TID_AType_AID_IP_EVENT_Created because I am not too sure how used a 6 column index would be.

Issue #112, opened 2017-08-04
Creating medium sized database set


monkeyiq:
A meduim sized database set would help in many cases such as regression testing, database schema modifications, performance enhancement where datasets are large enough to expose speed issues that are not seen on small / trivial databases.

This should open the door for adding more secondary indexes in cases where they will enhance performance on real world datasets. I previously used a dataset with 200,000 tuples in the AuditLog to see and assist the performance of the my transfers page.

It is likely to be useful to get some feedback on real world databases if possible. Mainly high level statistics on how many tuples are in some tables and the selectivity of certain queries. So, no sensitive data :)

The first round feedback might also indicate more high level stats that might be useful in replicating medium sized data. Obviously my current database still needs padding out to be more like a regular production environment.

```
MariaDB

select count(*) from UserPreferences;
+----------+
| count(*) |
+----------+
|        7 |
+----------+

select target_type,count(*) as count from AuditLogs group by target_type;
+-----------------+--------+
| target_type     | count  |
+-----------------+--------+
| Transfer        | 200041 |
| File            |    196 |
| Transfer        |    503 |
| User            |      7 |
+-----------------+--------+

select target_type,count(*) as count from TrackingEvents group by target_type ;
+-------------+-------+
| target_type | count |
+-------------+-------+
| Recipient   |     1 |
+-------------+-------+

1 row in set (0.00 sec)
select count(*) as count, status from Transfers group by status;
+-------+-----------+
| count | status    |
+-------+-----------+
|   118 | available |
|     6 | closed    |
|    86 | created   |
|     2 | uploading |
+-------+-----------+


select min(c) as files_per_transfer,count(*) as count
from (select count(*) as c from Files group by transfer_id) sub group by c;
+--------------------+-------+
| files_per_transfer | count |
+--------------------+-------+
|                  1 |   174 |
|                  2 |     3 |
+--------------------+-------+

select min(c) as recip_per_transfer,count(*) as count from (select count(*) as c from Recipients group by transfer_id) sub group by c;
+--------------------+-------+
| recip_per_transfer | count |
+--------------------+-------+
|                  1 |   126 |
+--------------------+-------+

select count(*) as count,translation_id from TranslatableEmails group by translation_id;
+-------+--------------------------+
| count | translation_id           |
+-------+--------------------------+
|     3 | file_deleted             |
|     6 | report_inline            |
|     6 | transfer_deleted_receipt |
|     5 | upload_complete          |
+-------+--------------------------+
```

[...]

PR #115, opened 2017-08-18
Looking to extend the CI to include and use the new test dataset

monkeyiq:
The CI system now imports a recent postgresql dataset, uses the migration script to bring that config forward to migrate to what the current php is expecting and then does some sanity checks as part of the phpunit execution.


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/112

monkeyiq:
now I can switch to cleaning this up a bit :)

PR #117, opened 2017-09-07
Import some of the docs from assembla

monkeyiq:
Use of Assembla is being decommissioned. These are a few pages from the old wiki which seems useful to have on the new github based site. There may be more to import, and the releases instructions need some more freshing up love for sure.


monkeyiq:
I'll give Travis a moment to regroup

Issue #113, opened 2017-08-17
Stall warning on large files causing failed uploads on 2.0beta1.


monkeyiq:
This bug was reported on the mailing list.

During an upload, despite the fact that I'll be getting good speeds, I'll get the following message.

"Upload seems to be stalled (way slower than expected), do you want to try restarting it or stop it ?"

The upload still continues in the background. I can choose ignore and it will continue, but often once then upload reached 100% it still  continues to uploads. Most recent test showed me uploading 10GB of a 6GB file.

Turning off terasender stops the issue. ( with performance drop). But more accurately, leaving terasender on and turning off stall detection, obviously stops the alert, but also seems to stop the sanity check entries in the logs.

To disable stall detection:
$config['stalling_detection'] = false;

madsi1m:
I've seen this on an older commit as well. And disabling stall detection did fix it. I re-enabled it on ours and we no longer have the issue. It could be a timing, with terrasender not all chunks respond in the order they were uploaded, but haven't done any digging (yet)

monkeyiq:
There are some assumptions in the old detection code which have been updated in https://github.com/filesender/filesender/pull/209 which should lead to less false positives for stall detection.

jornane:
> The upload still continues in the background. I can choose ignore and it will continue, **but often once then upload reached 100% it still continues to uploads.** Most recent test showed me uploading 10GB of a 6GB file.

This issue is not addressed in #209.

WebSpider:
@monkeyiq it would seem this issue is no longer valid, at least im not seeing incorrect stall reports in our production environment anymore

monkeyiq:
Thanks for the update! It is very hard to trigger such stalls in local testing.

WebSpider:
Relabeling so we can solve this (or determine it is solved) before 3.0 release

WebSpider:
Not seen any reports about stalling issues for over a year.

PR #118, opened 2017-09-07
Disable upload stall detection for now

monkeyiq:
It has been reported on the mailing list and referenced in https://github.com/filesender/filesender/issues/113 that at times the upload stall detection can wrongly think it is in a stall state, failing a good upload.

While a test, replicate, reduce and remove cycle happens, the mailing list discussion has chosen to disable stall detection for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: precise
+sudo: required
 language: php
 php:
 - 5.3
--- a/docs/_includes/menu.html
+++ b/docs/_includes/menu.html
@@ -3,9 +3,9 @@
 	<ul>
 		<li><a href="{{site.baseurl}}/roadmap/">Roadmap</a>
 	</ul>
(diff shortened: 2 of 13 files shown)


jornane:
I think that this is a great idea! However, I think that you also need to modify [ConfigDefaults.php](https://github.com/monkeyiq/filesender/blob/ff71244cf26bb60a73cba2c0f4b5f1f7ae3fa086/includes/ConfigDefaults.php#L99), otherwise it's still going to be on. The conditional in www/filesender-config.js.php may not actually be necessary.

monkeyiq:
Interesting, makes sense to take a look at others like terasender_worker_count which are effectively setting defaults in the code rather than shipping them in ConfigDefaults.

Much cleaner to ensure the default is in ConfigDefaults rather than having the current two tier setup for some of these.

monkeyiq:
This also cleans up a few cases where defaults were embedded in www/filesender-config.js.php rather than coming from includes/ConfigDefaults.php. Fewer places with default fallbacks should make for fewer surprises finding different values.

Issue #53, opened 2017-04-26
Improve error messages when auth_sp_saml_*_attribute do not match up


monkeyiq:
In AuthSPSaml.class.php / attributes() the throwing of AuthSPMissingAttributeException if uid and email etc are not found can make debugging the error a little difficult. For example, using a hosted UserPass setup in simplesamlphp if you happen to get the uid wrong in config.php "auth_sp_saml_uid_attribute" or do not set that configuration value then the error message generated does not offer great help in working that misconfiguration out.

it might be helpful to use a logger::error() or something in there with a reasonably verbose message to help the admin work out the issue.

This is likely to flow to other auth methods where you might have the email/uid either not set or set with a slight typo and may become frustrated trying to work out whey auth seems to work but is not working.

PR #127, opened 2017-09-16
sysadmin: Help the sysadmin with missing SAML attributes

monkeyiq:
When a key is mistyped or the sysadmin is expecting different keys in the auth session than are there then things can be hard to debug. For example, using a local user/password in simplesaml one might select the wrong uid attribute but have no idea what the right choice is. With this serverlogging then the sysadmin gets some more clues that can be found with the finding out details hex string from the browser.


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/53

PR #128, opened 2017-09-18
REST Client: fix for array and object when checking properties

monkeyiq:
This was taken out with https://github.com/filesender/filesender/pull/34/files

It turns out that options is an array() from the Rest Client and there is no exists() method on array. Until things are normalized to object or array throughout this check has to come back so that the rest clients can upload again.


jornane:
Isn't [`in_array`](http://php.net/in_array) a better fit than [`array_search`](http://php.net/array_search) here?

monkeyiq:
I just reverted that part of the code to what was there before the changes you made in PR34. There are likely improvements that can be made to the previous code that I brought back, but it does allow the REST client to work again.

madsi1m:
Hi All, this change seems to work. I'm deploying this on our production instance.

Cheers

Issue #98, opened 2017-07-11
Allow for different separator in query string


jornane:
We've noticed that some mail clients mangle the download URL by double-HTML-escaping it. This causes the clickable URL to contain `&amp;` instead of simply `&`. Of course, this is a bug in the mail client in question and not in FileSender, but it's annoying nontheless and causes confusion.

The HTTP standard allows for a different separator to be used, `;` is a commonly chosen as an alternative. [PHP allows the administrator to choose the preferred separator](http://php.net/manual/en/ini.core.php#ini.arg-separator.output). However, FileSender has `&` hardcoded in multiple locations, rendering this setting useless.

In order to fix this, FileSender must build URLs in a more standardized way. This means that all URLs with a query parameter must be generated using the [`http_build_query`](http://php.net/manual/en/function.http-build-query.php) function. As a consequence, no URLs can be generated in JavaScript anymore; URLs must be generated server-side and passed to JavaScript through Ajax or data-attributes in the DOM.

monkeyiq:
Perhaps the separator (```&``` by default) could be sent to the javascript and if that was used during building then the same code might work for servers configured with ```;``` as the alternative.

peter-:
See [this email](https://sympa.uninett.no/lists/filesender.org/arc/filesender-dev/2018-10/msg00010.html) where I explained how I force Filesender to use "clean URLs" in generated emails (and web server redirects to map them back), avoiding the use of request parameters and therefore any encoding issues completely. (Currently this requires hacking of `Guest.class.php`, `Recipient.class.php` and `TranslatableEmail.class.php` to return differently formatted links. If the generated URL format was configurable this wouldn't be necessary.)

As always there was no response, so it seems clear there is zero interest from the project in avoiding such issues, also evidenced by this year-old issue here.

jornane:
> As always there was no response, so it seems clear there is zero interest from the project in avoiding such issues, also evidenced by this year-old issue here.

FWIW, I was really happy seeing your PR #444, and I'm also glad that you're working on this. I got the same impression as you when I reported this; that the only response is (crickets). #137 was merged, but that still requires JavaScript to build URLs, which makes what you're trying to do really hard. My original proposal (from this issue) is to generate all URLs server-side, and pass them to the client, for example via `data-` attributes. Then clean URLs should be a lot easier to implement.

PR #137, opened 2017-09-27
Allow CGI parameter separator to be configured

monkeyiq:
This allows the separator for CGI parameters to be set to something other than "&". Note that you will have to ensure your php.ini allows your setting too. By setting the separator to ";" you can avoid conflicts with some email clients escaping the "&" too much for example. This issue was reported in https://github.com/filesender/filesender/issues/98.

This PR targets some specific locations and I have checked over some of the emails and links on the site to see the ";" appear if configured to do so.

### DIFF
--- a/classes/auth/AuthSPSaml.class.php
+++ b/classes/auth/AuthSPSaml.class.php
@@ -162,10 +162,11 @@ class AuthSPSaml {
             $target = Config::get('site_url').'index.php?s='.$landing_page;
         }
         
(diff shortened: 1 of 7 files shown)


monkeyiq:
hmm, this fails because enc_type was added in php 5.4 and the CI runs in 5.3.

jornane:
I'd recommend removing the config entry and reading the value directly from php_ini. The values must match anyway for parsing to work on the PHP side.

PR #139, opened 2017-09-27
Graphs: show average upload performance for last 30 days

monkeyiq:
Add a chart to the bottom of the upload page showing the average performance for file uploads over the last 30 days. This is only working on the mysql backend only at the moment, see the main query at the top of uploadGraph.php if you want the details.

This is disabled by default. See the index.md file for information on upload_graph_bulk_display to enable.

I made some changes to the patch, for example;

* SQL query is broken out from being a single long string to be easier to read
* the minimum file size to consider for stats is configurable, defaulting to 1gb.
* the y axis now has a label and the chart title displays the configured min file size
* configuration settings defaulting to off to allow the chart to be enabled or not

Another patch from Michael D'Silva & AARNet.


monkeyiq:
This is what it looks like:
filesender-chart

PR #143, opened 2017-09-30
Stats: Some more stats for the admin page #141

monkeyiq:
A new section on the admin page showing some more stats. This is based on a patch from Michael D'Silva & AARNet reworked to execute on MySQL and PostgreSQL backends. There is a new DBLayer class to help with date/time function differences between the two database types.


monkeyiq:
this is the same as https://github.com/filesender/filesender/pull/141 basically.

PR #145, opened 2017-09-30
cgi separator: move to just using php config

monkeyiq:


monkeyiq:
This moves to using the PHP parameter directly
https://github.com/filesender/filesender/pull/137

Issue #136, opened 2017-09-26
Ability to migrate from Filesystem to FilesystemChunked


monkeyiq:
The new FilesystemChunked storage allows files to be stored as chunks instead of being assembled into a single file on disk. There is no easy way to convert from storage_type Filesystem to FilesystemChunked while retaining the existing files at the moment.

Current winning option;
* store the storage_type used for the transfer in the database so the admin can change things as they like without breaking any existing uploaded files. Possibly add a migration script to fill the storage_type based on if it is a file or directory so that existing installations from master can be easily updated.

A few other options come to mind;
* Graceful fallback. If the uid is not a directory then fallback to Filesystem storage for that file so existing files can still be served while new files are created chunked.
* Script to convert files into chunked files. This might be less desirable than the above due to run time. The script could run two pass, in the first creating the chunks in directories and the later just renaming the directories and files. This would allow chunked files to be done while the system is running and lessen the downtime while switching storage to chunked (the second renaming pass).

jornane:
Add a field to the database so we know for each file which storage class was used. This will make it safe for an admin to change the value of the storage class without being afraid of stuff breaking, and it will prevent running uploads from being interrupted when the value is changed (the value is set at the start of the upload).

jornane:
For backwards compatibility, it could assume StorageFilesystem if no class is set (files created before the change) but that would break stuff for those already using FilesystemChuncked. I'd still pick this solution, and send out a warning to everyone, stating that you're not affected if you use the latest beta or older.

monkeyiq:
Adding the storage_type used during upload to the database seems like the right choice. There is some ability to workout fallbacks now but it will likely not scale as other storages are added.

monkeyiq:
I editing the top post to reflect the current desired winning solution.

madsi1m:
current winning solution will work fine. Probably not necessary for a change script. Maybe back fill the old transfers to default?

PR #153, opened 2017-10-10
Storage: bind storage_type to each file that is saved

monkeyiq:
The addition of the new chunked storage type creates some issues when folks might want to migrate to using it while still serving old files that have already been transferred.

This PR will require a database upgrade to work;
```
php /opt/filesender/scripts/upgrade/database.php
```

Note that I have set the default storage for existing transfers to 'Filesystem'. If you have been using the chunked storage type then you will need to update the database with the below command to override the default setting, note use files as the table in postgresql and Files in MySQL.
```
update files set storage_type = 2 ;
```

Since each file saves the int value for the storage_type that was active when it was created you can more easily change the storage_type in the future while still serving existing files.

### DIFF
--- a/classes/autoload.php
+++ b/classes/autoload.php
@@ -51,6 +51,7 @@ class Autoloader {
         'GUIPages' => 'constants/',
         'LogLevels' => 'constants/',
         'ReportFormats' => 'constants/',
+        'DBStorageType' => 'constants/',
         
         'Storage' => 'storage/',
         'Storage*' => 'storage/',
--- a/classes/constants/DBStorageType.class.php
+++ b/classes/constants/DBStorageType.class.php
@@ -1,0 +1,95 @@
+<?php
+
+/*
+ * FileSender www.filesender.org
+ * 
+ * Copyright (c) 2009-2014, AARNet, Belnet, HEAnet, SURFnet, UNINETT
+ * All rights reserved.
+ * 
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * *	Redistributions of source code must retain the above copyright
+ * 	notice, this list of conditions and the following disclaimer.
+ * *	Redistributions in binary form must reproduce the above copyright
+ * 	notice, this list of conditions and the following disclaimer in the
+ * 	documentation and/or other materials provided with the distribution.
+ * *	Neither the name of AARNet, Belnet, HEAnet, SURFnet and UNINETT nor the
+ * 	names of its contributors may be used to endorse or promote products
+ * 	derived from this software without specific prior written permission.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS 'AS IS'
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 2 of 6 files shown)


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/136

jornane:
Interesting solution enumerating all possible storage classes. I would think that this might cause issues with parties implementing their own storage, because they would have to add a non-overlapping number. Have you designed an addressing plan for storage types?

Have you considered simply storing the class name of the storage type in the database as a varchar?

madsi1m:
+1 in class name

On 10 Oct 2017, at 19:01, Jørn Åne <notifications@github.com<mailto:notifications@github.com>> wrote:


Interesting solution enumerating all possible storage classes. I would think that this might cause issues with parties implementing their own storage, because they would have to add a non-overlapping number. Have you designed an addressing plan for storage types?

Have you considered simply storing the class name of the storage type in the database as a varchar?

monkeyiq:
TLDR; I personally wouldn't choose to varchar it. But since folks want that I will move to it.

My addressing plan for the future was to use a lookup table in the database
```
storage_types
-----------+----
id         | int primary key autoincr
class_name | varchar(100)
```

The Files.storage_type would refer to a storage_types.id. The actual value of that ID is not important. No addressing plan required, nobody has to care about getting a number. A script run once per code update or on request with third party storage classes would ensure that every storage class has an entry in the lookup table.

then a view on Files could integrate the id->class_name resolution and return that as filesView.storage_class_name which would be the varchar. The important part here is you are only actually storing about 4 bytes instead of 120+ per tuple. I only used the php lookup table for now because there is no view support in the db layers and we only recently went from one to two storage_types in total.

Using an int or 4 byte int with a small lookup table (fully RAM cached) is so much more efficient than storing a varchar(30) per tuple. In 4 byte unicode that becomes 120 bytes+ of storage per tuple. While I agree that disk space is cheap it will also lead to larger tables and less efficient indexes and caches.

jornane:
Lookup in an on-the-fly populated database table is a good solution, lookup using hard-coded ENUMs is problematic I think.

 By using ENUMs as a "temporary" solution, you'd have to write a migrate script later to translate these ENUMs to database entries. That would make future setups a bit hairy, because then there would be some "special" numbers that must map to known storage types.

My opinion:

ENUM < VARCHAR < Lookup-DB

monkeyiq:
I feel very confident that I could arrange to insert two tuples into a newly created table during migration. 

The database table would never have been populated "on the fly", databases are kind of good at storing information, so I would trust it to store these two tuples for the join table.

madsi1m:
My feeling is when adding an extra custom storage type not already included in the code base and it getting an id because auto increment, followed by code being updated by adding another storage type, the ids would not be the same on everyone’s deployments, which isn’t bad, just slightly harder to support when other installs get stuck. I think sticking to the storage class name is simple enough and not taxing on the db. Also I don’t want to have to insert into a storage table each time I write a new storage backend.

On 22 Oct 2017, at 12:28, monkeyiq <notifications@github.com<mailto:notifications@github.com>> wrote:


I feel very confident that I could arrange to insert two tuples into a newly created table during migration.

The database table would never have been populated "on the fly", databases are kind of good at storing information, so I would trust it to store these two tuples for the join table.

monkeyiq:
there are easier pickings to db updates to improve performance.

PR #156, opened 2017-10-13
Do not require session to signal upload complete

jornane:
When a user uploads a large file and the security token expires during upload,
FileSender has code in place to allow the upload to continue.
However, when the whole file has been uploaded, the browser must signal upload complete.
Now, FileSender suddenly does check the session and aborts the upload, thus defeating the purpose of allowing the upload to continue.

This commit will allow for the signaling of upload complete without checking the security token, in addition to the already existing code to allow for uploading chunks without checking the security token.  Thus allowing a user to complete a upload when the security token expires.


monkeyiq:
seems I've been chasing travis all day, I'll keep looking at these as I can get the CI to complete for them. This one is next :)

PR #157, opened 2017-10-13
Fix the random number generator and generate valid UUIDs

jornane:
In order to understand this commit, you need a bit of history.

During the PINE security audit in 2015, a vulnurability was found in the way FileSender generates random identifiers (2.6.13);
PINE found that `unid_id` and `mt_rand` by themselves weren't sufficient and recommended using HMAC for random number generation.
The new implementation thus used `mt_rand` to generate a random HMAC key the first time this was needed,
and all subsequent random numbers would consist of a random number from `mt_rand`, HMAC-ed with this stored random value.
This implementation was never security checked as far as I know.

In 2017, a new security audit was conducted by Radically Open Security.  They found the same code using `mt_rand`, and acknowledged the HMAC in the code (FIS-001), but recommended that these be replaced by newer PHP functions that weren't available during the previous audit.

This was then implemented (PR#54) by adding a polyfill for the newer PHP functions and replacing occurrences of `mt_rand` with `random_int`, which is not in the spirit of the recommendation in FIS-001.

This pull request seeks to remedy FIS-001 the way that was proposed, by replacing the logic for generating random identifiers, instead of simply replacing a function call.

### DIFF
--- a/classes/utils/Utilities.class.php
+++ b/classes/utils/Utilities.class.php
@@ -34,7 +34,8 @@
 if (!defined('FILESENDER_BASE'))
     die('Missing environment');
 
-require_once(FILESENDER_BASE.'/lib/random_compat/lib/random.php');
+if (!function_exists('random_bytes'))
+    require_once(FILESENDER_BASE.'/lib/random_compat/lib/random.php');
 
 /**
  * Utility functions holder
@@ -79,11 +80,19 @@ class Utilities {
             return $uid;
         }
         
-        // Generate a simple random UID
-        
-        $rnd = self::generateRandomHexString();
-        
-        return substr($rnd, 0, 8).'-'.substr($rnd, 8, 4).'-'.substr($rnd, 12, 4).'-'.substr($rnd, 16, 4).'-'.substr($rnd, 20, 12);
+        // Generate a simple random UUID
+        $bytes = random_bytes(16);
+        $bytes{8} = chr((ord($bytes{8}) & 0x3F) | 0x80);
+        $bytes = bin2hex($bytes);
+        $bytes{12} = '4';
+        $parts = str_split($bytes, 4);
+        return implode('-', [
+                $parts[0] . $parts[1],
+                $parts[2],
+                $parts[3],
+                $parts[4],
+                $parts[5] . $parts[6] . $parts[7]
+        ]);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


monkeyiq:
We may be reading a different document containing the "new security audit was conducted by Radically Open Security". I can only go on what I have seen.

The recommendation as I see in my copy of the document is:

```
"ROS recommends to use a function who does generate cryptographically secure values, 
like random_int(), random_bytes(), or openssl_random_pseudo_bytes()"
``` 

I am not sure what part of the above recommendation moving to using random_int() does not address. This is also the explicit word for word recommendation from the document I have, so was surely a "remedy FIS-001 the way that was proposed".

If we are talking about "spirits" of security recommendations then a common one is to leave audited code in place and make the minimal code changes that are actually recommended by the audit.

I am happy to make changes that benefit the project. There is no need to try to rewrite history to do it though.

monkeyiq:
So you are saying that ROS are unable to articulate their recommendations. Even when they have a well structured document with an explicit recommendation section per issue and explicitly express the recommendation using the paragraph I quoted. I think they could have put any scope in that paragraph that they felt would address the potential issue.

"At best, this pull request solved a lingering security issue." -- so you have found a security issue that ROS did not articulate?

"At worst, we've simplified the code a bit." -- so apparently by replacing audited code with non audited code we can not introduce any bug.

If you had introduced the change as a simplification it would have been much easier and quicker to inspect and merge. Perhaps we could leave supercilious comments out.

jornane:
> Even when they have a well structured document with an explicit recommendation section per issue and explicitly express the recommendation using the paragraph I quoted.

Is your document missing the paragraph I quoted? Maybe you did get a different version than I did..? In the document that I got, both paragraphs (the one you quoted and the one I quoted) are under the "Recommendation" heading.

> "At best, this pull request solved a lingering security issue." -- so you have found a security issue that ROS did not articulate?

I am not a cryptographer, but introducing complexity in a cryptographic operation MAY break it. Adding HMAC hashing to a random number generator is such complexity. I don't know how it impacts the randomness of the resulting number. What I know from university cryptography classes though is that simple is better. Using random_bytes() is definitively as secure as it's gonna get, but using HMAC hashing MAY shave off some security. Then again, it may not.

> "At worst, we've simplified the code a bit." -- so apparently by replacing audited code with non audited code we can not introduce any bug.

Then don't merge it. You're the boss here.

PR #158, opened 2017-10-13
Remove instanceUID and unnecessary GET parameters

jornane:
Utilities::runningInstanceUID() returns a semi-unique identifier for the current FileSender instance.
This identifier is generated the first time it is requested and then stored in the tmp directory in FileSender.
The identifier doesn't actually provide any useful function, it's only used as a GET parameter for the TeraSender JS file,
but that file is static and doesn't parse the GET parameter.

My guess is that this is a way to defeat cache, but that shouldn't be necessary.
Any modern browser will follow the caching rules set by the webserver, and caching isn't done for POST and PUT requests anyway.

I can only assume that the code removed here was written with still-to-be-added functionality in mind,
but it was never added.  I don't know what this code is supposed to do, but I can say for sure that - right now - it doesn't do anything.


monkeyiq:
To double check, you have tested that TeraSender works fine with the changes? IIRC you are not using TS so I thought I'd check before merge.

monkeyiq:
I would have assumed testing but https://github.com/filesender/filesender/pull/160#issuecomment-340683051 makes me err on the side of asking first.

jornane:
I don't use TeraSender **in production**. It works fine on a test instance. Furthermore, a simple grep through the code shows that this `v=` variable really is unused.

monkeyiq:
This seems to have all been integrated into the codebase with the last bit in https://github.com/filesender/filesender/pull/474

PR #159, opened 2017-10-13
Stop locking files on write

jornane:
This may seem like a silly thing to do, but it's not.

FileSender will never attempt to write overlapping chunks, so there is no risk of conflict.
However, locking may cause tremendous overhead in terms of time, especially when using an NFS backend.

### DIFF
--- a/classes/storage/StorageFilesystem.class.php
+++ b/classes/storage/StorageFilesystem.class.php
@@ -337,27 +337,23 @@ class StorageFilesystem {
         // Open file for writing
         $mode = file_exists($file_path) ? 'rb+' : 'wb+'; // Create file if it does not exist
         if($fh = fopen($file_path, $mode)) {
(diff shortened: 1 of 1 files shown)


monkeyiq:
This is an over statement: "FileSender will never attempt to write overlapping chunks". The Rest client code would happily allow one to do this if they wanted. The user shouldn't want to do this as it doesn't make sense, but FileSender would certainly attempt to write overlapping chunks if the case was presented.

I will probably be merging this soon.

jornane:
> This is an over statement
I should have provided some context there. Chuncks are not overlapping so locking doesn't make sense; there won't be any transaction where a race condition could occur that could be prevented by locking.

In the context of locking files, FileSender will never write overlapping chuncks. But it might write a chunck twice, locking doesn't fix that.

PR #160, opened 2017-10-13
Remove redundant filename check

jornane:
As part of the effort to remedy FIS-007 (PR#64), a filename check was introduced.
However, another filename check was already in place.  The new filename check is better,
because it uses the configuration setting to decide which filename is good and which isn't.

This commit removes the old filename check.

### DIFF
--- a/www/js/transfer.js
+++ b/www/js/transfer.js
@@ -169,12 +169,6 @@ window.filesender.transfer = function() {
             return false;
         }
 
-        
-        if (!/^[^\\\/:;\*\?\"<>|]+(\.[^\\\/:;\*\?\"<>|]+)*$/.test(file.name)) {
-            errorhandler({message: 'invalid_file_name', details: {filename: file.name}});
-            return false;
-        }
-        
         if (!file.size) {
             errorhandler({message: 'empty_file'});
             return false;


monkeyiq:
I assume you have checked that the default values for the new regex cover the old check too.

jornane:
I didn't, but this check is client-side only. So if it is a security risk not to have this check, then that risk was already present before this change.

monkeyiq:
I agree that security checks on client side are to be considered "optional" and offer no security gains.

On the other hand, for non nefarious users the client checks allow bad file names to be noticed quicker. Thus having a different collection of checks client and server side leads to less than optimal user experience.

jornane:
> On the other hand, for non nefarious users the client checks allow bad file names to be noticed quicker. Thus having a different collection of checks client and server side leads to less than optimal user experience.

I agree. As I said:
> However, another filename check was already in place. The new filename check is better,
because it uses the configuration setting to decide which filename is good and which isn't.

The only thing that changes is that, when a filename is correct according to server configuration, but incorrect according to the hardcoded regex, the user would be stopped by the GUI. That code is removed now, so that the server configuration is the only check that's done.

jornane:
Since you seemed spooked (#158) by my statement that the other checks don't check the exact same thing, here's the relevant code that does all necessary checking: https://github.com/jornane/filesender/blob/dbd5cf70a4e6d92b66b159ddf99e635cf8018637/www/js/transfer.js#L188-L210

monkeyiq:
When reviewing a PR I start out assuming that it has come from somebody like me. Obviously things adjust over time from that initial point.

me: "I assume you have checked that the default values for the new regex cover the old check too."
you: "I didn't, but this check is client-side only."

I was looking to see how the user experience changes. Obviously this is client side only and so doesn't provide any form of real security. But if I was removing a check like this I would first like to know how much it would effect user experience. I consider it a fair question to ask of a pull request for one line if the submitter has checked what effect it will have. It is spooky only if the effect has not been considered. In the extreme we could remove all client side checks, user experience would suffer greatly but such checks offer no real security anyway.

jornane:
> When reviewing a PR I start out assuming that it has come from somebody like me.

When reviewing a PR I look at the code it changes.

In the old situation the user would get a filename rejected in the GUI because it matched an arbitrary regular expression. The helpdesk would have a hard time finding out where this error comes from, because the filename is not forbidden in the server-side settings. I fail to recognize how that could be a good user experience.

The check is unnecessary too, as I already stated in my first message "However, another filename check was already in place." Now in retrospect I can see how you could interpret that as a server-side only check, but then I referred you to the actual code where it shows clear as day that the server-side check is also done at the client side. No more ambiguity.

> In the extreme we could remove all client side checks, user experience would suffer greatly but such checks offer no real security anyway.

I never stated that checks should be removed because they don't offer security benefit, I removed an arbitrary check in the client-side code **because a better check is already there**. Please don't put words in my mouth. I mention the lack of security benefit in the old check so nobody has to worry about the security aspects of removing this check.

monkeyiq:
It would seem that you missed the point. I **obviously** have to "I look at the code it changes." to review a PR. I was saying in what proved to be to subtle of a way that I have respect for the person before even looking at the code. I assume if they are changing a single line they have thought about it. Or that they have a well thought out reason for making the PR.

PR #161, opened 2017-10-13
NO MERGE Remove instance id and secret from tmp

jornane:
This removes the Utilities::generateRandomHexString function as well

DO NOT MERGE, blocked by #157 and #158


jornane:
As #157 was rejected, we can't get rid of `generateRandomhexString()`.

Issue #129, opened 2017-09-19
My Transfers page: Move to paginated display


monkeyiq:
While it is handy to be able to pull up all transfers on a single page, it might be less cluttered for users to be able to see the 20-30 most recent transfers and page back to the older ones. 

Bonus points for placing some sort of timeline marker on the page so you can see "these 4 top ones are done in the last 2 weeks" and scrolling down you will start seeing older items which might not be so interesting.

PR #163, opened 2017-10-22
My Transfers page is now paginated

monkeyiq:
DBObject::all() had some support for offset and limit (count) data though it was evaluated client side. Moving that to the db side allows for efficient evaluation with the minor cost that the client doesn't get to know the total number of matches any more.

The admin transfers page is the only client of the old code which has been updated to work with the db side queries. The number of pages is currently only a guess, though the speed of loading should more than make up for that temporary inconvenience.

I used the old fetch one more trick to allow the my transfers page to know if there are more records (a next page) or not. This is why transfers_page selects limit+1. The +1 is not shown to the user and only used to know if there is a next tuple or not.

With the seondary indexes already in place this should make the "My Transfers" page much more useful to medium to heavy system users.


monkeyiq:
once I retest this I will merge tomorrow.

monkeyiq:
https://github.com/filesender/filesender/issues/129

PR #179, opened 2017-11-10
config test: array of array that works on php 5.6 and 7.0

monkeyiq:
The old test works ok in php 5.6 but fails badly in php 7.0


madsi1m:
This fixes the error, cheers

PR #180, opened 2017-11-10
docs: these are some community contributed ideas

monkeyiq:
This started with some ideas that I have had, then I added some from Michiel Leenaars, Peter Schober, and others.


monkeyiq:
yay ci is frozen :/

Issue #138, opened 2017-09-27
Downloading large zip makes zip corrupt


madsi1m:
Hi All,

Found a new (or yet to be discovered old) bug with the downloads page.

To reproduce:
1. Upload several files with a total size of around ~20gb. (uploads fine)
2. on the download page, tick at least 5gb worth and download as a zip file (downloads fine)
3. try unzipping the file. (header of a file in the zip will be corrupt)

Note that downloading the files one by one without the zip seems to be fine.

monkeyiq:
I am trying with 20 files which are 1gb each copied from /dev/zero. I can get 5gb (5 files) fine. The md5sum is the same as they are all zero anyway. Maybe I need to use real data, maybe larger files. I'll keep trying.

monkeyiq:
Excellent, so it happens for files of zero bytes too it seems.
```
$ unzip  transfer.zip 
Archive:  transfer.zip
 extracting: gb8                     
 extracting: gb7                     
 extracting: gb6                     
 extracting: gb5                     
 extracting: gb4                     
file #6:  bad zipfile offset (local header sig):  947913089
file #7:  bad zipfile offset (local header sig):  1996489166
file #8:  bad zipfile offset (local header sig):  3045065243
```

TOC seems somewhat ok, extract isn't though...

```
$ unzip -l transfer.zip 
Archive:  transfer.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
1048576000  10-31-2012 15:09   gb8
1048576000  10-31-2012 15:09   gb7
1048576000  10-31-2012 15:09   gb6
1048576000  10-31-2012 15:09   gb5
1048576000  10-31-2012 15:09   gb4
1048576000  10-31-2012 15:09   gb3
1048576000  10-31-2012 15:09   gb2
1048576000  10-31-2012 15:09   gb1
---------                     -------
8388608000                     8 files
```

monkeyiq:
This appears to be a bug in the zip64 handling code. It will expose for files at offsets over 2^32 bytes in, so you want 5 or 6 one gig files in the zip to trigger it. 

I have tried looking at various extension blocks where the 8 byte offset lives, the best I've made so far is a zip that can expand ok but the tools complain about the format of the headers still.

madsi1m:
Could we try another zip lib? https://github.com/maennchen/ZipStream-PHP might work

monkeyiq:
[...]

PR #181, opened 2017-11-12
zip64 working for files greater than 5gb

monkeyiq:
This fixes an issue with downloading multiple files as a zip where files over the 4gb (32 bit) offset were not able to be extracted before https://github.com/filesender/filesender/issues/138.

### DIFF
--- a/classes/autoload.php
+++ b/classes/autoload.php
@@ -81,6 +81,8 @@ class Autoloader {
 
         'TestDatabase*' => 'test/database/',
 
+        'PHPZipStreamer*' => '/PHPZipStreamer/src/',
+
         // Must be last
         '*' => 'utils/' 
     );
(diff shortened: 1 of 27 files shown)


monkeyiq:
The really good part is that the way the code is makes it easy to switch to other libraries if desired because the data comes from what is essentially a virtual filesystem implemented inside php :) Without the stream wrapper stuff we would have no zero copy to stream data to the zip library, which would be *really, really bad* for large files.

It turns out the CI fails for the PR on one test because the CI itself doesn't support zip64. It worked before because when filesender made small zips it was not making zip64 files. Probably a bad idea because it creates an implementation defined split. For a project designed for large files we probably want to just use zip64 all the time anyway.

I have merged this with that one test disabled and will create and merge an update to bring back the test later. I know disabling a test to merge is 'bad form' but having zips working for the code in master is higher value than having one non zip64 CI test working before merge. Users on sites are expecting to be able to download 5+gb zips and have not been pleased to discover that they can not extract the result.

PR #188, opened 2017-11-28
db: column names are all lower case for easier cross db interaction.

monkeyiq:
oops, left one :/


monkeyiq:
yes, I'll get rid of these merge remote-tracking things very soon.

PR #190, opened 2017-12-04
Replace random generator with a better one

jornane:
Same rationale as #157.

Problems with the current random generator:

- It's overly complicated, due to an old, no longer needed, HMAC calculation which was supposed to mitigate the use of `mt_rand()`.
- It doesn't generate valid UUIDs (version number is random, but should be set to version 4, aka [random](https://en.wikipedia.org/wiki/UUID#Version_4_(random))).
- The password generator for file encryption uses a random generator that is believed not to be very random (`Math.random()`) (fixed since [545b987])

The following indirectly related problem is also solved:

- On first use, FileSender will generate an **Instance UID** for the running instance. This ID is communicated to the client and sometimes used to generate URLs. However, this ID is never actually used for anything. On static files, the GET parameter `_` is used, but static files don't care about GET parameters. On FileSender itself, the GET parameter `v` is used, but there is no reference to `$_GET['v']` or anything similar in the code. Also UNINETT has been running without the Instance UID for multiple months now without that causing any issues (empirical evidence).
- The complicated random generator writes an instance secret which it uses as HMAC key. This is not needed and doesn't add any security. Any code needing this secret file is removed.

SURFnet will release a report written by a cryptographer. This report will back up the cryptographic claims in this pull request. I am not a cryptographer myself, but I have some basic understanding of cryptography.


jornane:
Credits to @SURFnet who paid for this PR and credits to @fkooman for his help in finding these issues.

jornane:
I've updated this PR to fit with the crypto changes that have been made the last few weeks. Honestly, I'm a bit surprised to see this hadn't been merged yet.

jornane:
Has some overlap with #158, which I now see hasn't been merged either.

monkeyiq:
It would be wonderful if I could get a copy of the SURFnet report you mention.

jornane:
> It would be wonderful if I could get a copy of the SURFnet report you mention.

I'm surprised you haven't received it. @meijer can you make sure Ben gets a copy of the report?

monkeyiq:
Though I have not received the report.
I have updated this code and merged it in https://github.com/filesender/filesender/pull/445

jornane:
Thanks 👍

PR #192, opened 2017-12-04
NO MERGE Various crypto improvements

jornane:
**WARNING** If you merge this, all existing encrypted files will become unreadable! Explaination follows

The crypto code has a couple of issues:
- [X] Some configurable values are hardcoded, breaking things if you change them from the default
- [X] No key derivation algorithm is used
- [ ] No use of salt in key derivation
- [ ] Encryption is done blockwise, with a new encryption call per block
- [X] Code hard to read because of few comments
- [X] Hard to debug because of errornous message in integrity check
- [ ] Allows offline attacks

Most of the issues have been fixed here, but some are not. Here follows a detailed explanation of the changes made, and what is still missing. Please note that I'm not a cryptographer, but SURFnet will issue an official report from a cryptographer, which will comment these changes.

For every chapter I have made a checklist of things that would be "nice to have", but I was not able to check everything off the list. The report from SURFnet will comment on these.

# 🗝 Key derivation

When encrypting data using a password, you need to convert the password to a key, as encryption/decryption algorithms need a fixed-length key. The naive implementation (as FileSender used) is to hash the password and use the hash as a key. This allows for dictionary attacks, because since there are few passwords, there are few possible keys.

The better solution is to use a key derivation algorithm. In this pull request, I'll use [PBKDF2](https://en.wikipedia.org/wiki/PBKDF2), which takes a **password**, **hashing function**, **salt** and an amount of **iterations**. It will then produce a key. The problem is that the salt should be random for each file, but it must be stored somewhere. Since I have nowhere to store it right now, I was not able to use a random salt. Because of this, **the code in this pull request does not use a salt**, but it should (#195).

- [X] Start using key derivation
- [X] Make derivation parameters configurable
- [ ] Use salt
- [ ] Store used salt alongside file
- [ ] Store parameters alongside file, so an admin can safely change them for newer files
- [ ] Backwards compatibility for decryption keys derived without PBKDF2

## ⚠️ Important

Because the key derivation procedure changed in this pull request, **_no encrypted files uploaded by the old code will be readable by the new code._** I strongly recommend that **encryption algorithm** and **hashing algorithm** will be stored alongside encrypted files, so that future changes can be made without breaking all existing encrypted files. This may be solved in the same fashion as #136.

# 📦 Chunked encryption

As most will know, FileSender chops files up in chunks, which are 5MB size blocks of files. When encrypting files, FileSender will encrypt each block with its own initialization vector (IV). The IV is then prepended to the block (growing the block with the size of the IV, 16 bytes in the old code, 32 bytes in the new code), growing the block with this amount of bytes. Additionally, the `AES-CBC` and `AES-GCM` encryption algorithms will add a checksum to the file, to allow the decryption algorithm to decide whether decryption has succeeded. This grows the chunk with (up to) an additional 16 bytes.

From a cryptographic standpoint, the data is chopped up in smaller chunks (blocks) less than 1K, which all are encrypted. For the first blocks ciphertext, the IV is used together with the derived key to encrypt the plaintext. For the second block and thereafter, the previous ciphertext is used as "IV", together with the derived key it will encrypt the plaintext to the ciphertext. By having FileSender chop up files in 5MB chunks, we have to make a new IV for every chunk. I'm not currently aware of a way to fix this, as the WebCrypto library only allows for encrypting strings and not streams. Doing large files at once could consume too much memory.

- [X] Support both `AES-CBC` and `AES-GCM` block ciphers
- [X] Make block ciphers configurable
- [ ] Store block cipher configuration alongside file, so an admin can safely change them for newer files
- [ ] Sanity check for block cipher configuration instead of silent crash
- [ ] Encrypt whole file without new IV every 5MB

# 🕵️ Offline attacks

When a file is uploaded to FileSender using encryption, it is still possible for anybody who has the download link to download the encrypted file, and start brute-forcing it.  This is not really trivial to solve, as the server is supposed to never see the password, so the user must not send the password to the server.

A possible solution is that the uploader must encrypt a nonce known to the server, and upload that alongside the file. The server can then provide the downloader with the encrypted nonce, and only release the encrypted file if the downloader can proof ownership of it by presenting an unencrypted nonce.

- [ ] Store encrypted and unencrypted nonce alongside the file
- [ ] Prevent downloading the file without presenting the unencrypted nonce

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -746,11 +746,20 @@ class Transfer extends DBObject {
 
         // padding on the last chunk of the file
         // may not be a full chunk so need to calculate
-        $lastChunkPadding = 16 - $lastChunkSize % 16;
-        if ($lastChunkPadding == 0)
-            $lastChunkPadding = 16;
-            
-        return $file->size + ($chunksMinusOne * $echunkdiff) + $lastChunkPadding + 16;
+        switch(Config::get('crypto_crypt_name')) {
+            case 'AES-CBC': {
+                $lastChunkPadding = 16 - $lastChunkSize % 16;
+                if ($lastChunkPadding == 0)
+                    $lastChunkPadding = 16;
+                break;
+            }
+            case 'AES-GCM': {
+                $lastChunkPadding = 16;
+                break;
+            }
+            default: throw new ConfigBadParameterException('crypto_crypt_name');
+        }
+        return $file->size + ($chunksMinusOne * $echunkdiff) + $lastChunkPadding + Config::get('crypto_iv_len');
     }
 
     /**
--- a/classes/rest/endpoints/RestEndpointFile.class.php
+++ b/classes/rest/endpoints/RestEndpointFile.class.php
@@ -284,9 +284,19 @@ class RestEndpointFile extends RestEndpoint {
                 // Calculate the correct length
                 $chunkLength = strlen($data);
                 // The encryption adds padding and a checksum
-                $paddedLength = 16 - $client['X-Filesender-Chunk-Size'] % 16;
-                if ($paddedLength == 0) {
-                    $paddedLength = 16;
+                switch(Config::get('crypto_crypt_name')) {
+                    case 'AES-CBC': {
+                        $paddedLength = 16 - $client['X-Filesender-Chunk-Size'] % 16;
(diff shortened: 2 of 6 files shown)


monkeyiq:
since these are old and "no merge" I am closing them for now.

PR #194, opened 2017-12-04
NO MERGE Security analysis

jornane:
This PR combines #190 and #192. It's used as a single commit that will receive a security audit via SURFnet.


monkeyiq:
since these are old and "no merge" I am closing them for now.

PR #196, opened 2017-12-05
Performance improvements

Veredior:
Several minor optimizations which should be compatible with PHP 5.3.


.gitignore
```diff
@@ -10,3 +10,4 @@
 /www/skin
 /vendor/
 /testmails
+/.idea
```
jornane: I think you should add this to your own global .gitignore.

classes/utils/Utilities.class.php
```diff
@@ -385,7 +385,7 @@ public static function sanitizeInput($input) {
             
             // Render potential tags useless by putting a space immediatelly after < which does not already have one
             $input = html_entity_decode($input, ENT_QUOTES, 'UTF-8');
-            $input = preg_replace('`<([^\s])`', '< $1', $input);
+            $input = preg_replace('`<([^\S])`', '< $1', $input);
```
jornane: Are you sure this is correct? I think changing `\s` to `\S` switches around the meaning, doesn't it? Maybe you wanted to change `[^\s]` to `\S`?

jornane:
There are some really good catches in here. I did make some inline comments.

.gitignore
jornane: See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

classes/utils/Utilities.class.php
Veredior: indeed the change of \s to \S change is a mistake :)

Veredior:
I'm reverting all the regex changes as \d includes other valid numbers like ۰߀०০੦૦୦௦౦೦൦๐໐０

classes/data/Transfer.class.php
```diff
@@ -227,7 +227,7 @@ public static function fromGuestsOf($user) {
         // Gather user's guests transfers
         $transfers = array();
         foreach(Guest::fromUser($user) as $gv) {
-            $transfers = array_merge($transfers, $gv->transfers);
+	        $gv->transfers += $transfers;
```
jornane: Does this work? The following test code didn't work for me:

	<?php
	$a = array('a');
	$b = array('b');
	$b += $a;
	var_export($b);

Yields

	array (
	  0 => 'b',
	)

.gitignore
```diff
@@ -9,4 +9,4 @@
 /www/doc
 /www/skin
 /vendor/
-/testmails
+/testmails
```
jornane: You still have a modification here, you removed the tailing newline that was there before. :)

classes/utils/DBI.class.php
```diff
@@ -293,7 +293,7 @@ public function __call($method, $args) {
             foreach($args[0] as $key => $value) {
                 if(is_array($value)) {
                     $values = array_values($value);
-                    for($i=0; $i<count($values); $i++)
+                    for($i=0, $iMax = count( $values ); $i < $iMax; $i++)
                         $args[0][$key.'___'.$i] = $values[$i];
```
jornane: Since `array_values` is used, I think you may simplify this even more with:

	foreach($values as $i => $value)
		$args[0][$key.'___'.$i] = $value;

classes/data/Transfer.class.php
monkeyiq: I notice the same thing. There is a subtle difference between using += and array_merge() which might not be what we are expecting in filesender in general.

If you have identical keys in two arrays of length 3 then array_merge will preserve all elements and change the keys to work. += will retain only the key-values from one array and ignore the other. It seems there is less possible surprise using array_merge() especially when the indexes 

```
php >  $a = array(1,2,3);
php > $b = array(10,11,12);
php > $c = array_merge($a,$b);
php > var_dump($c);
array(6) {
  [0]=>
  int(1)
  [1]=>
  int(2)
  [2]=>
  int(3)
  [3]=>
  int(10)
  [4]=>
  int(11)
  [5]=>
  int(12)
}
php > $d = $a;
php > $d += $b;
php > var_dump($d);
array(3) {
  [0]=>
  int(1)
  [1]=>
  int(2)
  [2]=>
  int(3)
}
```

In this case $gv->transfers is from Transfer::fromGuest() which is from self::all() which uses the primary key as the array key. So, without having tested it I think it might be ok. Though still might not be worth it to have to verify each time instead of just using array_merge() which does what one would expect.

monkeyiq:
I will have to take another look, but I think I was happy to merge this without the array_merge() change.
https://github.com/filesender/filesender/pull/196/files/41f31665a283eadd25ee070602d6ac0f0967aeee#r160045685

monkeyiq:
I might even push a revert of that one line and roll with it from there.

PR #198, opened 2017-12-07
Prevent Internet Explorer security settings from crashing FileSender

jornane:
Internet Explorer may throw an exception when trying to use localStorage while it is forbidden for security reasons. Returning false will keep FileSender running, just disabling features that Internet Explorer won't let us use.


monkeyiq:
I will merge this once CI passes & I return to keyboard.

Issue #197, opened 2017-12-07
Better config sanity checking


jornane:
This is pretty much a recap of #49, but it's something that I got into a while back.
When an invalid configuration option is mentioned in the config file, no error is thrown until the user tried to do something.

How to reproduce:
1. Add an invalid option:

		$config['guest_options'] = array(
			// […]
			'yolo_swag' => array(
				'available' => true,
				'advanced' => true,
				'default' => true
			),
			// […]
		)

2. Try to make a guest voucher

You will notice that you don't see an error before you actually click the «Send Voucher» button.

Expected behaviour: You get an error message when opening the "Guest" page instead of after filling it out.

jornane:
Fixed in 23ed95a, thanks!

PR #203, opened 2017-12-14
Config validation: fail very fast for bad configuration settings in guest

monkeyiq:
As has been reported in issue 197, if you have a configuration for
guest_options that uses an invalid key then you will have only
discovered this when a guest voucher sent out is attempted.

This PR moves the detection of an invalid option to a top level event.
You will not see a filesender website if there is an invalid setting
but will get a log message telling you about this, what key is
offending, and what the possible keys are in case you have mistyped
something.


monkeyiq:
This should fix https://github.com/filesender/filesender/issues/197 and lay some ground for other fail fast misconfiguration detection code.

Issue #88, opened 2017-06-29
cronjob problem


keros:
Hi

I tried to run:
/path/to/webspace/scripts/task/cron.php

But it returns:
...
PHP Fatal error:  require_once(): Failed opening required '/path/to/webspace/lib/dompdf/dompdf_config.inc.php' ...
...

The file "/path/to/webspace/lib/dompdf/dompdf_config.inc.php" does not exist (not in my webspace and not on git).
Looks like it was removed with #61

keros:
Strange, I got this error 4 times in a row and now its gone (seems cron does not run the complete code every time?)
The full error was:

```
PHP Warning:  require_once(/path/to/webspace/lib/dompdf/dompdf_config.inc.php): failed to open stream: No such file or directory in /path/to/webspace/classes/utils/Report.class.php on line 38
PHP Fatal error:  require_once(): Failed opening required '/path/to/webspace/lib/dompdf/dompdf_config.inc.php' (include_path='/path/to/webspace/simplesaml/vendor/openid/php-openid:.:/usr/share/php') in /path/to/webspace/classes/utils/Report.class.php on line 38
```

keros:
I got a new problem after updateing to 2.6 (92813f2087824a686c737f5a4804a678ba740c5c)
sudo -u php-user php /data/www/filesender.mydomain.com/httpdocs/scripts/task/cron.php
```
PHP Fatal error:  Uncaught PropertyAccessException: no_such_property in /data/www/filesender.mydomain.com/httpdocs/classes/data/StatLog.class.php:365
Stack trace:
#0 /data/www/filesender.mydomain.com/httpdocs/classes/data/DBObject.class.php(558): StatLog->__get('browser')
#1 /data/www/filesender.mydomain.com/httpdocs/classes/data/StatLog.class.php(166): DBObject->toDBData()
#2 /data/www/filesender.mydomain.com/httpdocs/classes/data/StatLog.class.php(299): StatLog->save()
#3 /data/www/filesender.mydomain.com/httpdocs/scripts/task/cron.php(52): StatLog::createGlobal('global_storage_...', 11256791040)
#4 {main}
  thrown in /data/www/filesender.mydomain.com/httpdocs/classes/data/StatLog.class.php on line 365
```

monkeyiq:
It seems teh issue 557 is fixed and this is perhaps resolved now too. Please reopen (or report a new if the issue is different) if that is not the case.

PR #204, opened 2017-12-15
cron: cleanup for old pdf config file

monkeyiq:


monkeyiq:
This should fix https://github.com/filesender/filesender/issues/88

Issue #189, opened 2017-11-29
sometimes Transfer->makeAvailable() is called before all Event#file_uploaded are called


madsi1m:
Sometimes Transfer->makeAvailable() is called before all Event#file_uploaded are called.

$config['terasender_enabled'] = true;
$config['terasender_worker_count'] = 4;
$config['terasender_start_mode'] = 'multiple';

madsi1m:
I do not seem to have this problem if

$config['terasender_start_mode'] = 'single';

jornane:
I have the same problem. In very rare cases this can even happen without TeraSender. I can only assume that it's filesystem-specific.

When I say "In very rare cases" I mean that I've never reproduced it myself, but I do occasionally get a bug report from one of our users that it has happened again.

madsi1m:
first glance i could not see where makeAvailable is being triggered, or why it is triggered early. Its possible the logic is a bit wrong, from the logs, it "looks" like it triggers once the last chunk of the last file is uploaded, rather than all the chunks from all the files. when the error pops up and i leave the web page idle it still goes on to finish the transfer but makeavailable is not called a 2nd time

jornane:
I just found out that this issue is even more serious when using encryption in combination with TeraSender.

monkeyiq:
@madsi1m to reproduce this are you just brute forcing batches of 6 or so files until it triggers?

monkeyiq:
I assume the message is
```
Transfer's files are not done uploading
When reporting this error please give the following code to help the support finding out details 
```

madsi1m:
I usually just drop in 10 files I think my settings are to do 4 parallels

On 15 Dec 2017, at 21:28, monkeyiq <notifications@github.com<mailto:notifications@github.com>> wrote:


@madsi1m<https://github.com/madsi1m> to reproduce this are you just brute forcing batches of 6 or so files until it triggers?

monkeyiq:
This is targeted for the weekend :)

monkeyiq:
@jornane I haven't dug into the non TeraSender code searching on this bug yet. That will be soon :)

monkeyiq:
[...]

PR #206, opened 2017-12-17
Race condition of file.complete and transaction.complete

monkeyiq:
The patch is mostly comment, better to be informative there than here.

The basic version is that there was a race between calling
file.complete and transaction.complete. Logging can show that for the
final two chunks nothing was explicitly stopping transaction.complete
from landing before the last chunk of the last file. A test was
already stopping transaction.complete from being called more than
once. There are multiple timeout delays involved in eventually calling
transaction.complete so the chances of that happening on a LAN were
less. Normally on a LAN the workers would themselves complete and call
file.complete before the timeouts ran and the already scheduled
transaction.complete would actually call the server.

This patch makes it explicit that transaction.complete is only
scheduled for the final chunk uploaded on the final file.

So I guess like many races, the finding and reasoning largely
outweighs the size of the code patch required to mitigate the
scenario.

### DIFF
--- a/www/lib/terasender/terasender.js
+++ b/www/lib/terasender/terasender.js
@@ -227,6 +227,12 @@ window.filesender.terasender = {
         var workers_on_same_file = false;
         var min_offset = file.uploaded;
         var fine_progress = 0;
+
+        var uploading_count = 0;
+        for(var i=0; i<this.workers.length; i++) {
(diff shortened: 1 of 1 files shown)


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/189

monkeyiq:
yay CI is stalled.

PR #209, opened 2017-12-23
Stall Detection: avoid false positives

monkeyiq:
It doesn't make much sense to compare the number of chunks that one worker
has uploaded relative to the others. This number will rely heavily on the
computer doing completely fair scheduling which there is no guarantee of.

Instead we are probably better off comparing a worker only to itself.
If it varies too much from it's past performance then we might have a
problem. If we are only starting an upload then a worker will not have
much history and so comparing to average of past chunk times will be a
little more prone to error. So there is more scope in there to be late
if we only have 1,2,3 chunk history as the worker is not established
yet and we don't have a good estimate for the mean for a chunk upload
for the worker.

This should hopefully avoid false positives for stall detection.

### DIFF
--- a/www/js/transfer.js
+++ b/www/js/transfer.js
@@ -82,6 +82,12 @@ window.filesender.transfer = function() {
         
         /**
          * Max duration divergence (the maximum ratio between a process's current upload time and the global average)
+         * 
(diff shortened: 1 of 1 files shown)


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/113

PR #210, opened 2017-12-27
New Feature, StorageFilesystemExternal allows for filesystem function…

madsi1m:
…s to be run by an external script, rather than by php itself


classes/storage/StorageFilesystemExternal.class.php
```diff
@@ -0,0 +1,207 @@
...
+    public static function readChunk(File $file, $offset, $length) {
+
+	if ($file->transfer->options['encryption'])
+	    $offset=$offset/Config::get('upload_chunk_size')*Config::get('upload_crypted_chunk_size');
+
+	$out=self::run('fs_readChunk "'.$file->uid.'" '.$offset.' '.$length);
+	if ($out['status']!=0) {
+		//something bad
+        	//if(!file_exists($chunkFile))
+	        //    throw new StorageFilesystemFileNotFoundException($file_path, $file);
```
jornane: Shouldn't this cause an error?

madsi1m: You are right, will fix

monkeyiq:
It might be nice for the StorageFilesystemExternalStream::run() to be removed and the StorageFilesystemExternal::run called in it's place. These two runs seem the same right now.

monkeyiq:
I think I'll bring this in with the run() merge as a post option.

Issue #212, opened 2018-01-04
saml: failing to find simplesamlphp does not give good error message


monkeyiq:
During testing I forced the following config to a wrong location

```
$config['auth_sp_saml_simplesamlphp_location']
```

This fails to load but does not produce a useful error message. This is used in loadSimpleSAML() in a require_once() call which stops the script on error. Perhaps readability and other tests should be made first and verbose error messages created to help correct this issue during setup.

PR #213, opened 2018-01-04
saml: better reporting for things going wrong in setup

monkeyiq:
This offers some more information in the log if the saml files are not found where FileSender is expecting them to be or for some reason they are not readable (permissions, selinux).


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/212

PR #215, opened 2018-01-07
externalStorage: share the run() and streams update

monkeyiq:
The stream class is updated to avoid a name clash with the Chunked stream.


monkeyiq:
This is a small update to https://github.com/filesender/filesender/pull/210

Issue #80, opened 2017-06-13
crypto & resume


efef:
While uploading a file encrypted there mustn't be resume functionality.
We've seen that filesender 1.7 is resuming encrypted files and this leads to too many chunks on the server-side and other issues. Unfortunately it even looks like a known issue in FS 1.7:
https://app.assembla.com/spaces/file_sender/tickets/realtime_list?tickets_report_id=filter:u470673&page=1&ticket=853

 Filesender 2.0 has a comparable issue. When uploading a large file (e.g. 100MB) encrypted. Stop the uploading half-way. Send the same file again with same password and the server will probably use 'resume'.
Try to download the file and you will experience that  you'll get a password error.

monkeyiq:
I am hoping to make a smallish test case to replicate and remove the issue. When you stop the uploading half way are you clicking on pause on the web page, leaving the web page by loading some other location, or leaving some other way.

I know the delay here makes thing more difficult.

monkeyiq:
I will also dig into the assembla export to track this in ticket 853 for more details.

monkeyiq:
From ticket 853:
```
<summary>Crypto: how to handle resume ?</summary>
<created-on type="datetime">2012-11-12T14:19:17Z</created-on>
  <description>- check what happens with current codebase
- should we allow it ? 

The risk here is that people enter a different key with the resumed upload.
</description>
```

monkeyiq:
I am assuming this relates to stopping an upload with encryption on and then visiting the upload page again and seeing the "resume upload" dialog offering to setup the transfer for you. When I do this I notice that the files are shown as partially transferred. I have added a configuration option in https://github.com/filesender/filesender/pull/218 which when active will explicitly forget partial transfers instead of offering to resume them when you revisit this page.

I tested just hitting the pause and play button while the transfer is active and that seems to work and also not give you the option to change the password while a temporary pause of the transfer is in effect. The option could be extended to remove the pause button completely from the upload page if encryption is on if desired.

PR #218, opened 2018-01-08
crypto and resume: option to not remember transfer if if crypto is on

monkeyiq:
This only effects when you break an upload and then enter the Upload
page again. Normally a dialog "Restart failed transfer" appears
offering to setup the transfer again and "Restart" instead of start
the transfer.

This could be problematic if the user enters a different password for
the resume, having part of the upload encoded with one password and
part with another. It was also reported on
https://github.com/filesender/filesender/issues/80 that resumed
crypted uploads were having issues, so until that is tested to a
greater extent it is better to offer the abilitiy to not allow resume
for crypted transfers.


monkeyiq:
I have left the default for this setting to false as it is very easy to update the default when policy is decided.

PR #225, opened 2018-01-17
getting saml version can cause problems in some cases

monkeyiq:
This can cause downloads to fail to complete due to excess RAM usage in some cases.


jornane:
Why is simpleSAMLphp called upon at all when downloading? Wouldn't it be better to avoid calling upon simpleSAMLphp at all when downloading? That would avoid both session locking and the bug you describe here.

Issue #226, opened 2018-01-17
AUP settings broken?


fransward:
In the previous 1.6/1.7 release of Filesender the AUP function was working and according to the documentation is still supported, but I can't get it to work.

[](https://github.com/filesender/filesender/blob/master/docs/v2.0/admin/configuration/index.md#aupdefault:)

> aup_default
>     description: if set to 'true' the AuP (terms of service) checkbox (if enabled) is already pre-ticked for the user
> 
> aup_enabled
>     description: If set to 'true' the AuP (terms of service) checkbox is visible AND mandatory for the user to tick.

Our config:
```
$config['aup_enabled'] = true;
$config['aup_default'] = false;       // AuP value is NOT already ticked
```
But no acceptable use policy tickbox appears....

monkeyiq:
This seems to be the same issue reported by William van Santen on the filesender-dev mailing list. 
```
* In filesender 1.6 there is the possibility for a checkbox for accepting 
the terms and conditions of the service. We don't see this in 2.0:
```

PR #234, opened 2018-01-29
AUP: bring back the checkmark to bind users to the site terms and conditions

monkeyiq:
There seemed to be some partial renaming going on there. This should make the T&C checkmark work again when you have
```
$config['aup_enabled'] = true;
$config['aup_default'] = false;       // AuP value is NOT already ticked
```


monkeyiq:
This should fix https://github.com/filesender/filesender/issues/226

fransward:
It did. Thanks for the fix!

Issue #229, opened 2018-01-28
Limits: ability to set maximum file size for encrypted files


monkeyiq:
It is not possible to set the maximum encryption file size

Feature request: possibility to configure the maximum file size for files a user want to encrypt. Our test (mac os with firefox) result in failed file decryption for files with a filesize from 2GB and up.

This was originally reported by William van Santen on the filesender-dev mailing list.

monkeyiq:
I split this out with the "maximum encryption file size" part here and support for encrypted files over 2gb issue being in https://github.com/filesender/filesender/issues/230

fransward:
Check! I've set the max_transfer_encrypted_file_size at 2GB:
`$config['max_transfer_encrypted_file_size'] = 1 << 31 ; 2^31 bytes = 2GB`

PR #235, opened 2018-01-29
Limits: ability to set maximum file size for encrypted files

monkeyiq:
This is checked client and server side. One little twist on the client
side is that enabling and disabling encryption will have a toggle
effect on if files are valid.

The code sort of assumes that if the client side transaction.addFile()
doesn't work then it will not magically start to work. The max file
size stuff toggles so that you can enable encryption and see an error
for file too large, but if you then disable encryption again you
should see that the file is ok again.


monkeyiq:
This should address https://github.com/filesender/filesender/issues/229

Issue #232, opened 2018-01-28
ui:  The download page has different fontsizes


monkeyiq:
The download page has different fontsizes, this looks very sloppy:

<img width="728" alt="mjeoafgjlgpgikae" src="https://user-images.githubusercontent.com/387108/35488840-995c9dbc-04da-11e8-87d4-a33d3814d8ac.png">

This was originally reported by William van Santen on the filesender-dev mailing list.

PR #237, opened 2018-01-30
2018/download page font cleanup

monkeyiq:
this also slipstreams some other clutter on the page.


monkeyiq:
This should address https://github.com/filesender/filesender/issues/232

Issue #228, opened 2018-01-28
Stop button should pause upload when dialog is shown


monkeyiq:
During uploading a file: When you press the STOP button, there appears a Confirmation Window. Filesender asks you: "Do you really want to stop the upload and remove already uploaded data ?"

If you don't respond to the question the upload continues. 

Feature request: When you press the STOP button, the upload of your file(s) pauses, and there appears a Confirmation Window. 

This was originally reported by William van Santen on the filesender-dev mailing list.

PR #238, opened 2018-01-30
Pause uploading if the stop button is pressed.

monkeyiq:
If the user cancels that dialog (cancel the stop) then start uploading
again too.


monkeyiq:
This should resolve https://github.com/filesender/filesender/issues/228

Issue #242, opened 2018-02-01
encryption password: ability to set generated_min_length > password_min_length


monkeyiq:
https://github.com/filesender/filesender/issues/231 introduced the ability to set a minimal password length for encryption. As part of 231 the generated password length was extended to be the minimal password length.

It might be nice for the generated password length to be able to be set distinctly to the minimal password length. 

For example, to set minimal password length for users is for example 10 (for usability), but if you let FileSender generate your password using the 'Generate Password' option it is a different length (but never shorter then the minimal password length). For example, to have FileSender generate a safer password with a length of 20.

PR #244, opened 2018-02-01
encryption password: allow generated password to be longer

monkeyiq:


monkeyiq:
This resolves https://github.com/filesender/filesender/issues/242

PR #255, opened 2018-02-15
cloud: amazon s3 storage backend

monkeyiq:
This has not been tested against a real S3 server. It works against Zenko CloudServer.

See http://docs.filesender.org/v2.0/cloud/ for information on setup.

### DIFF
--- a/classes/storage/StorageCloudS3.class.php
+++ b/classes/storage/StorageCloudS3.class.php
@@ -1,0 +1,231 @@
+<?php
+/*
+ * Store the file as chunks instead of as a single file on disk.
+ *
+ * FileSender www.filesender.org
+ *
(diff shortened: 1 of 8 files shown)


kdermody:
Tested against real S3 and uploaded successfully

some initial reactions:

- Setting cloud_s3_region in config.php or in ConfigDefaults.php doesn't seem to work. Always uploads to us-east-1
- Setting the cloud_s3_endpoint to include a region (e.g. s3.eu-west-1.amazonaws.com) throws errors and must be set to s3.amazonaws.com
- Each upload creates it's own bucket, might be nice to have option to specify a default bucket and have files in some kind of sub folder.
- Performance seems unusually slow, averaging 3MB/s, not sure if this is just related to region.

Issue #257, opened 2018-02-15
Deleting a transfer does not refresh my transfers tab


monkeyiq:
In the My Transfers tab, Using the trashbin icon results immediately in an email:

```
Dear Sir or Madam,

The transfer n°99 has been deleted from filesender by the sender (sender@filesender.example.com) and is no longer available for download.

Best regards,
filesender2 
```

But the My Transfers tab is not refreshing, it is still listed under the 'Currenty available transfers'. When I try to use the link that to the file that was send when it was uploaded, it takes more than one minute to show an error message:

![img](https://user-images.githubusercontent.com/387108/36256403-82f7c82c-129e-11e8-9455-5c294216ab41.png)

monkeyiq:
I have found that using a very long delay in storage_filesystem_file_deletion_command can make the UI appear to be frozen. This use case is to simulate running a shred(1) as the file_deletion_command on larger files. If you push this to the extreme you can get the dialog showing "The gateway did not receive a timely response from the upstream server or application.".

I am still considering what the solution to this situation should be. From a privacy standpoint it is very desirable to shred files but the current code is not graceful to the timeframes needed to do that.

monkeyiq:
The first drop of code to mitigate this is at https://github.com/filesender/filesender/pull/259

Issue #243, opened 2018-02-01
cancelling an upload takes a long time


monkeyiq:
It seems that when you stop and cancel an upload, it takes a long time before filesender responds.

monkeyiq:
For normal and chunked storage types I am seeing less than a second delay to abort an upload with the stop and then "ok" on the dialog button. Backed by a relatively spartan Postgresql database.

If this is being tested on an existing filesender install, maybe the auditlogs and or statlogs are getting large and something is interacting with those tables?

monkeyiq:
Does it only slow down when you cancel an upload?

In other words, are the following paths faster than canceling the upload:
* upload to completion is ok.
* delete transaction from my transfers page is ok.

monkeyiq:
if the file deletion command was shred(1) then it may take a great amount of time to cancel a transfer as you would have to wait for the shred to complete.

PR #259, opened 2018-02-18
Shredding: new clearing house to shred deleted files

monkeyiq:
Running shred(1) directly in the
storage_filesystem_file_deletion_command can make the Web interface
appear to be broken at times. This is due to the execution of shred
being done synchronously.

This new code creates a clearing house for shredding. When a file is
deleted through the Web interface it is moved and marked for shredding.
See the new docs shredding page at docs/v2.0/shredding/index.md
for details.

There are many design options for this. Having a database table allows
failures to be tracked which might not otherwise be noticed. The
original file names are scrubbed so the files to be shredded does not
expose the user who used to own the files.

### DIFF
--- a/classes/autoload.php
+++ b/classes/autoload.php
@@ -61,6 +61,7 @@ class Autoloader {
         'DBObject' => 'data/',
         'Transfer' => 'data/',
         'File' => 'data/',
+        'ShredFile' => 'data/',
         'Recipient' => 'data/',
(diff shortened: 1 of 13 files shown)


monkeyiq:
This is to help cases like https://github.com/filesender/filesender/issues/257

monkeyiq:
This code should help with https://github.com/filesender/filesender/issues/243

PR #274, opened 2018-03-12
terasender: resume again

monkeyiq:


madsi1m:
pause and resume seems to pause and resume now. but the upload fails sometimes with many files if i pause during upload. refreshing the page and reloading the failed download seems to want to upload the last chunks of most files then upload competes successfully.

PR #298, opened 2018-04-24
Directory tree upload

glbrimhall:
directory-tree-upload feature tested with the StorageFilesystem and StorageFilesystemPreserveName successfully using google chrome. Validated legacy single file upload still works with chrome & firefox.

When a zip is created, it is automatically including the full paths of the uploaded dirtree, which is pretty cool. All email communication sent to recipients lists the files under their full paths.

Refer to filesender-dev posting on deep details of enhancement:

https://sympa.uninett.no/lists/filesender.org/arc/filesender-dev/2018-04/msg00013.html


glbrimhall:
Re-submitted pull request fixing accidental remove of StorageFilesystem::deleteFile, as well as fixed errors reported by Travis wrt default method params not being null

monkeyiq:
This is a great feature, thanks for the PR and the interactions on the mailing list!

One little snag of using poeditor for the translations is that many of the updates changing file.name to file.path in the language/* files will be overridden when the next export from poeditor happens. I have created a little update to help mark those files https://github.com/filesender/filesender/pull/309 to try to help future contributors. I'll have to take a look at poeditor to see if there is a general way to replace these terms in the translations or if that will have to be done by a more manual process.

It seems the only change to the trans files is the file.name becomes file.path.

I have done a database upgrade here and have started testing out the pr locally.

monkeyiq:
It might be an interesting idea to have the database upgrade script also populate the collectiontypes table with the defaults. I'm thinking maybe adding some sort of new method after the secondary indexes are made to allow a class to "do stuff" to the new table if it desires so at https://github.com/filesender/filesender/blob/master/scripts/upgrade/database.php#L149 

Though this is more an idea and less of something that needs to be done before this is merged.

PR #303, opened 2018-05-10
add json logs option

madsi1m:
Gives admins the option to log in json format

### DIFF
--- a/classes/utils/Logger.class.php
+++ b/classes/utils/Logger.class.php
@@ -109,6 +109,9 @@ class Logger {
                 $facility['level'] = LogLevels::INFO;
             }
             
+            if(!array_key_exists('output', $facility))
+                $facility['output'] = 'text';
+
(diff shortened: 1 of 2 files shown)


classes/utils/Logger.class.php
```diff
@@ -324,6 +331,13 @@ public static function log($level, $message) {
                 if(self::$levels[$level] > $max) continue;
             }
             
+            // Build final message ...
+            if ($facility['output'] == 'json') {
+                $message = json_encode($messageArray);
+            } else { //text
+                $message = (array_key_exists('user', $messageArray)?'[user '.$messageArray['user'].'] ':'').'['.$messageArray['process'].':'.$messageArray['level'].'] '.$message;
```
monkeyiq: It might be better to ensure a user key exists and is blank? This one line reads like a bit of a nightmare.

Issue #293, opened 2018-04-10
FileSender can damage a file during upload by forgetting a chunk


jornane:
One of my users complained about a damaged 2GB .WAV file that was sent through FileSender. The report received my inbox with quite a delay, so I was not able to look at the logs and the original file was already expired. The user has kept the file as proof, so he sent the damaged file to me via FileSender, which went OK.

After inspecting the file, I noticed that it has 5242880 bytes of zeroes somewhere in the file, [where 5242880 is the default FileSender chunk size](https://github.com/filesender/filesender/blob/master/includes/ConfigDefaults.php#L90). Because of this, I believe the file was damaged by FileSender. My best guess is that somehow the chunk was not written while a succesful write still was logged in the database.

jornane:
Browser: Windows Internet Explorer 11
TeraSender: off

The transfer was stopped and resumed at least once.

jornane:
We're on commit 1ed00e0, with a backport of 847852b.

meijer:
Jørn, aren't you planning to upgrade to beta-4?

monkeyiq:
I notice an fflush was also removed in https://github.com/filesender/filesender/commit/847852b53c48cbc9fb7746d96d54a5007302ef93 along with the locking.

git master code has checks in there for the return value of fwrite() and logs errors and throws to the client if that fails.

jornane:
> I notice an fflush was also removed in 847852b along with the locking.

That's correct, `fflush` is not necessary because it is done implicitly when closing a file handle ([manpage reference](https://www.freebsd.org/cgi/man.cgi?query=fclose&sektion=3), Linux and MacOS manpages are similar)

Locking is not needed when the offset is known, but when I look at the code now I also see that there is a possibility to omit the offset `is_null($offset)`. I don't know if and where that is used, but I doubt it's relevant for this issue (we would've seen moved chunks instead of missing chunks).

I've considered whether this could be a locking issue, but I don't think it is because of this:
Since TeraSender was not used, there were no concurrent chunk uploads and therefore no concurrent writes to the filesystem.

> git master code has checks in there for the return value of fwrite() and logs errors and throws to the client if that fails.

Yes, that's correct. My user reported that he got a «Resume» prompt halfway through the upload. Is it possible that when the write failed for some reason (speculation on my part), the browser stopped the upload because of the failure, but the upload was not removed from the server, and the failed chunk was attested for in the database?

[...]

PR #304, opened 2018-05-14
baddata: mitigation for issue 293

monkeyiq:
Setting file.uploaded to the end of the chunk before the actual putChunk() was complete could cause a bad upload if there was a forced stall (turn wifi off during upload). This is a non terasender path mitigation.

If file.uploaded is set before the putChunk() is complete then that value is stored into local storage by transfer.reportProgress(), when it calls this.updateFileInRestartTracker(). This is fine as long as file.uploaded is actually the data that *is* uploaded and not what would be uploaded if the currently being processed chunk of data was completed.

So now I have moved setting file.uploaded to when the putChunk() is actually done.

### DIFF
--- a/www/js/transfer.js
+++ b/www/js/transfer.js
@@ -658,6 +658,9 @@ window.filesender.transfer = function() {
         
         // From here we should have everything we need to restart
         filesender.ui.log('Restarting failed transfer #' + this.id);
+        for(var i=0; i<tracker.files.length; i++) {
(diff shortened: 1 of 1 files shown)


monkeyiq:
This is a potential mitigation for https://github.com/filesender/filesender/issues/293

I have produced the issue from 293 locally by disabling wifi during upload (non ts) and restarting. With this code I no longer see the bad data.

PR #307, opened 2018-05-16
Small codestyle improvements (PSR1, PSR2)

FrontEndCoffee:
There were some inconsistencies in the code styling, this PR improves the overall code quality.

The enhancements have been made with the PSR1 and PSR2 standards in mind.


monkeyiq:
This all looks ok from the reasonable chunk I have just skimmed over. 

The reason I ask about the tooling is that one of my reservations is how this will effect the merging of two current PRs https://github.com/filesender/filesender/pull/302 and https://github.com/filesender/filesender/pull/298.

It seems there will be a clash one way or the other (if I merge the above then this PR will likely not merge and if I merge this then those two will likely have problems).

FrontEndCoffee:
It was achieved by running [squizlabs' code beautifier](https://github.com/squizlabs/PHP_CodeSniffer) and [FriendsOfPHP/PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) configured with the PSR1 and PSR2 standards.

If you want I could add a script (e.g. Makefile, composer script, etc.) so that devs can run the code fixers themselves. For example using:
```
composer run style:fix
```
or
```
make style:fix
```

jornane:
> It seems there will be a clash one way or the other (if I merge the above then this PR will likely not merge and if I merge this then those two will likely have problems).

I agree, this is a merge conflict waiting to happen. I think we should prioritize improvements to the application logic over improvements to the coding style of existing code.

Another thing to consider is the Git history: By merging this, 7553 lines of code in this project will be attributed to this PR instead of the original author and the original commit+PR, while this PR doesn't (or at least shouldn't, didn't check everything) make any chances to the application logic. This makes it harder to find out which code was introduced in which commit using tools like [`git blame`](https://www.git-scm.com/docs/git-blame).

Git blame is really useful for finding the context of the changes to the code, e.g. what other changes were introduced at the same time. Thus, this PR can make future debugging sessions harder.

FrontEndCoffee:
@jornane I agree with you, you have a good point. Although I think it is very important to have consistent code. I completely understand it if this PR gets closed, but I think it is important for the project in the long term to have consistent code that has a similar style to other PHP projects. 

This is of course the reason that [these standards (PSR)](https://www.php-fig.org/) were created and the adoption of these standards (by a.o. FriendsOfPHP, PHPLeague) is the reason that the PHP community has survived these past years.

monkeyiq:
I think it is a good idea to document this on docs.filesender.org and make it easy for filesender devs to run the code cleaner which I am happy to help make happen. 

I will check if there is a way to bring forward the original line author information in git because it can be useful to see the provenance of code blocks. 

I know my personal style is slightly different to the PSR which will likely lead to some code bring generated which is not PSR compliant.

monkeyiq:
I think the touching of lines issue is less of one. git blame is quick to invoke but there are better solutions, even for code that has not seen style only changes of this nature. As some rework or modification might change the last person to change a line a more interesting command follows, taking a line or range.

```
$ cat /home/common/bin/git-blame-extended 
#!/bin/bash

FILE=$1
LINE=$2

git log  --pretty=short -u -L $LINE:$FILE
```

There are likely to be many more git invocations along these lines. It is a bit surprising that links to commands of the above format are not made at the outset. I know many developers (myself of recent included) can be pensive about git blame becoming clouded.

monkeyiq:
I have created a small PR to bring in a script that can reformat the code in https://github.com/filesender/filesender/pull/422. I will probably run that on a dedicated repo so the user/email can be set to something to identify future PRs created that are restricted to only formatting the code.

Thanks @FrontEndCoffee for bringing this up. Unfortunately at the time there were a few large PRs awaiting merging such as the dirtree patch.

monkeyiq:
This reformat has been merged in https://github.com/filesender/filesender/pull/429

See https://docs.filesender.org/v2.0/dev/formatting/ for documentation on running the formatter locally if you want to do that. It should be easy with composer and the supplied configuration file.

Issue #305, opened 2018-05-14
TeraSender sometimes has workers hang which causes a bad resume interaction


monkeyiq:
I have seen a case where uploading a 90gb file (IIRC) over a fairly fast WAN (50-100mb/s) a case where some terasender workers seem to never report failure of a chunk and seem to just disappear. 

Consider a case where for example this happens at around 9% of an upload. When the upload fails at 99% (because there are workers that have not completed their chunk) the user can elect to navigate to my transfers and back to upload and is offered the chance to resume the upload. If you add the file and start you see the upload start at the 9% where the first chunk failed instead of the 99% that a user might be expecting. William van Santen had the great insight to produce this explanation of why this might be happening and checked the browser console to confirm when workers were "going away".

This has apparently been a bug that was available in the FileSender release 1.6 or 1.7 days.

I am attempting to track this down by performing large uploads locally, looking for browser combinations that exhibit the bug. I have started on files 5-15gb in size. Firefox on osx, Edge (41.16299.371.0) on win10, Firefox 59+ on Linux. If anybody knows of a browser/platform that does this fairly regularly I would love to know that combination to help somewhat reliably reproduce it so I can remove the bug.

monkeyiq:
Safari 11.1 (13605.1.33.1.2) seems ok after a few 15gb uploads using tersender with 8 workers. This might take a while to get it to trigger though.

jornane:
Related: #35 #187.

I've also written about this on Assembla but I can't find it anymore in the dump. However, I remember a key thing about my testing two years ago:

https://github.com/filesender/filesender/blob/a9b0e1946a4c71aa18fffcb6b6a4ebc64f53f5b6/www/lib/terasender/terasender_worker.js#L224 - That comment is wrong, that code is also run in some error situation, but I never figured out what kind of error it is.

I found this when I uploaded using the newest Firefox on the newest MacOS around two years ago. The console would sometimes display `broken, exiting`, and we can see from the code that this incapacitates that one worker. The upload would seemingly continue, but naturally that one chunk would be missing, preventing the upload from ever completing, and as such being stuck on 99%.

I'm not at all sure how to fix it, but my idea would be to make a worker restart after a few seconds delay when this happens (to prevent the worker from taking up new work in case we're actually dealing with a browser refresh), but under no circumstance should a worker be able to just stop working.

Good to see this issue is being worked on! 👍

monkeyiq:
[...]

PR #308, opened 2018-05-16
terasender: automatically restart uploading a chunk if the previous upload failed 

monkeyiq:
This allows up to a configurable amount of attempts reuploading a chunk before it is considered to have failed. The default is 20 retries, which is per chunk, so should cover most network issues.

There is also a new testing option to simulate a chunk failure on the third chunk up to allowed_retries-1 in order to test the recovery path. This is off by default and enabled through the config so that Travis can possibly enable the code to ensure recovery is working on a CI run.

This should mitigate https://github.com/filesender/filesender/issues/305 and currently will turn the failure of the 20th time a chunk upload is attempted into a user viewable failure message. Better than silent failure as the user does not expect more data to have been transferred than will be shown.

### DIFF
--- a/www/lib/terasender/terasender_worker.js
+++ b/www/lib/terasender/terasender_worker.js
@@ -10,6 +10,7 @@ importScripts(
 	'../../js/crypter/crypto_app.js'
 );
 
+
 var terasender_worker = {
     /**
      * Worker properties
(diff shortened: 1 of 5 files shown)


www/lib/terasender/terasender_worker.js
```diff
@@ -222,8 +229,25 @@ var terasender_worker = {
             
             this.reportDone();
         }else if(xhr.status == 0) { // Request cancelled (browser refresh or such)
```
jornane: Good work! Maybe you should fix the comment as well..

PR #335, opened 2018-06-12
Improve password generator, more bits

### LINKED ISSUE #334: Math.random() is not a CSPRNG
https://github.com/filesender/filesender/blob/1901e8d561220ac4a8ec034e52b7b80e46359867/www/js/upload_page.js#L1018

MDN writes:

> Math.random() does not provide cryptographically secure random numbers. Do not use them 
> for anything related to security. Use the Web Crypto API instead, and more precisely the 
> window.crypto.getRandomValues() method.

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random

jornane:
Fixes #334 

Uninett FileSender has been running with this for a few months already.

### DIFF
--- a/www/js/upload_page.js
+++ b/www/js/upload_page.js
@@ -1015,13 +1015,16 @@ $(function() {
         return false;
     });
     filesender.ui.nodes.encryption.generate.on('click', function() {
-        var genp = function() { return Math.random().toString(36).substr(2, 14); }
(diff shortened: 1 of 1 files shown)


monkeyiq:
It seems that https://github.com/filesender/filesender/pull/336 takes over from this PR

PR #336, opened 2018-06-12
Documentation and configuration of password generation in webcrypto

jornane:
Improved version of #335, this also improves the configuration settings and documentation.

### DIFF
--- a/classes/utils/Config.class.php
+++ b/classes/utils/Config.class.php
@@ -241,10 +241,7 @@ class Config {
             'Maybe you have set $config["log_facilities"] = array("type" => "file",...) instead of $config["log_facilities"] = array(array("type" => "file",...))' );
         }
 
-        if(!self::get('encryption_generated_password_length')) {
-            self::$parameters['encryption_generated_password_length'] = self::get('encryption_min_password_length');
-        }
-        if(self::get('encryption_min_password_length') > self::get('encryption_generated_password_length')) {
+        if(self::get('encryption_min_password_length') > self::get('encryption_generated_password_bits')) {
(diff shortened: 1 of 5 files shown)


monkeyiq:
I think it may be simpler to retain the existing encryption_generated_password_length setting rather than force a config file update to use bits instead. While using bits is more precise it also might lead to a false sense of security with the expectation that 1024 bits of key has a specific meaning.

monkeyiq:
Thanks for making this PR! 

It is great to see an issue https://github.com/filesender/filesender/issues/334 and notice somebody has made a PR to address it already :)

jornane:
I named the configuration variable wrong, should be bytes instead of bits.

As for the `encryption_generated_password_length` option, it was misleading to me, because it is 0 by default.

https://github.com/filesender/filesender/blob/1901e8d561220ac4a8ec034e52b7b80e46359867/includes/ConfigDefaults.php#L96

I've changed the name because `encryption_generated_password_length` is something different than `encryption_generated_password_bits`.

monkeyiq:
I have merged https://github.com/filesender/filesender/pull/337 which takes ideas from this PR and some other thoughts. I thank you for this PR. I thought it would be quicker for us both if I would include this update in 337 along with some other changes.

I wanted to retain encryption_generated_password_length and encryption_min_password_length.
This allows the generated length to be larger than the minimal length and also lets the user edit a generated password while still enforcing the min length constraint. If the encoding is known then the entropy in the generated password is known. Both are fairly low unless the generated length is set high, perhaps on the order of hundreds. So I went with allowing sysadmins to set the number of encoded chars instead of bytes of entropy directly.

As random generation and encoding those bytes is likely to be interesting to other code I have abstracted that to the crypto_app level.

PR #367, opened 2018-07-06
actually use UTF-8 for MySQL/MariaDB

fkooman:
See: https://medium.com/@adamhooper/in-mysql-never-use-utf8-use-utf8mb4-11761243e434

### DIFF
--- a/docs/v2.0/install/index.md
+++ b/docs/v2.0/install/index.md
@@ -471,7 +471,7 @@ Create the filesender database:
 
 ```
 mysql -u root -p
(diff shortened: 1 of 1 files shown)


monkeyiq:
This looks like the right move. I plan to verify an install before merging. It seems that the old issue of varchar-255 and 4 byte encodings running into a 768 byte boundary has been mitigated since issue https://github.com/filesender/filesender/issues/25

For me on mariadb I am able to make up to 3072 bytes (= div 4 char) indexes now. Though maybe the default indexing has changed since last year.

PR #370, opened 2018-07-12
Issue 342 fix

glbrimhall:
Fixed issue#342 - StoragePreserveName not storing dirtree files under UID=dirtree by adding back loading of CollectionTree's fileCache if it is null. This scenario happens when uploading files, where Collection::fromTransfer is not automatically called which would have automatically loaded the fileCache.


monkeyiq:
I am happy to merge this but the changes to the lang files will get overridden when an export from poeditor.com takes place. The project is currently using poeditor as the gold source of translation strings, see for example, https://docs.filesender.org/v2.0/i38n/

It doesn't seen that you are introducing new terms in the core code changes. Would you be happy with just the changes to the non lang files being merged?

monkeyiq:
Please let me know if things work as expected after I merge https://github.com/filesender/filesender/pull/378/files

monkeyiq:
fixed in 378

PR #373, opened 2018-07-15
allow for mandatory config settings

monkeyiq:
Instead of removing config settings and placing them in magic constants or using some other mechanism a new provision for mandatory settings is being introduced. From there some settings will be documented as private and there is an explicit location that a system admin can look to see what the chosen value for these mandatory settings is for the installation.

### DIFF
--- a/classes/utils/Config.class.php
+++ b/classes/utils/Config.class.php
@@ -155,7 +155,22 @@ class Config {
             
             self::merge(self::$parameters, $config);
         }
+
+        // ensure mandatory config settings file exists
+        $mandatory_config_file = FILESENDER_BASE.'/includes/ConfigMandatorySettings.php';
+        if (!file_exists($mandatory_config_file)) {
+            throw new ConfigBadParameterException('Mandatory config settings file is missing. '
+                                                + 'Please recheck your filesender php code is all installed. '
+                                                + 'Looking for ' . $mandatory_config_file );
+        }
         
+        // load mandatory config settings
+        $config = array();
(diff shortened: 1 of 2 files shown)


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/334.

jornane:
I think it's a really good idea to have a central place for constants, but I'm not sure if the configuration is the best place to do that. It might cause an admin to be confused because he sets a configuration value but doesn't see it applied. It could cause some questions on the mailing list or invalid bug reports.

Could it be an idea to have a Constants class, which serves the same purpose as mandatory configuration settings? Since the value is not a configuration option anymore, it doesn't confuse the admin. As an added bonus, this would also make it immediately clear in the code whether a "magic" value can be changed by the administrator or not.

monkeyiq:
I was thinking that maybe comparing the default to the current setting when loading mandatory settings and throwing an error might be a useful addition. No silently ignoring a setting to make sure everybody knows that the change is in effect.

It is easier to walk the mandatory settings $config and check this than it would be to check each constant in a class. And at times I imagine we would retire a setting from being user set and make it mandatory which would want such code to tell the sysadmin that this has happened.

This is especially true if the settings a site chose to run with are incompatible with the new mandatory setting in the config class.

Personally I would just prefer to document things that a sysadmin might like to avoid setting, such as most of the heavy crypto settings (what algorithm to use etc) and have sane defaults. This would allow a site to setup how it wishes. Assuming that most people are not interested in tweaking crypto details then they would just be the defaults set by the system.

Issue #338, opened 2018-06-15
importKey not safe to "derive" key


fkooman:
https://github.com/filesender/filesender/blob/545b987a2a8f1f2bcf6d2dbd0cab2f49ebde3292/www/js/crypter/crypto_app.js#L27

One should use https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveKey in order to derive a key from a password.

Using "raw" is NOT safe with a string, not even with a (base64|acsii85) generated string! You have to make sure (depending on how many bits the encryption algorithm uses) you have the exact number of bits of entropy you feed to "raw".

TL;DR: use `SubtleCrypto.deriveKey()`.

fkooman:
I didn't actually look at the implementation and example there on MDN on what you need to do exactly. The WebCrypto API is horrible as far as I can tell. What I had in mind: https://download.libsodium.org/doc/password_hashing/the_argon2i_function.html (Example 1: key derivation).

Maybe you can find something similar in JS...

fkooman:
> one minor aside is that changing how the password is imported will need some sort of versioning so that files encrypted with the old direct importKey() will still decrypt.

You could keep this information also in the database together with the file information. Or use a special "marker" at the start of the password, e.g. a character that is not used by your password generation function. That is, assuming the decryption function doesn't know if it is able to successfully decrypt. If it does, you can just try decrypting first with the new method, if that fails fall back to old method (for some time).

fkooman:
The only way out seems to be to use libsodium with argon2i(d), aka the `crypto_pwhash` function. The C API: https://download.libsodium.org/doc/password_hashing/the_argon2i_function.html#key-derivation

The JS implementation: https://github.com/jedisct1/libsodium.js

Note: 
> Keep in mind that in order to produce the same key from the same password, the same algorithm, the same salt, and the same values for opslimit and memlimit have to be used. Therefore, these parameters have to be stored for each user.

In the Filesender case for each file, not for each user.

fkooman:
libsodium has the other benefit of having utilities for encoding/decoding of base64/hex etc in constant time. Also random data generator is available.

fkooman:
https://github.com/filesender/filesender/issues/377

PR #375, opened 2018-07-16
encryption: new Key derivation using importKey then deiveKey

monkeyiq:
This is an update for how an encryption key is created from the given
password when encrypting or decrypting files. As the manner in which
the key is created from the password can create different keys from
the same password this PR also includes versioning for encrypted
files. This way, files encrypted with the old key derivation will
continue to be decrypted while all new files will switch to the new
key derivation scheme.

This PR moves the code closer in style to the example of the
SubtleCrypto API that is published by mozilla.org at
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveKey#Example

Key type 1 forces the encryption to AES-CBC at 256 bit and the number
of hashing rounds is chosen to be secure through around 2020. This is
a trade off where I think the magic numbers will have to be brought
out because there is simply no "one size fits all" for the hash
iteration count. If a site is dealing with sensitive material it may
need 10 times the iterations to meet the site policy. I have left this
setting inline for now.

This is to address the original report of the issue in
https://github.com/filesender/filesender/issues/338

Note that the salt is generated server side and sent out to the client
for both initial encryption and made available for decryption. This
does not allow the server any real help in decrypting the data but
does make each transfer of the exact same data with the exact same
password give different results when encrypted. As such it should help
to stifle password hash tables and the like as each individual
transfer will need to be attacked.

The provision of salt moving server to client in the new
encryption_details object paves the way for more information to be
sent over such as file IDs or other data for the transaction that
might help bind the password to the transfer and improve security.

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -708,7 +708,7 @@ class Transfer extends DBObject {
             if( strlen($this->salt)) {
                 return $this->salt;
             }
-            $this->salt = substr(Utilities::generateRandomHexString(),0,32);
+            $this->salt = Crypto::generateSaltString(32);
             $this->save();
             return $this->salt;
         }
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -100,7 +100,19 @@ class Transfer extends DBObject {
         'options' => array(
             'type' => 'text',
             'transform' => 'json'
-        )
+        ),
+        'key_version' => array(
+            'type'    => 'uint',
+            'size'    => 'small',
+            'null'    => false,
+            'default' => 0
+        ),
+        'salt' => array(
+            'type'    => 'string',
+            'size'    => '32',
+            'null'    => true,
+        ),
+        
     );
 
     protected static $secondaryIndexMap = array(
@@ -138,6 +150,8 @@ class Transfer extends DBObject {
     protected $expires = 0;
     protected $expiry_extensions = 0;
     protected $options = array();
+    protected $key_version = 0;
+    protected $salt = '';
     
     /**
      * Related objects cache
@@ -600,7 +614,7 @@ class Transfer extends DBObject {
         if(in_array($property, array(
             'id', 'status', 'user_id', 'user_email', 'guest_id',
             'subject', 'message', 'created', 'made_available',
-            'expires', 'expiry_extensions', 'options', 'lang'
+            'expires', 'expiry_extensions', 'options', 'lang', 'key_version'
         ))) return $this->$property;
         
         if($property == 'user' || $property == 'owner') {
@@ -690,6 +704,14 @@ class Transfer extends DBObject {
             $recipients = array_values($this->recipients);
             return $recipients[0]->download_link;
         }
+        if($property == 'salt') {
+            if( strlen($this->salt)) {
+                return $this->salt;
+            }
+            $this->salt = substr(Utilities::generateRandomHexString(),0,32);
+            $this->save();
+            return $this->salt;
+        }
         throw new PropertyAccessException($this, $property);
     }
     
@@ -743,7 +765,10 @@ class Transfer extends DBObject {
             
         }else if($property == 'options') {
             $this->options = self::validateOptions($value);
-            
+
+        }else if($property == 'key_version') {
+            $this->key_version = $value;
+
         }else throw new PropertyAccessException($this, $property);
     }
 
--- a/classes/rest/endpoints/RestEndpointTransfer.class.php
+++ b/classes/rest/endpoints/RestEndpointTransfer.class.php
@@ -57,6 +57,7 @@ class RestEndpointTransfer extends RestEndpoint {
             'expires' => RestUtilities::formatDate($transfer->expires),
             'expiry_date_extension' => $transfer->expiry_date_extension,
             'options' => $transfer->options,
+            'salt' => $transfer->salt,
             
             'files' => array_map(function($file) use($files_cids) {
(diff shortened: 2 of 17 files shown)


classes/data/Transfer.class.php
```diff
@@ -690,6 +704,14 @@ public function __get($property) {
             $recipients = array_values($this->recipients);
             return $recipients[0]->download_link;
         }
+        if($property == 'salt') {
+            if( strlen($this->salt)) {
+                return $this->salt;
+            }
+            $this->salt = substr(Utilities::generateRandomHexString(),0,32);
```
fkooman: In effect you are *halving* your salt length here. 16^32 = 2^128 so you actually have only a salt of 128 bits. Which is still adequate, but this is a bad idea because it is very confusing! 

How to properly generate a random hex string in PHP:

```ParagonIE\ConstantTime\Hex::encode(random_bytes(32));```

In JS you would reverse this by 'hex decoding', **in constant time** this "salt" and use it in the key derivation algorithm.

Also please note that it would be much better to use Argon2 for deriving a key from a password in this day and age.

jornane:
Good work, but could you hold off merging this until tomorrow? I'll have a call with SURFnet about this before 17th of July, 16:00 CEST.

monkeyiq:
@fkooman I have moved the salt generation into a new server side Crypto class. The new implementation  uses base64 to string encode the entropy. This should give 24 octets of entropy in the 32 octet database field. Yes, this is wasteful not to store binary data in the database but it does simplify things a little and can be revised at a later time if the wasted disk bytes are seen as an issue.

fkooman:
Assuming for a minute you ignore @jornane's excellent suggestion (in #376) to move generating the salt to the browser, there are still issues with this code of varying severity:

1. You don't use constant time Base64 encode
2. You don't generate a salt string, you generate an encoded salt string
3. The salt should be 128 or 256 bits entropy **exactly**, not "75%"
4. The output length of the encoded salt is really **not** important, it is all about the entropy and you should use the salt as 'binary' anyway, so whether it is hex encoded or base64 encoded doesn't make any difference
5. In JS you have to (constant time) *decode* this salt before using it, not use it _as is_
6. Your method has (and needs!) documentation because of your unconventional approach of generating salts, it should not need documentation for anyone with a passing familiarity with cryptography

Again: please consider switching to [libsodium.js](https://github.com/jedisct1/libsodium.js/) in the browser and get rid of all this stuff, you'll do yourself and anyone reviewing the code a big favour.

monkeyiq:
If the salt is generated in the browser it opens up the need to verify that it is actually "random". I am not sure of the advantage to generating salts on the client unless you assume the server's random number generator is weak. The salt must go server side anyway because it is needed for anyone who intends to decrypt files.

It seems by your emboldening of the constant time issue in previous comments that you see using a non constant time base64 encoding is the weakest part of this PR. If so I am truely happy. I think it would take a long time to reduce the source of a reasonable rng based on the difference in string encoding over network, server, and other random latencies. This code can also be updated in a future PR without any other effect on the system, meanwhile people enjoy the key derivation using the recommendations on mozilla.org. Surely the later is by far (orders of magnitude) more important than the encoding timing.

I will be happy to discuss moving to the constant time salt functions in a future PR.

I don't see that it matters if I generate an encoded salt string or not. The entropy will be less because the string is expanded in the encoding but if you are 24 octets of real entropy expanded to 32 octets of encoded string you still have and only have 24 octets of entropy. This is above the 2^128 minimal amount and when the encoded string with 2^256 bits of length is used we have enough real entropy and are the right length.

This code opens the possibility of moving to other key generation in the future while retaining the ability to decrypt existing files.

jornane:
> If the salt is generated in the browser it opens up the need to verify that it is actually "random". I am not sure of the advantage to generating salts on the client unless you assume the server's random number generator is weak. The salt must go server side anyway because it is needed for anyone who intends to decrypt files.

When using client-side cryptography, you do this because you do not trust the server entirely. If you would trust the server entirely, why encrypt in the first place? The server (or server admin) could willfully generate weak salts and the client wouldn't know. The verification problem that you mention is therefore really the other way around. Since the client does the cryptography, the client should pick the salt. If the client's crypto library is good enough to do key derivation and encryption, it's good enough to generate a salt.

Yes, the salt has to be sent to the server as well, but the salt is not secret. The important thing here is that you should make the salt yourself, because it must be sufficiently random. If the server generates it, you cannot prove that it it sufficiently random, because if you don't trust the admin, and therefore you don't know that the admin hasn't changed the server-side random generator.

> It seems by your emboldening of the constant time issue in previous comments that you see using a non constant time base64 encoding is the weakest part of this PR.

Constant time Base64 is important in a cryptographic context because you want to keep the values being encoded or decoded a secret. Without constant-time, timing attacks can be attempted to read the data being encoded/decoded in Base64. There is not really a reason *not* to use constant time; it's a low hanging fruit. For example, use [sodium_bin2base64](http://php.net/manual/en/function.sodium-bin2base64.php) and you've already solved the problem for PHP >= 7.2.0.

However, @fkooman makes good points beside constant-time as well. Take a look at [libsodium.js](https://github.com/jedisct1/libsodium.js/), #377, and maybe you would be willing to reconsider looking at #376?

monkeyiq:
I will be looking at some of the PRs that were still outstanding. It seems 376 is closed now though.

monkeyiq:
I am happy to merge an update to generate the salt client side and use constant time en/decoders on the client side for the salt. This patch allows key versioning and salt transfer both of which will enable an update much easier than before.

PR #376, opened 2018-07-16
fix random hex string generation

fkooman:
- use `paragonie/constant_time_encoding` to create a hex string from
  random bytes in constant time
- no longer generate random length hex strings (?)
- always generate 256 bits random string


fkooman:
**NOTE** this only fixes the PHP side of things, you'll have to update the JS as well to use this properly!

fkooman:
Also, maybe it wouldn't be such a bad idea to start using [Composer](https://getcomposer.org/)!

fkooman:
See also my comment here: https://github.com/filesender/filesender/pull/375/files#r202602348

fkooman:
> I would not bother in this case since it's only salt generation we're talking about

This attitude *is* a big problem. Just because maybe it would be safe _in this case_, you ignore the possible future use of this method for other purposes.

Dealing with cryptography you MUST use constant time functions, ALWAYS, exactly to avoid shooting yourself in the foot later.

That said, the costs of (external) dependencies should not be underestimated, however, this dependency is of high quality, and minimal in size and itself has no additional dependencies. A no-brainer really.

jornane:
> Dealing with cryptography you MUST use constant time functions, ALWAYS, exactly to avoid shooting yourself in the foot later.

I do not disagree with this, but there are some other considerations here. Looking at the bigger picture, I believe we have no reason to include a server-side crypto library at this point.

My considerations:

1. #375 introduces a call to an otherwise dead function (Utilities::generateRandomHexString, https://github.com/filesender/filesender/pull/375/files#diff-0184e42c84eb01a3bcd332cbd799ef48R711), this function has issues already by itself and I have tried to remove this function in an earlier (but still unmerged) pull request #190.
2. The flow of this code is to generate a salt server side, which is then passed to the client-side JavaScript to do the key derivation. Why not simply generate all the crypto material on the client-side? What is the advantage of generating the salt server-side, even if it is in constant-time?
3. Iff we decide that generating the salt server-side is still the right thing to do, then the salt will already be generated in constant time from PHP 7. The function generateRandomHexString should then still be renamed to better reflect what it does (generateHexSalt).

My proposed solution:

We remove generateRandomHexString as per #190, and then #375 should be updated to generate the salt client-side.

fkooman:
> My proposed solution:
> We remove generateRandomHexString as per #190, and then #375 should be updated to generate 
> the salt client-side.

Sounds good! Even better actually :)

fkooman:
https://github.com/filesender/filesender/issues/377

PR #378, opened 2018-07-17
update of pull 370

monkeyiq:
It appears the main change is the line in the collection class so I thought I'd bring that out and merge it.


glbrimhall:
Yep, that's it. thx

-geoff

PR #382, opened 2018-08-01
db: Move from scattered user_id strings to userid numeric ids

monkeyiq:
In filesender 2.1 and previous versions, the database schema scattered
user_id strings into multiple tables. The user_id strings originated
from the saml auth system as the user identifier. That information
was (potentially) a long string and by it's nature could obviously
contain information that identified a user on the system. The
UserPreferences table placed the user_id into the 'id' column which
was the primary key of the table. RI was not explicit in the schema
for these user_id and userpreferences.id columns.

The new schema moves to using big numbers (8 bytes for example) to
identify the user and stores the saml user_id into a new
authentications table. Because we still want to link a user to the
auth system the new userpreferences table has an authid which
references the auth.id auto increment primary key column. This design
allows more flexibility over how long the link to SAML user id is
maintained while allowing the transfers and other data to remain
valid. Using numeric keys and references also greatly helps speed and
can be efficiently undone with joins using views.

Transfers, Guests, and ClientLogs are also updated to now link to the
userpreferences.id column as the "userid" instead of directly storing
information taken from the saml session.

Part of this update moves to add RI to the system. ClientLogs.userid
is captured as explicitly referencing userpreferences.id and likewise
with Transfers and Guests.

I have tested the migration to the new database schmea with small
sized mariadb and postgresql databases and also migrated the
filesender-2.0beta1 dataset in postgresql and checked that things look
valid after migration in these three cases.

The userauthview view is useful to join with when you want to get the
user_id (saml user id string) field back from the userid numeric data.

For example, finding transfers and inspecting subsets with the following
```
select f.id as fid,transfer_id,f.name,t.userid,a.user_id from files f, transfers t, userauthview a
where t.id = transfer_id and a.id = t.userid
and ((transfer_id > 300 and transfer_id < 350) or (transfer_id > 20000 and transfer_id < 20100));
```
And looking at guests and the user_id saml auth string that invited
the guest with
```
select g.id,userid,g.email,g.subject,a.user_id from guests g, userauthview a where a.id = g.userid
and email like 'guest19%@localhost.localdomain';
```
In the old schema these are, respectively:
```
select f.id as fid,transfer_id,f.name,t.user_id from files f, transfers t
where t.id = transfer_id
and ((transfer_id > 300 and transfer_id < 350) or (transfer_id > 20000 and transfer_id < 20100));

select id,user_id,email,subject from guests
where email like 'guest19%@localhost.localdomain';
```

### DIFF
--- a/classes/autoload.php
+++ b/classes/autoload.php
@@ -51,6 +51,7 @@ class Autoloader {
         'GUIPages' => 'constants/',
         'LogLevels' => 'constants/',
         'ReportFormats' => 'constants/',
+        'DatabaseSchemaVersions' => 'constants/',
         
         'Storage' => 'storage/',
         'Storage*' => 'storage/',
@@ -70,6 +71,8 @@ class Autoloader {
         'User' => 'data/',
         'TrackingEvent' => 'data/',
         'TranslatableEmail' => 'data/',
+        'Metadata' => 'data/',
+        'Authentication' => 'data/',
         '*Log' => 'data/',
 
         'DBLayer' => 'utils/',
--- a/classes/constants/DatabaseSchemaVersions.class.php
+++ b/classes/constants/DatabaseSchemaVersions.class.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * FileSender www.filesender.org
+ * 
+ * Copyright (c) 2009-2014, AARNet, Belnet, HEAnet, SURFnet, UNINETT
+ * All rights reserved.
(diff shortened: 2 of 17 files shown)


monkeyiq:
It is a shame that the db migration ran out of space to happen in travis, but longer term I might have not wanted to keep doing it there anyway because of the slow down per PR. So now for some more testing for a final run to check if things didn't regress in some other cases.

PR #384, opened 2018-08-13
move to just sha1 for that

monkeyiq:
This part was slightly ahead of it's time and the password_hash was the wrong choice for that hashing.

### DIFF
--- a/classes/data/Authentication.class.php
+++ b/classes/data/Authentication.class.php
@@ -161,12 +161,7 @@ class Authentication extends DBObject {
     }
 
     private function updateHash() {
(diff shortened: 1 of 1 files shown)


jornane:
Assuming that [sha1 will be considered insecure](https://arstechnica.com/information-technology/2017/02/at-deaths-door-for-years-widely-used-sha1-function-is-now-dead/) in the future, how will FileSender handle an eventual migration to a new hashing algorithm? Will all user mappings be lost in such an event, or will FileSender attempt multiple hashing methods to match a user?

Issue #388, opened 2018-08-17
Database creation issue


BWar81:
Hello,

I have been following the installation document (https://docs.filesender.org/v2.0/install/) I reach step 9 and the script fails with the following:

`PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function ensureAuthSetup(), 2 passed in /opt/filesender/filesender/scripts/upgrade/database.php on line 128 and exactly 3 expected in /opt/filesender/filesender/scripts/upgrade/database.php:111
Stack trace:
#0 /opt/filesender/filesender/scripts/upgrade/database.php(128): ensureAuthSetup('filesender-upgr...', 'local upgrade s...')
#1 /opt/filesender/filesender/scripts/upgrade/database.php(170): ensureAuthenticationsTableHasReservedIDs(true)
#2 /opt/filesender/filesender/scripts/upgrade/database.php(488): ensureAllTables()
#3 {main}
  thrown in /opt/filesender/filesender/scripts/upgrade/database.php on line 111
`

I am using MariaDB.

Any help would be greatly appreciated!

BWar81:
Thanks for responding to this so quickly :) Apologies its taken me so long to test, I have been away from work. However I have now tried using the new script and I get the following error:

current db_database is filesender
PHP Fatal error:  Uncaught DBIConnexionMissingParameterException: dbi_missing_parameter in /opt/filesender/filesender/classes/utils/DBI.class.php:90
Stack trace:
#0 /opt/filesender/filesender/classes/utils/DBI.class.php(125): DBI::load()
#1 /opt/filesender/filesender/classes/utils/DBI.class.php(202): DBI::connect()
#2 /opt/filesender/filesender/classes/utils/DatabaseMysql.class.php(45): DBI::__callStatic('prepare', Array)
#3 /opt/filesender/filesender/classes/utils/Database.class.php(56): DatabaseMysql::tableExists('Metadatas')
#4 /opt/filesender/filesender/classes/data/Metadata.class.php(99): Database::tableExists('Metadatas')
#5 /opt/filesender/filesender/scripts/upgrade/database.php(70): Metadata::getLatestUsedSchemaVersion()
#6 {main}
  thrown in /opt/filesender/filesender/classes/utils/DBI.class.php on line 90

BWar81:
After reading through #394 I upgraded to MariaDB to 10.2.17 and the script runs through fine without error. Sorry I should have read through other topics first!

monkeyiq:
[...]

PR #389, opened 2018-08-18
db: pass addUser to fix empty db case

monkeyiq:


monkeyiq:
This should fix https://github.com/filesender/filesender/issues/388

Issue #394, opened 2018-08-20
Another DB error (Filesender 2.2)


fransward:
Just installed the new 2.2 version with a fresh clean install (also with a clean DB) and got an error when running the DB initialise command: ([Step 9 - Initialise the FileSender database](https://github.com/filesender/filesender/blob/master/docs/v2.0/install/index.md)).

` php /opt/filesender/filesender/scripts/upgrade/database.php`

My environment:

```
$ mysql --version
mysql  Ver 15.1 Distrib 5.5.56-MariaDB, for Linux (x86_64) using readline 5.1

$ php --version
PHP 5.4.16 (cli) (built: Apr 12 2018 19:02:01)
Copyright (c) 1997-2013 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2013 Zend Technologies

$ cat /etc/redhat-release
CentOS Linux release 7.5.1804 (Core)
```

The results of the DB initialise script:

```
$ php /usr/share/filesender/filesender/scripts/upgrade/database.php
current db_database is filesender
You currently have database major version 0
This execution will move also move to major version 22
Note that a major migration will only be needed when more advanced SQL is used to migrate

Look for table Metadatas
Table is missing, create it
running database update script on existing schema version 0
Removing views from database so base tables are free to change
database views will be added back later in this script
Checking class User
Removing views for table UserPreferences
Array
(
    [userpreferencesview] => Array
        (
            [mysql] => select * , DATEDIFF(now(),created) as created_days_ago   , transfer_preferences LIKE '%encryption":true%' as prefers_enceyption  , DATEDIFF(now(),last_activity) as last_activity_days_ago  , DATEDIFF(now(),aup_last_ticked_date) as aup_last_ticked_days_ago  , id as email_address  , id is not null as is_active   from UserPreferences
            [pgsql] => select * , extract(day from now() - created ) as created_days_ago   , transfer_preferences LIKE '%encryption":true%' as prefers_enceyption  , extract(day from now() - last_activity ) as last_activity_days_ago  , extract(day from now() - aup_last_ticked_date ) as aup_last_ticked_days_ago  , id as email_address  , id is not null as is_active   from UserPreferences
        )

    [userauthview] => Array
        (
            [mysql] => select up.id as id,authid,a.saml_user_identification_uid as user_id,up.last_activity,up.aup_ticked,up.created from UserPreferences up, Authentications a where up.authid = a.id
[...]

PR #395, opened 2018-08-20
docs install db mariadb: update to instructions for views

monkeyiq:
Some updates for the new view stuff and permissions.


monkeyiq:
This should resolve https://github.com/filesender/filesender/issues/394

PR #407, opened 2018-08-31
ci: do the tests against mariadb as well as postgresql

monkeyiq:
We should setup a matrix to run the tests against postgresql and mariadb (version 10.2 or whatever we select) to better protect against issues that happen only on one of these database servers.


monkeyiq:
As this has to run on the CI to 'test' it, things will change as it gradually, hopefully moves toward a working CI udpate. ie, I have to run things on the CI to test that the changes actually work.

PR #410, opened 2018-09-02
ci: lets split the 2 jobs into 5

monkeyiq:
Now that we work against postgresql and mariadb (https://github.com/filesender/filesender/pull/407) it is probably a good idea to split from 2 jobs, one for postgresql and one for mariadb, into 5 jobs. The simple end game here is to make each job run for a shorter amount of time so that CI completes quicker. As can be seen from the last 2 job build it takes 15.5 minutes to run a CI batch (https://travis-ci.org/filesender/filesender/builds/422983697). 

I think Selenium is the slowest single path so that might be able to be split into 2 jobs as well in order to bring overall CI runtime down at the expense of 7 parallel builds. But that will be a job for a subsequent PR.

For this PR, the "core" php only tests are split out to make the 2 existing jobs (postgresql and mariadb) into 4 jobs (browser tests and core for each database) and then the large dataset job is split out of the postgresql core job into it's own job. This gives a bit of wasted cpu cycles as five VMs are made and setup in parallel, but the real world time from making a PR to a yes/no should be shorter. And as core jobs run quicker than selenium tests we should be able to add more core tests without extending the overall time for CI.


monkeyiq:
Hmm, only down to 14.5 minutes. Handy to have availability to extend core without impacting the time frame but splitting the selenium into 2 parts is going to be a great addition.

monkeyiq:
it comes to around 13 minutes with the split in d44 above. It might get down to 12 if test shuffling is spot on but there is also loss where VMs are not all started right off the bat.

Issue #408, opened 2018-08-31
Errors DB upgrade on MYSQL


Veredior:
Just tried to upgrade our filesender 2.0 server, but it seems there are some incompatibilities with a MySQL 5.5 database due to a Debian 8 install.

`[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021] "SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'integer) as target_id_as_number  from AuditLogs' at line 1"
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021] Array
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021] (
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021]     [name] => exec
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021]     [args] => Array
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021]         (
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021]             [0] => CREATE OR REPLACE VIEW auditlogsview as select * , DATEDIFF(now(),created) as created_days_ago , cast( target_id as integer) as target_id_as_number  from AuditLogs
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021]         )
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021]
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021] )
[2018-08-31 13:49:33] [upgrade:error] [DBIUsageException details uid:5b892b4d92021]`

After some research, you cant use the CAST function like this `cast( target_id as integer)` instead you can do this : `cast( target_id as unsigned)`

I will create a merge request for this issue.

Second major issue is with the filesbywhoview view which contains a subquery, and this is a big no for all versions before MySQL 5.7.7 . One solution is to create a view for this subquery and call it in filesbywhoview view.

madsi1m:
do we split this into two views? any performance hit?

madsi1m:
oops, just noticed that 10.2 is minimum version for v2.3

[...]

PR #414, opened 2018-09-04
Db upgrade fixes

Veredior:
Should fix the db upgrade on mysql, mariadb and pgsql.

### DIFF
--- a/scripts/upgrade/database.php
+++ b/scripts/upgrade/database.php
@@ -74,7 +74,8 @@ DBI::beginTransaction();
 // Get data classes
 function getClasses() {
     $classes = array();
-    $classes[] = 'User'; // This must be before Authentication so users can be setup too
+	// Those classes must be before everything else and in this order to create filesbywhoview
+	array_push($classes, 'User', 'Authentication', 'Guest', 'Transfer');
     foreach(scandir(FILESENDER_BASE.'/classes/data') as $i) {
         if(substr($i, -10) != '.class.php') continue;
         $class = substr($i, 0, -10);
--- a/scripts/upgrade/database.php
+++ b/scripts/upgrade/database.php
@@ -226,40 +226,49 @@ function updateIntoTable( $updateTable, $fromTable, $setClause, $whereClause )
 }
 
 //
-// Create forign keys if they are not there already
+// Create foreign keys if they are not there already
 //
 function ensureFK()
 {
     $dbtype = Config::get('db_type');
-    $tbl_auth      = call_user_func('Authentication::getDBTable');
-    $tbl_transfers = call_user_func('Transfer::getDBTable');
-    $tbl_guests    = call_user_func('Guest::getDBTable');
-    $tbl_clientlog = call_user_func('ClientLog::getDBTable');
-    $tbl_user      = call_user_func('User::getDBTable');
+	$tbl_auth       = call_user_func('Authentication::getDBTable');
+	$tbl_user       = array(
+		'tbName' => call_user_func('User::getDBTable'),
+		'indexName' => 'UserPreferences_authid',
+		'reference' => $tbl_auth
+	);
+	$tbl_transfers    = array(
+		'tbName' => call_user_func('Transfer::getDBTable'),
+		'indexName' => 'Transfers_userid',
+		'reference' => $tbl_user
+	);
+	$tbl_guests      = array(
+		'tbName' => call_user_func('Guest::getDBTable'),
+		'indexName' => 'Guests_userid',
+		'reference' => $tbl_user
+	);
+	$tbl_clientlog    = array(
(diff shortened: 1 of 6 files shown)


monkeyiq:
I am taking a look at this PR now. I plan to move the call call_user_func('Transfer::getViewName') in the Files view definition to something slightly different. I think it is simpler for a class to create a duplicate view of it's own instead of relying on class ordering in database.php. I think the later is not so scalable as other cases might crop up in the future. These redundant (from a subselect compatible database point of view) views will also have to be also dropped in the reverse order they are made in for each class in order to cater for this one view depends on the other case.

monkeyiq:
I am also bringing over the comment from https://github.com/filesender/filesender/issues/408#issuecomment-418301554 "#414 pull request should fix the issues on mysql. I cant test for now on mariadb and pgsql.". I am happy to test things knowing what state something is in.

It seems on postgresql the table name is stored in the information schema as lower case.
```
filesender=> SELECT COUNT(1) FROM INFORMATION_SCHEMA.table_constraints WHERE constraint_name='transfers_userid' AND table_name='transfers';
 count 
-------
     1
(1 row)

Time: 2.274 ms
filesender=> SELECT COUNT(1) FROM INFORMATION_SCHEMA.table_constraints WHERE constraint_name='transfers_userid' AND table_name='Transfers';
 count 
-------
     0
(1 row)
```
For postgresql 9.6.9 on Fedora and pgsql 9.3.24 on debian jessie/sid.

This is an interesting additional case, maybe the CI should run the database.php twice to catch these extra cases. It feels like there would be an attempt to recreate the fk on pgsql on a subsequent run.

scripts/upgrade/database.php
```diff
@@ -226,40 +227,49 @@ function updateIntoTable( $updateTable, $fromTable, $setClause, $whereClause )
...
+		// We check the presence of the foreign key
+		if( $dbtype == 'pgsql' ) {
+			$checkQuery = DBI::query("SELECT COUNT(1) FROM INFORMATION_SCHEMA.table_constraints WHERE constraint_name='".$tbl['indexName']."' AND table_name='".$tbl['tbName']."';");
+		}
+		else {
+			$checkQuery = DBI::query("SELECT COUNT(1) indexExists FROM INFORMATION_SCHEMA.STATISTICS WHERE table_schema='".Config::get('db_database')."' AND table_name='".$tbl['tbName']."' AND index_name='".$tbl['indexName']."';");
+		}
+
+		if ( $checkQuery->rowCount() === 0) {
+			array_push( $a,'ALTER TABLE '.$tbl['tbName'].' ADD CONSTRAINT '.$tbl['indexName'].' FOREIGN KEY (authid) REFERENCES '.$tbl['reference'].' (id)  on delete cascade on update restrict ;');
```
monkeyiq: Note that the fk is authid on the first case and userid on the next 3. This SQL statements doesn't run on postgresql fir Transfers because there is no authid column in that table. 

It seems this got through the CI because the foreach iterating over $a and actually executing the SQL to create the FK is removed from the PR.

monkeyiq: '.$tbl['reference'].' in the SQL will give an Array instead of the table name.

monkeyiq:
I will have an update to this PR shortly.

monkeyiq:
I have an update to the FK code in https://github.com/filesender/filesender/pull/423 which should help with the first point above.

Veredior:
I think you are right for userid, but the authid in User class, as Authentification class did not exist before August. 
The authid column in User when created is putting 0 by default, which throws an error .
Adding nullable true prevents this.

monkeyiq:
One of the things that makes this update so difficult is the many ways it can interact, new database vs update to existing and various versions of RDBMs.

For a migration from older versions these new columns are first created with 'null'=>true because you can't create a populate them in one go:
https://github.com/filesender/filesender/blob/master/scripts/upgrade/database.php#L431

The Auths table is then updated to contain every known saml id
https://github.com/filesender/filesender/blob/master/scripts/upgrade/database.php#L458

The Users table should then has the authid set for every user.
https://github.com/filesender/filesender/blob/master/scripts/upgrade/database.php#L471
Because the Auths table got a union of all saml ids then we should be fairly sure the above update is complete, leaving no user with an unknown authid.

So after the above, setting the users.authid to be 'not null' should be a safe operation. This is why the users class was made with authid to be not null. A new entry into the users table should only be done after a new entry to the auths table has completed, so we should be able to set the authid to the auths.id at that time and not need a null.

I would really like to fix the case where the user.authid is left as 0 (unset) for a user rather than allowing a user to be created with a null authid. Do you recall an interaction that can lead to this case so I can investigate and update?

monkeyiq:
I think issues that soften the RI might be best discussed in new PRs / issues. Thank you for the PR, hopefully the updates I have merged based on it are inline with what is needed.

PR #421, opened 2018-09-15
sql: bring back more advanced view, fk creation no longer uses if exists

monkeyiq:
This is an update to some of the code from
https://github.com/filesender/filesender/pull/414


monkeyiq:
While the shell of the code is updated, this doesn't work as yet because the check for existing fk queries are not updated yet.

monkeyiq:
still testing on pgsql and mysql

monkeyiq:
I am not sure if it is worth trying to abstract the fk creation out to that level.

PR #429, opened 2018-09-16
This formats the classes directory

monkeyiq:
This should give a comparison to the formatting performed in https://github.com/filesender/filesender/pull/307/files

One thing I notice is this configuration is dropping array( into [ which might not be what we want.


monkeyiq:
So the short array option had made it's way into .php_cs.dist. The flipside here, as with many automated formatting tools is that it is effective in all cases. So instead of a 
```
function  taking(foo = []) 
you get
function takling(foo = array())
```

PR #430, opened 2018-09-19
Fix escaping quoting

Veredior:
After editing the french translation on poeditor, i tried to use the import-translation-for-language command, but the function squote is deleting the letter before escaping a single quote.

Example :
 - Export Poeditor : `M\'ajouter aux destinataires`
 - After the actual squote function : `\'ajouter aux destinataires` 

The commit fixes this issue.


monkeyiq:
Looking at the logs I suspect the CI failure is unrelated to this change. As I update the branch after merging other stuff I will see if it comes good.

PR #441, opened 2018-10-23
Expose config transfer option 'get_a_link' to JS

### LINKED ISSUE #439: get_a_link pop-up misses download link when hard-coded
When `available => true` is set for the `get_a_link` option in the `transfer_option` array the pop-up/overlay shown after the upload contains the full download link (which is the desired behaviour):

```php
  'get_a_link' => array(
    'available' => true,
    'advanced' => false,
    'default' => true,
  ),
);
```

get-a-link_available_true

But when (only) changing `available => false` for this setting (causing the checkbox to disappear from the UI for uploads, and the parameter to always be active) after the upload completes the download link is missing from the pop-up/overlay:

get-a-link_available_false

(In both cases after the pop-up/overlay is closed one is taken to the transfer for the finished upload, and the download link can be seen there, though somewhat burried in lots of other information.)

The expected behavior is that whether _the same parameter_ is hard-coded in the config or set in the UI the result should be the same, in this case the pop-up/overlay should contain the download link.

peter-:
This PR fixes #439 only for the case of the `get_a_link` config setting.

(Any other config parameters set in config.php that may related to UI behaviour will continue to be ignored, AFAICT.)


www/filesender-config.js.php
```diff
@@ -47,6 +47,12 @@ function value_to_TF( $v )
     return json_encode( ($v ? true : false) );
 }
 
+$to = Config::get('transfer_options');
+$to_getalink = $to['get_a_link'];
+if ($to_getalink['available'] === false && $to_getalink['default'] === true) {
+  $get_a_link = true;
+}
+
```
peter-: I suspect a more general way to deal with this may be in order, to expose all `transfer_options` (and `guest_options`) that have been set in config.php to be `available => false` AND `default => true` (i.e., they're active and _not_ exposed to the UI), since AFAICT they're all currently being unseen/ignored by the JavaScript code.

As it is this serves to validate the approach in one particular case (for the `get_a_link` transfer option) and fixes an existing bug #439. A new issue can be opened for the more general case in case more options from config.php need to be exposed.

peter-:
I can't see how the changes in this PR would have caused those checks to fail. Any suggestions?

monkeyiq:
In the screenshots I see some of the buttons are just hyperlinks. eg the send, clear all, and send files. At times that I have had that it during development it was a javascript issue.

https://app.saucelabs.com/tests/4ad773c8a4564e05ba138bd516a720c0/commands?auth=bdddb310229b40f36ef157c6bd340488#9

Bringing this down locally I do not see that issue. I have done an update to give it a chance to rerun travis just in case it was an intermittent problem. If it remains I will look closer at the failing tests and try to reproduce the issue locally.

peter-:
Any updates?

peter-:
Any updates?

monkeyiq:
I have created https://github.com/filesender/filesender/pull/471 which has a similar effect. The main design difference is that options are sent from upload_page.php to the browser as hidden form elements. This allows the javascript code to not have any special case checks if something is a user set or sysadmin set option. IMHO it would be all too easy for javascript code to miss doing a check if an option exists, is set, or is in the filesender.config.transfer_options. 

I think it also makes sense for the form to always have all elements so that javascript doesn't need to check if get_a_link exists in the options, which makes the check reduce to if( transfer.options.get_a_link ) { ... }. But that is food for another PR.

At the moment 471 is only set to allow get_a_link to pass through to the browser if it is available=false. There may be security issues in making all options that are not available flow client side so I have created a configurable passlist to make that explicit.

monkeyiq:
Hopefully you like 471 and the design of having the defaults sent this way rather than explicitly checked for in the code.

monkeyiq:
Feel free to reopen the PR if the design of 471 doesn't work for you. If you have other variables that are fairly safe to be in the passlist along with get_a_link then I am happy to extend to include those aswell.

PR #444, opened 2018-10-25
Consistently use Utilities::http_build_query

peter-:
when building self-referential URLs with query strings, as used by other [Templates](https://github.com/filesender/filesender/blob/filesender-2.4/templates/download_page.php#L25-L34) or Classes, e.g. [Recipient.class.php](https://github.com/filesender/filesender/blob/filesender-2.4/classes/data/Recipient.class.php#L292) or [TranslatableEmail.class.php](https://github.com/filesender/filesender/blob/filesender-2.4/classes/data/TranslatableEmail.class.php#L382).

Actually I'm interested in replacing calls to `Utilities::http_build_query()` with another function that makes the generated URLs site-configurable beyond the `site_url` (e.g. in order to support "clean URLs"), but let's make the existing code more consistent first.

### DIFF
--- a/classes/auth/AuthSPFake.class.php
+++ b/classes/auth/AuthSPFake.class.php
@@ -166,7 +166,7 @@ class AuthSPFake
             if (!$landing_page) {
                 $landing_page = 'upload';
             }
-            $target = Config::get('site_url').'index.php?s='.$landing_page;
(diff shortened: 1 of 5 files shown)


jornane:
This is a very nice code cleanup, great job! It's also a good start for solving #98 and #137.

PR #454, opened 2018-12-18
language fix

madsi1m:
changed "to to" to "to"


monkeyiq:
I will merge this soon. I have also updated the translations on poeditor.com for these files looking at the guest_upload_complete token that generates them.

PR #470, opened 2019-03-17
trying to get pr 441 to pass CI

monkeyiq:


monkeyiq:
This is now a quest to see why the CI didn't like the code as it is small and might gain useful knowledge down the line

monkeyiq:
so it appears to not like undefined values

PR #471, opened 2019-03-17
create some hidden options to set not available options in the client javascript.

monkeyiq:
This is an alternate way to handle what https://github.com/filesender/filesender/pull/441 is doing.

The idea of this PR is to put the available=false options into hidden form elements. The code can then get those options just as it gets the shown options. The goal is to avoid having javascript code that does form.optionx is set or config.transfer_options_optionx is set because the later is communicated to the client in the hidden field which creates form.optionx. 

Yes, the client can then reverse this or mangle the javascript to show these hidden options but the server should recheck them to make sure nothing funny is going on. This issue is present no matter how these available=false options are sent. For get_a_link they must be sent to the browser as the javascript code has to operate differently when the option is set (available=false && default=true).

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -618,6 +618,21 @@ class Transfer extends DBObject
             return true;
         });
     }
+
+
+    /**
+     * Get options that are not available for user setting
+     *
+     * @return array
+     */
+    public static function forcedOptions()
+    {
+        return array_filter(self::allOptions(), function ($o) {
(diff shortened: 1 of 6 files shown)


monkeyiq:
Hopefully the above makes sense. This PR itself could also be extended on. For example, maybe using a model that is bound to UI elements instead of putting state directly into UI elements in upload_page.php.

monkeyiq:
This is for issue https://github.com/filesender/filesender/issues/439

PR #476, opened 2019-03-21
Update dutch language (fix subject, small changes) 

GPlaisier:


fransward:
The source of the Dutch translation is handled by PoeEditor. See: https://poeditor.com/.

monkeyiq:
I see that a lot of these change the "subject:" line back which is great. I am reluctant to go in an step on the toes of the people already listed on the Dutch translation on poeditor to directly import these changes. Also since I don't make any claim to speak the language at any level. 

@fransward if you think these changes look good I am happy to bring them into poeditor.

monkeyiq:
I think I will close this as poeditor is the place to make the changes and the PR itself is unlikely to be merged this way.

Issue #466, opened 2019-02-14
php 7.2 compatible?


madsi1m:
Is filesender php 7.2 compatible? I noticed php 7.0 is now unsupported

monkeyiq:
I have been running it on php 7.2.13 for a little while. I will be doing a bunch of updates so will post if I see anything not working on 7.2 (and likely fix it if that is the case while I am at it ;)

FWIW I test on the desktop against mariadb-10.3.11 and postgresql-server-10.6

The CI env uses php 7.0 and mariadb 10.0 (and postgresql too, but less specific on version) https://github.com/filesender/filesender/blob/master/.travis.yml#L5. Maybe we should bump the CI to php 7.2 instead of testing on an unsupported language version.

monkeyiq:
so... now to update the docs!

monkeyiq:
As mentioned in 488 above, the CI is now using php 7.2 to help ensure future compatibility.

monkeyiq:
CI is now on php 7.2 and latest simplesamlphp along with using the vendor phpunit to avoid the incompatibility of selenium with some versions of phpunit.

So in short, php 7.2 is now the recommended version :)

PR #486, opened 2019-03-25
lets run the CI against php 7.2 now instead of 7.0

monkeyiq:
As noted in https://github.com/filesender/filesender/issues/466 php 7.0 is unsupported so we should run CI against a version that will be used in the wild.


monkeyiq:
Unfortunately selenium doesn't like this very recent phpunit that was included in php 7.2 by default. I have made and merged https://github.com/filesender/filesender/pull/488 to update to php 7.2 and continue working with selenium for now.

Issue #452, opened 2018-11-27
webcrypto-shim.js: bad algorithm name with FS2.4 and IE11


fransward:
We discovered another problem after we updated to Filesender 2.4. Several users are complaining that they cannot use file encryption when using IE11.

Just tested myself and the upload process fails after a while with seemingly no progress:
"No progress has been reported for your currently uploading transfer for some time, if you think it is tuck and before reporting it to your support team please send the last log entries from your user interface by clicking this button: [send logs]"

The IE debugger window shows:
`throw new SyntaxError(Bad algorithm name");
Exception thrown and not caught`

fransward:
Also when you receive an encrypted file IE11 does not decrypt when the sender used another browser.

monkeyiq:
The IE11 incompatibility was introduced to address a security issue raised at https://github.com/filesender/filesender/issues/338 which recommends using deriveKey().

Unfortunately deriveKey() is not supported in IE11 and the shim code to call the web crypto api does not give a very meaningful error. Although the documentation (https://github.com/filesender/filesender/blob/master/www/vendor/node_modules/webcrypto-shim/README.md) mentions that ie11 and safari lack deriveKey support I found using the test page (https://vibornoff.github.io/webcrypto-examples/index.html) that safari supports PBKDF2 + deriveKey. I have also just performed another test upload on encrypted content using safari on a recent osx machine to confirm. On the other hand, on current Windows 10 using its IE11 version there is no deriveKey in any row on the compatibility page.

I will put an issue on github to allow the crypto version to be set in config.php instead of in the code itself for new files so that admins can choose which key code to use in order to support the browsers they wish to support.

monkeyiq:
I have created https://github.com/filesender/filesender/issues/472 to track the version selection.

monkeyiq:
and merging https://github.com/filesender/filesender/pull/475 allows the key version to be explicitly set in the config.php if an admin wishes to use force using older key versions to support older browsers.

PR #487, opened 2019-03-25
docs: ie11 and keyversion

monkeyiq:
This offers the reader some information about the ie11/keyversion https://github.com/filesender/filesender/issues/452 when installing. It also opens the door for other compatibility issues to be listed at an appropriate time for new installations.


monkeyiq:
love it when a docs only build somehow can break ci. getting that not to happen will be a huge plus.

PR #488, opened 2019-03-26
php72 but use the vendor phpunit

monkeyiq:
This moves the CI to php 7.2.

Because selenium doesn't like the very latest phpunit that is used by php 7.2 by default I have also moved to using the phpunit that is installed from composer in the vendor directory to ensure compatibility.

There was an issue with the very old simplesamlphp being used so I have updated that to the latest and made moves to allow the version to be updated again in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: trusty
 sudo: required
 language: php
 php:
-  - 7.0
+  - 7.2
 services:
   - postgresql
   - mariadb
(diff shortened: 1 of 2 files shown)


monkeyiq:
This is to address the issue where php 7.0 (unsupported) was being used as reported in https://github.com/filesender/filesender/issues/466.

This is a cleaner way to handle things than https://github.com/filesender/filesender/pull/486 which ran into the issue of phpunit being too new for selenuim to handle.

Issue #417, opened 2018-09-11
Subject in mail is Filesender when using dutch language in browser


janwout:
Also in German. When you send a file with filesender 2.1, with a browser with dutch or german language, the subject you enter is ignored and Filesender is the subject of the email sent to the receiver.

monkeyiq:
I have added some information to the i18n page about this
https://docs.filesender.org/v2.0/i18n/

basically, the system uses the "subject: " key in the translation file as a specific thing to change for email subjects. The dutch language file was translating that to "onderwerp: " which was then not being seen by FileSender as something it should translate and use as the email subject line. It is a little confusing to have this special english keyword that must remain english in the translation on poeditor, but that is how the system runs at the moment. It simplifies the code not to try to look for a language translation of "subject:" but just for the english keyword for this purpose. I have updated the poeditor translation for this email and will be importing updated translations from poeditor to github shortly.

monkeyiq:
I see that https://github.com/filesender/filesender/pull/476 changes a bunch of the onderwerp: to subject: too. I have just rechecked the following list which appear to already have been updated. I imagine there is a history somewhere in poeditor though it isn't jumping out at me. I think I will email the filesender-dev list to mention this to other folks.

```
bounce_report
daily_summary.mail.php
download_complete.mail.php
email_feedback.mail.php
file_deleted.mail.php
files_downloaded.mail.php
guest_access_upload_page.mail.php
guest_cancelled.mail.php
guest_created.mail.php
guest_created_receipt.mail.php
guest_expired.mail.php
guest_reminder.mail.php
guest_upload_complete.mail.php
guest_upload_start.mail.php
recipient_deleted.mail.php
recipient_feedback.mail.php
report_attached.mail.php
report_inline.mail.php
storage_usage_warning.mail.php
transfer_autoreminder_receipt.mail.php
transfer_available.mail.php
transfer_deleted.mail.php
transfer_deleted_receipt.mail.php
transfer_expired.mail.php
transfer_expired_receipt.mail.php
transfer_reminder.mail.php
translate_email_footer.mail.php
upload_complete.mail.php
transfer_deleted_receipt
transfer_expired_receipt
guest_upload_complete
guest_upload_complete_confirmation_to_guest
upload_complete
guest_upload_start
guest_created
guest_created_receipt
guest_reminder
guest_cancelled
guest_expired
transfer_available

```

monkeyiq:
[...]

PR #495, opened 2019-03-28
docs: make sure not to translate the subject keyword from english in the email headers

monkeyiq:
this relates to the subject not being set correctly for some locales https://github.com/filesender/filesender/issues/417.


monkeyiq:
yay for the CI

PR #504, opened 2019-03-31
translations dutch

monkeyiq:


fransward:
Thanks Ben for merging the Dutch translations.
Just updated the code with the latest master branch, but now we see an error:

`[Sun Mar 31 11:38:38.510504 2019] [:error] [pid 39204] [client 92.110.82.192:30394] PHP Fatal error:  Default value for parameters with a class type hint can only be NULL in /usr/share/filesender/filesender-2.5/classes/utils/Lang.class.php on line 263`

Line 263 in Lang.class.php is:
`public static function setlocale_fromUserLang( int $category = LC_ALL )`

Btw, we are still on PHP 5.4.16

monkeyiq:
I think this is due to the hints on $category. I have https://github.com/filesender/filesender/pull/505 to help remove that issue. Feel free to test the change of 505 if you would like to get going before I merge it.

fransward:
Just updated to PHP 7.2 and dit a service httpd reload and all is fine now!

` php -v
PHP 7.2.16 (cli) (built: Mar  5 2019 14:45:10) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies`

Meaning that PHP 7.2 is mandatory now.

monkeyiq:
I think the PR I linked to would allow things to run on 5.x still. 

If making sure things work on 5.x is still a good idea is a more open question. php 7.0 and 5.6 both stopped getting security updates 2-3 months ago. Longer for 5.4. Also there are language features we could be using from later php (like the typing that was not compatible with 5.4) which can help to improve the code and project.

Issue #325, opened 2018-06-01
Ability to create a tar archive as well as zip


monkeyiq:
One platforms like osx and Linux the ability to download a tarball might be more convenient than a zip file. For example, on osx a tarball can be extracted from finder by double clicking on it. For osx this would get around the zip64 issue by not needing any additional software install in order to deal with larger archives.

monkeyiq:
Offering tar links by default on osx will also get around this issue https://github.com/filesender/filesender/issues/295

jornane:
An additional advantage of tar is that the filesize is known in advance, which means that you can send a Content-Length header.

monkeyiq:
The above PRs 513 and 514 bring this in.

PR #512, opened 2019-04-02
osx: always show zip64 warning on download page

monkeyiq:
This is a band aid until https://github.com/filesender/filesender/issues/325 is completed.


monkeyiq:
And https://github.com/filesender/filesender/pull/513 will better help osx users!

PR #514, opened 2019-04-05
tar: calculate the content-length and set header

monkeyiq:
The content-length can be calculated using a shallow pass over the data, building an archive without needing to read/write the contents of the files from storage. Using the same API gives the advantages that the padding and other metadata included in the tar is going to be the same as what is generated using a non shallow (real) copy of the data when we stream the real archive to the user.

Note that if you have something like the following running for a tree that includes www/download.php then the size information will be lost in transit. 
```
<Directory "/opt/filesender">
<IfModule mod_deflate.c>
SetOutputFilter DEFLATE
</IfModule>
```

If you are going to on the fly compress like that then maybe use something like the below to override for just download.php. There are likely more elegant ways to disable it.

```
<Files "download.php">
  SetOutputFilter INCLUDES
</Files>
```

### DIFF
--- a/classes/utils/Archiver.class.php
+++ b/classes/utils/Archiver.class.php
@@ -90,6 +90,47 @@ class Archiver
     }
 
 
+    /**
+     * This is a bit sneaky, we create a temporary tarball to include
+     * all the byte offsets and padding but do not actually read/write
+     * the real storage contents, only record the metadata in the archive.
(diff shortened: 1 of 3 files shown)


monkeyiq:
This completes the new tar support
https://github.com/filesender/filesender/issues/325

PR #522, opened 2019-04-09
Update Armenian language file

madsi1m:
Old file had too much google translate which may not be 100% accurate.


monkeyiq:
I still haven't worked out what it is that is making the CI flakey at the moment :((

PR #523, opened 2019-04-09
db statlog: move is-encrypted to being a real column

monkeyiq:
It isn't a really great design to take a boolean value and put it into additional_attributes only to try to pick it out again in the SQL query and group by it. We might as well store it as a bool in the database and avoid the unneeded to string and LIKE loops.


monkeyiq:
This improves on the PR https://github.com/filesender/filesender/pull/521 to be more complete.

PR #528, opened 2019-04-11
Revert "Adding sl_SI translations"

monkeyiq:
Doing an import of Slovenian from poeditor shows a number of differences between the PR and what is on poeditor.
Reverts filesender/filesender#526


monkeyiq:
I am not sure which translator on poeditor the username creating the PR matches up with either.

1katoda:
My POEditor username is derived from my student email, so its Tj3143. As far as the differences are concerned - if you are reffering to the differences between the PR and POEditor, the only differences are:
- no html commas (couldn't get rid of them on POE)
- 7 backslashes escaping single quote, as per en_AU translation in php (en_AU translation in POE shows 3)
The translations and PR were added by me and reviewed by my coworker, so they should be fine. There might have been a few (grammatical) corrections to the original translations.

1katoda:
Never mind the html commas part, I forgot I already solved that POEditor nuissance yesterday by exporting, fixing with sed and importing again. They were there in the first place because I had some strings already translated locally and I imported them via a csv file, in which commas are the separators.

monkeyiq:
Cool, I'll leave this revert closed instead of using it. 

Nice to meet you! I felt a little odd merging when I couldn't see the direct link to one of the poeditor users for the language and had no history to go on.

1katoda:
Nice to meet you too!
It's understandable, you don't want a shady commit of a lot of php files from an unknown source :)
I'm not that familiar with POEditor, so I'm not sure how you see the pages and data, but there is no slovenian translation option found, so I "suggested a new language", which might have added to shadiness.

PR #538, opened 2019-05-06
security: start removing the possible use of frames.

monkeyiq:
There is only one area that I can see that uses frames/iframe in filesender. This is to support legacy browsers that do not support the FileReader API. The FileReader is supported on IE10 and many older browsers (https://developer.mozilla.org/en-US/docs/Web/API/FileReader#Browser_compatibility).  This makes the legacy iframe code very unlikely to be useful these days. 

As such I have made the default policy for X-Frame-Options to be deny. The iframe code has not been pruned yet in case it is discovered to be required by some sites and the X-Frame-Options can be adjusted to allow it again. If nobody notices that x-frame is denied then the legacy code can be pruned away.

### DIFF
--- a/classes/rest/RestServer.class.php
+++ b/classes/rest/RestServer.class.php
@@ -304,7 +304,7 @@ class RestServer
             //
             // Security that applies to all REST requests
             //
-            header('X-Frame-Options: deny');
+            header('X-Frame-Options: sameorigin', false);
 
             //
             // Output data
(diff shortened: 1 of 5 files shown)


madsi1m:
Will this break filesender if i run filesender it self in an iframe?

monkeyiq:
Ah, good point. Will sameorigin work for you there?

madsi1m:
Should do, can test once it is merged

monkeyiq:
If it doesn't work I'll revert later in the day.

Issue #537, opened 2019-05-02
404 for translate_email_page.js


fransward:
There is a reference for a **translate_email_page.php** in **templates/translate_email_page.php**,
but this file doesn't exists and therefore produces a 404.

Offending line:
`<script type="text/javascript" src="{path:js/translate_email_page.js}"></script>`

fransward:
btw, we're using the latest codebase 2.6

monkeyiq:
It seems many of the templates look for their name in path:js/name.js. I don't see a file js/translate_email_page.js here either. I am tempted to add a blank file rather than remove the reference.

PR #542, opened 2019-05-07
add stub translate_email_page.js file to avoid 404

monkeyiq:
This follows the same pattern as other template.php files and allows easy future expansion.


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/537

PR #549, opened 2019-05-11
external libs: a better plan for keeping up to date

monkeyiq:
I will perform some testing on this tomorrow. It is almost certain that I have some dandling path reference to the old www/lib location for at least some files. The PR is up here mainly as an indication of intent, until more testing is done.

I have moved the graph and terasender subdirectories from www/lib as they are internal libraries. They have been moved to www/js. This leaves only external libraries in www/lib such as jQuery et al. An attempt was made with grep to update all references, but testing is needed to verify.

The plan is to stage files in scripts/dev/www-lib-update and use scripts in that directory to show the npm installed version and the versions (from bundle files) that are in use in www/lib. Another script will be able to update the later (in use) from the former (node_modules). This should make it more automatic to keep up to date.

Instead of directly using the node_modules directory a minimal install will be copied to www/lib from the node_modules directory when an update is desired. This will avoid having testing files and other files that are undesirable to serve from www/lib.


monkeyiq:
This moves to the following versions and information is comtained in README.txt in scripts/dev/www-lib-update as to how to see the existing versions and how to update to newer versions as desired.

```
       chart.js      2.8.0  chart.js/Chart.bundle.min.js
   font awesome      4.7.0  font-awesome/css/font-awesome.css
         jQuery      3.4.1  jquery/jquery.min.js
      jQuery UI     1.12.1  jquery-ui/jquery-ui.min.js
          reset        2.0  reset/reset.css
```

monkeyiq:
This is now getting into shape. I want to also move www/vendor into this new format but will likely do that in a subsequent PR.

PR #580, opened 2019-06-19
Update for language de_DE

baumrasen:


monkeyiq:
I am bringing in all the translations from poeditor now in https://github.com/filesender/filesender/pull/618

PR #587, opened 2019-07-01
fixed the browser/os stats display in admin_statistics_section.php (FS > 2.5)

belnet-ming:
Fixed the  browser/os/encryption stats display, somehow...
OS detection doesn't seem to work?

-------------------------------------------------------
Older stats (pre 2.6) are displayed like before with commit f24ab04 
```
Unknown
Unknown
{"encryption":true}
```

-------------------------------------------------------

Older stats (pre 2.6) are displayed like newer stats (>= 2.6) with commit 26b2b60 .
image


belnet-ming:
Commit 38e0bd0 fixes the OS parsing from the HTTP_USER_AGENT.

In DBConstantOperatingSystem.class.php:
- all regexes changed to be case insensitive (ie. Linux not matched because of regex=linux, user agent= Linux)
- regexes for Windows version fixed

In DBConstantBrowserType.class.php:
- all regexes changed to be case insensitive (for consistency with DBConstantOperatingSystem.class.php)

image

PR #589, opened 2019-07-05
small fix for transfers_table.php, purely aesthetic

belnet-ming:
Change from:
image
image

to:
image
image


monkeyiq:
Thanks for the PR. I thought it might be best to abstract out the maxColSpan calculation to the top of the file and use that variable in the HTML and have created and merged slightly rearranged version of the update from this PR in https://github.com/filesender/filesender/pull/601.

PR #590, opened 2019-07-08
adding a simple templating mechanism using the $config['theme'] variable

zivel:
This patch will change the path from where the template files are loaded from.
If no theme is set, the standard template files are loaded from the template folder.
If you set a theme (for example 'bootstrap') it will first check whether a template file is in the folder template/[themename] (ie. template/bootstrap) and if yes, uses this and if not uses the original from the template/ folder.


monkeyiq:
I updated the code a little into https://github.com/filesender/filesender/pull/604 which is now merged. It should work as this PR does with the 'theme' config setting.

PR #597, opened 2019-07-26
nitpicking fix: "cryptage" is incorrect

guillomovitch:
Correct french translation is "chiffrement".


monkeyiq:
Would you be happy to make the changes on poeditor so that they do not get forgotten in a future import from poeditor?  https://docs.filesender.org/v2.0/i18n/

guillomovitch:
I just did. There doesn't seem to be any reviewing process, tough.

monkeyiq:
I'm not sure how the collaborators for a language discuss things on poeditor to make sure nobody is stepping on anybodies foot. This point might be better to raise on filesender-dev to see if anybody has suggestions.

For this PR. I will reimport the whole French translation (and others) from poeditor before the next filesender release to keep the translations fresh with respect to what is on poeditor. As this update is on poeditor I will close this specific PR as it will be handled by an upcoming routine import.

PR #602, opened 2019-08-03
cron db: move to a subselect for the auditlog cleanup

monkeyiq:
This moves joins into a subselect. Tested on mariadb and postgresql locally.


monkeyiq:
I have also added a new issue https://github.com/filesender/filesender/issues/603 to try to catch issues in the cron jobs before new code is merged.

Issue #591, opened 2019-07-15
Dont re-use get_a_link transfer_option for (forced) guest_transfer_options


peter-:
For "normal" use of the system by authenticated subjects it makes sense to have the `get_a_link` setting within the `transfer_options`: The person de-/selecting that checkbox is a trusted user of the system and knows best how to communicate the URL resulting from an upload to the intended recipients of the uploaded files: Either by having Filesender notify the recipients with a generic email or by the uploading subject using whatever means they prefer.  
But that setting is currently re-used for the notification of third parties -- recpients of uploads performed by guests using a voucher issued to them -- where this isn't (or isn't always) appropriate, especially since those "Created transfers options" (which I think really means "transfer options forced upon guest voucher users") cannot be changed by the voucher user, contrary to the `get_a_link` setting within the `transfer_options` where the setting originates.

E.g. within the `transfer_options` I have `get_a_link` set up this way:
```php
'get_a_link' => array(
  'available' => true,
  'advanced' => false,
  'default' => true,
),
```
so an authenticated subject can decide how they themselfs prefer to notify the intended recpients.

But that same setting and default value does not make sense when applied in the context of creating guest vouchers with forced settings: Nothing is gained by preventing the voucher user (the uploader) from using Filesender to notify any intended recpients of uploaded files via email. The uploader (voucher user) can even post the Filesender download URL to Twitter (for all I know) so why disallow them to use Filesender to do that notification?

Even worse: Other sensible settings within `guest_options` such as `can_only_send_to_me` (to make sure at least one leg of a Filesender transfer is always tied to the community running the Filesender instance) now may have lost their chosen `'default' => true` setting since the `get_a_link` setting from the `transfer_options` takes prececends over this setting (and both cannot sensibly be true at the same time) and `can_only_send_to_me` cannot be selected as long as `get_a_link` is selected (which it is by default in our case, though not intended to limit how guest voucher users notify the intended recipients of their own uploads.)

So now people creating vouchers will have to remember to
1. deselect the `Get a link instead of sending to recipients` checkbox and then
2. select the `Can only send to me` checkbox

and I cannot change this with configuration alone.

In short: It does not make sense to re-use that same setting and value also for voucher users,  particularly as they cannot change that setting themselfs.  
[...]

PR #606, opened 2019-08-07
guests: improved interaction of get_a_link and send_to_me

monkeyiq:
This updates a bad interaction when the config.php shown below is used. In this case can_only_send_to_me will be checked in the interface and get_a_link will be unchecked by default.  There is now a notification if you subsequently select get_a_link letting you know that can_only_send_to_me has been changed to false and disabled. A subtle background colour is used to link these two options on the page as well. It is possible that a different user interface element such as a single selection drop down or list should be used to handle this setting making the choice explicit and not offering combinations that do not make sense.

```
$config['guest_options'] = array(
        'can_only_send_to_me' => array(
            'available' => true,
            'advanced' => false,
            'default' => true
        ),
```

This relates to https://github.com/filesender/filesender/issues/591.

### DIFF
--- a/language/en_AU/lang.php
+++ b/language/en_AU/lang.php
@@ -505,3 +505,5 @@ $lang['you_can_report_exception_by_email'] = 'You can report this error by email
 $lang['you_can_send_client_logs'] = 'In order to help your support team to find out what happened you can send the last log entries from your user interface by clicking this button :';
 $lang['archive_tar_download'] = 'Download as single (.tar) file';
 $lang['invalid_file_name'] = 'Invalid file name';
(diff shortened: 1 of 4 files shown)


monkeyiq:
I'll have to add that new string on poeditor before merging.

peter-:
OK, so you've essentially turned those two options into radio buttons (mutually exclusive choices: checking "get a link" also unchecks and disables "can only send to me")? A small issue with that is that it's asymmetrical:

filesender-2 8-pr606

I.e., the reverse (checking "can only send to me" which also unchecks "get a link") is not possible: One has to first uncheck the checked checkbox "get a link" only then becomes the "can only send to me" checkbox selectable again.

Issue #600, opened 2019-08-01
Can't easily resume interrupted upload


peter-:
If an upload is being interrupted due to "vis major" (e.g. your ISP or ADSL/cable model having issues, your WiFi being down or just out of reach, you fat-fingering your Ethernet cable out of the socket, etc.) there's no obvious way to resume such an upload, AFAICT:  
Once my network connection was interrupted no amount of hitting `pause`/`resume` will cause the upload to continue and the only other button being offered is `stop`, kindly offering to "remove already uploaded data" which is the least desirable option. (Imagine you're at 98% of a 100GB file when this happens, as such things tend to go.)

What seems to happen in such cases though is that the `nothing_happened_as_of_late_you_can_send_client_logs` message is shown at some point (after several minutes of inactivity) and I am being offered a button to `Send logs` somewhere unspecified. Clicking that button causes the following things to happen (Firefox 68):

* Render a (FF-specific?) JS overlay/popup with the text "This page is asking you to confirm that you want to leave - data you have entered may not be saved" offering the two choices `Stay on page` and `Leave page`. Picking either one does *not* send any logs anywhere from looking at the server's MTA's logs (on a machine fully configured with Filesender sending emails successfully otherwise).
* Even *before* chosing between those two options below the overlay I can see a green "Logs successfully sent" message when in fact nothing has been sent (AFAICT).
* Clicking `Stay on page` does not seem to do anything (which is probably expected but non-obvious given the browser's generic  and possibly unfortunate -- warning about potentially losing data when leaving the page).
* Clicking `Leave page` finally brings up yet another dialog asking me to "Restart failed transfer?". That now offers three buttons to click, the one for resuming the transfer now being called `Load failed transfer` (maybe overly literally), the other two being `Forget about it` and `Ask me again later`. "Forget about it" means "remove already uploaded data" (as it was put before then clicking the `stop` button) but is phrased phrased differently and somewhat ambiguous. (If you let me delete all my uploaded  data I'd appreciate being clear about this.)
* So eventually chosing `Load failed transfer` one can resume the interrupted upload (by re-selecting the same files as before and re-starting the upload, as one is being informed).  
[...]

PR #609, opened 2019-08-11
Updates for automatic resume of file uploads

monkeyiq:
Improved automatic resume for file uploads after a network interruption.

A number of configuration changes are made to set the default timeouts to be closer to the test environment. Note that upload_considered_too_slow_if_no_progress_for_seconds is for any upload progress, so a 30 second timeout should work even on a congested network because some bytes should have moved in that time frame.

The number of seconds since any data was sent is now shown on the upload page to provide insight to the user. Once that delay for no data sent reaches upload_considered_too_slow_if_no_progress_for_seconds then the upload will have a resume automatically scheduled.

A new reconnect and continue button was added to the upload page to allow the user to force a reconnection to the server and a resume. This could sort of be done before by pausing the transfer and clicking on the play button to resume though the reconnect button is a more heavy operation in that it will close terasender connections and respawn them in order to "reconnect" which may be needed if the laptop has changed IP address for example.

I had thought of trying to combine the pause/play and reconnect into a single interface but they are similar but different enough features that attempts to bring them into a single user interface would likely lead the user to trying to second guess what was happening (eg, does this pause/play trigger
a reconnection so that the change in wifi network will be fixed by doing it). The terasender_failed_after_many_retries message will now present as human readable when it happens. Failure to start terasender workers such as might happen if the network happens to be down at the time is now reported in a much nicer way.

Config changes
```
from   0 to 30 for upload_considered_too_slow_if_no_progress_for_seconds
from  10 to 50 for automatic_resume_number_of_retries
from 360 to 10 for automatic_resume_delay_to_resume
```


monkeyiq:
This was created to in response to https://github.com/filesender/filesender/issues/600

Issue #374, opened 2018-07-15
encryption: key is regenereated for each blob and could be cached instead


monkeyiq:
Each time a new chunk (blob) is read for upload the TeraSender (and I assume other chunk related code  paths) call crypto_app().encryptBlob() such as: https://github.com/filesender/filesender/blob/master/www/lib/terasender/terasender_worker.js#L139

Inside encryptBlob and decryptBlob this.generateKey() is called each time. Such as https://github.com/filesender/filesender/blob/master/www/js/crypter/crypto_app.js#L38

The encryption key should likely be cached rather than potentially regenerated each blob. This caching could be done completely inside generateKey() to avoid larger code changes https://github.com/filesender/filesender/blob/master/www/js/crypter/crypto_app.js#L23

fkooman:
https://github.com/filesender/filesender/issues/377

monkeyiq:
There is an issue with trying to do this in that each chunk is encoded with it's own IV. That IV has to be passed to deriveKey() to setup the cipher for the block. So it isn't possible with the API.

monkeyiq:
My above comment about the IV during keygen is how the code is but not how the code needs to be, for example the parameters for pbkdf2 do not include the encryption IV:

https://www.w3.org/TR/WebCryptoAPI/#pbkdf2-params

This unlocks the possibility to refactor the code and then sustain much higher pbkdf2 iteration counts with a once off (per terasender worker) impact on cipher performance during upload. And download speeds will be acceptable with larger files and high pbkdf2 round counts. The security should be unchanged by this optimization as it is just replacing many unnecessary calls to deriveKey with one call and reuse of the cached key. There are some other implementation details there for sure. The catch is realizing that the codepath is executed both in main context and web worker contexts so it has to be resilient to that, which may be what lead to the key handling in the current implementation.

Regarding the rolling your own crypto, replacing the whole crypto system and migrating away from WebCrypto to something else that provides an ultra high level API to out of order chunked uploads is likely something to be discussed in another issue, and is a matter for the filesender board to decide on.

monkeyiq:
This has been addressed in the above merges

PR #625, opened 2019-09-02
crypto: new password generation that is 256bit random data

monkeyiq:
A new password handling and encoding is added that allows passwords that are 256 bits random data to be used. These passwords are base64 encoded before being presented to the user for ease of transfer. The code stores both the way to handle the password and what encoding was used so that decryption sites can properly handle a text password. ie, it can be a plain text password as it was before or a value that needs to be base64 decoded before use. As the version information also capture if the password is a 256bit random value to crypto can also tune itself knowing that the password has good entropy.

The number of password hashing rounds is now stored in the database for each encrypted transfer. A new setting is brought out for adjusting that for new files. A new admin/testing page is created to experiment with the performance impact in the system of various password hashing settings before deploying changes.

The password box has been made larger to hint to the user that longer is better and allow the better presentation of random passwords.


docs/v2.0/admin/configuration/index.md
```diff
@@ -1039,6 +1040,28 @@ If you want to find out the expiry timer for your SAML Identity Provider install
...
+* __description:__ This setting only has an effect when encryption_key_version_new_files is 1 or above (the default for that setting). What follows is an abstract overview of how hash iterations may be used. When a text password that the user has entered is used for encryption a number of hashing rounds is applied to that password. The hashing uses both the user password and a random salt value combined. The salt is used so that the same password does not lead to the same key each time. It also means vast tables of precomputed keys can not be used for all transfers as the salt changes per transfer. The number of hashing rounds might be set for example so that deriving a key might take a second to perform. While that is not a huge delay for the user of the system, a potential attacker would have to spend some time to hash each guessed password. The attacker might have better or much newer hardware or use a GPU so that the time to has might be less than the second a user has spent. See https://en.wikipedia.org/wiki/PBKDF2#Key_derivation_process. To get an idea of how this setting will impact system performance see the admin/testing page on filesender.
```
vanrein: Those "vast tables" are usually called "rainbow tables".  Don't know where that name came from :-)

Please give a realistic estimate of the secure-until so that the most stringent applications know what to choose.  See the [formula and] table on Page 7.

Consider using the number of bits in the counter instead of a counter value to avoid that trap of falsely assuming a linear relationship where there really is an exponential one, due to computation power doubling in fixed times.

It should be very clear what responsibility a user is taking.  This text supports users in not acting responsibly, lulling them into a false sense of security, which is the last thing you should want with a security patch.  I am concerned that this low setting and the suggestion of linearity will lead to silly/insecure treatment of sensitive user data in places where it really matters.

If it was me, I would set a very high default too, and have the operator read this before they tune down the value.  I know that is not exactly the nicest way to demonstrate FileSender security to newcomers, through sharing the pain of passwords with them.  Then again, passwords must be discouraged for proper security practices, and making it a realistic alternative by default is not helpful in that respect.

includes/ConfigDefaults.php
```diff
@@ -226,6 +226,12 @@
     // see crypto_app.js for constants in the range crypto_key_version_constants
     // Generally higher is newer + better.
     'encryption_key_version_new_files' => 1,
+
+    // for details of possible values see crypto_password_version_constants
+    // in the file js/crypter/crypto_app.js
+    'encryption_random_password_version_new_files' => 2,
+    
+    'encryption_password_hash_iterations_new_files' => 150000,
```
vanrein: Suggest to reference the .md for an explanation, and/or warn against frivolously editing this value.  Again, this value is quite low, almost invalid at the present date (safe until 2020).  Passwords are really, really rotten for anything but the simplest form of security and should not be considered by any secure usage pattern, and it should not be a default, or if it is then at least at a secure level by default.  Security until 2020 is rediculously low.

vanrein:
Thinking about the number of rounds once more, it is probably best to be perfectly open by suggesting the year until which data should be secure.  The formula on Page 7 can help you compute the number of rounds.  A good setting might also be +20 to indicate "20 years from now" as an alternative to "2039" for "up to 2039".

This sounds to me like a much better option than debating what number of rounds, and not knowing (as a user) what they are doing.  A final year of security (or a number of years from "now") is the sort of things that their policies are likely to dictate, if they have any, and such cases would be properly supported with this.

The number of years may also inspire the size of the keys and choice of encryption algorithms.  All in one go, without getting technical.  You would be encapsulating the technicalities of the crypto into your code, which is the right place for it.

includes/ConfigDefaults.php
monkeyiq: I added (once CI passes) the reference with the introduction of https://github.com/filesender/filesender/pull/649. I think the https://docs.filesender.org/v2.0/admin/configuration/#encryption_password_hash_iterations_new_files link is where most admins will be looking to start with, but if it helps somebody who found this value directly then it is well worth putting the link in there.

monkeyiq:
I like this number of years approach too. Much less abstract than the number of rounds. It is easy for the code to set the internal iteration count when loading the configuration and use a most friendly "number of years" setting for admins to set inline with their policy. I will defer to the filesender board as to what the default number of years should be for this setting.

monkeyiq:
I recently added a 'testing' section to the admin page to allow investigation of things like the number of hashing iterations on performance. Yes, I said the nasty performance word which is not to put it above security. Though a sys admin might like to see that a really huge number of iterations will take 30 minutes on the machines that their users have at hand so they will not find that out later when the users start complaining. Or worse, the users not encrypting any more because that is too slow.

So, on a recent i7 laptop, 16384000 iterations takes 9.5 seconds on Chrome and 24.8 on Firefox. It might not be a bad assumption to make to imagine that an attacker might be using a more optimized implementation.

I had noticed that the key handling regenerates the same key for each filesender chunk https://github.com/filesender/filesender/issues/374. This would need to be refactored to sustain reasonably high pbkdf2 iteration counts while maintaining acceptable upload/download performance. 

If the PBKDF2 is only done once per terasender worker then a high iteration count can be explained to the user and the upload should proceed with similar performance once the first key derivation per worker has completed. One upside is that the UI explaining the delay can also inform the user that the delay would not happen if they elect to use a random password.

Issue #557, opened 2019-05-20
cron.php error no_such_property


fransward:
When running the cron.php we see the following error:
`php /opt/filesender/filesender/scripts/task/cron.php`

Error:

> PHP Fatal error:  Uncaught PropertyAccessException: no_such_property in /usr/share/filesender/filesender-2.61/classes/data/StatLog.class.php:365
> Stack trace:
> #0 /usr/share/filesender/filesender-2.61/classes/data/DBObject.class.php(558): StatLog->__get('browser')
> #1 /usr/share/filesender/filesender-2.61/classes/data/StatLog.class.php(166): DBObject->toDBData()
> #2 /usr/share/filesender/filesender-2.61/classes/data/StatLog.class.php(299): StatLog->save()
> #3 /usr/share/filesender/filesender-2.61/scripts/task/cron.php(52): StatLog::createGlobal('global_storage_...', 92362768384)
> #4 {main}
>   thrown in /usr/share/filesender/filesender-2.61/classes/data/StatLog.class.php on line 365

Have you seen this before?

keros:
Same here (#88)

peter-:
Yes, Frans (and keros) this also affects us exactly the same way on a fresh install of branch `filesender-2.7`, using PostgreSQL.

If the cron job is what expires files (and sends email notifications about this) I guess this is a total show stopper that prevents any real use of the software.  
(At least I'm not looking forward to sending out hundreds of such notifications after an eventual fix is deployed, months after people have uploaded something in the expectation -- in fact promise, as per the `/?s=privacy` page -- that it would expire a few days later. Not to mention that disk capacities are not unlimited and without expiration soon any disk would be filled, making this an operational issue, too.

kadafax:
Same problem here, it seems :
https://github.com/filesender/filesender/commit/d5981869cc789692428f86d2a9002b15120477ba
and
https://github.com/filesender/filesender/pull/586/commits/5ea92c552420cc19bfd8961db40be66cfe9cb227

fixed the problem

yent:
> Same problem here, it seems :
> [d598186](https://github.com/filesender/filesender/commit/d5981869cc789692428f86d2a9002b15120477ba)
> and
> [5ea92c5](https://github.com/filesender/filesender/commit/5ea92c552420cc19bfd8961db40be66cfe9cb227)
> 
> fixed the problem

Same here using mysql

peter-:
[...]

PR #631, opened 2019-09-04
Amend #629 w/ proper fix for #557, as per @fkooman

peter-:
Thanks for catching this @fkooman, all the constants in the two changed files' paths in #629 must have confused me. ;) Hopefully with this PR issue #557 can be closed for good.

### DIFF
--- a/classes/data/constants/DBConstantBrowserType.class.php
+++ b/classes/data/constants/DBConstantBrowserType.class.php
@@ -84,7 +84,7 @@ class DBConstantBrowserType extends DBConstant
     static function currentBrowserToEnum()
     {
         $b = $_SERVER['HTTP_USER_AGENT'];
-        if(!defined($_SERVER['HTTP_USER_AGENT'])) {
+        if(!isset($_SERVER['HTTP_USER_AGENT'])) {
             return self::lookup(self::T_UNKNOWN);
         }
         if( preg_match( '/Firefox/i', $b )) {
--- a/classes/data/constants/DBConstantBrowserType.class.php
+++ b/classes/data/constants/DBConstantBrowserType.class.php
@@ -83,10 +83,10 @@ class DBConstantBrowserType extends DBConstant
 
     static function currentBrowserToEnum()
     {
-        $b = $_SERVER['HTTP_USER_AGENT'];
         if(!isset($_SERVER['HTTP_USER_AGENT'])) {
             return self::lookup(self::T_UNKNOWN);
         }
+        $b = $_SERVER['HTTP_USER_AGENT'];
         if( preg_match( '/Firefox/i', $b )) {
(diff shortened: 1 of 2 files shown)


fkooman:
`$b = $_SERVER['HTTP_USER_AGENT'];`

should this be done _after_ the check?

peter-:
@fkooman: Sry, fixed. I haven't really been paying attention, paradoxically due to the triviality of the PR.
@monkeyiq: The failing test seems unrelated.

monkeyiq:
The CI system is not as good as one might like. If one of the selenium test runs fails by itself and you are not deep into db specific code in the PR it is a fairly good indication that it is a CI execution issue rather than a code issue. Both the selenium runs do the same thing but use mysql or postgresql as the backing store just in case there are db issues in the PR.

monkeyiq:
generally the "restart build" button on the travis-ci details link from this page is useful as a first retry in that case. I've just fired that off to see what happens in 15 minutes.

peter-:
I noticed another instance of this in our web server error logs, recently:

```
[Mon May 04 05:17:00.828833 2020] [proxy_fcgi:error] [pid 2642:tid 140379512887040] [client 74.82.47.2:54976] AH01071: Got error '
PHP message: PHP Notice:  Undefined index: HTTP_USER_AGENT in /opt/filesender/classes/utils/GUI.class.php on line 86
PHP message: PHP Notice:  Undefined index: HTTP_USER_AGENT in /opt/filesender/classes/utils/GUI.class.php on line 87
PHP message: PHP Notice:  Undefined index: HTTP_USER_AGENT in /opt/filesender/classes/utils/GUI.class.php on line 92'
```

Caused by a request from a [bot](https://en.wikipedia.org/wiki/Shadowserver#Data_collection):

```
74.82.47.2 - "" [04/May/2020:05:17:00 +0000] "GET / HTTP/1.1" 200 18050 "-" "-"
```

Something like this should do to avoid `browser_is_*` functions from even getting called when there is no UA header:

```diff
diff --git a/classes/utils/GUI.class.php b/classes/utils/GUI.class.php
index fd69f022..3fbdab79 100644
--- a/classes/utils/GUI.class.php
+++ b/classes/utils/GUI.class.php
@@ -94,7 +94,11 @@ class GUI
 
     public static function use_webasm_pbkdf2_implementation()
     {
-        return Utilities::isTrue(GUI::browser_is_ie11() || GUI::browser_is_edge());
+        if (isset($_SERVER['HTTP_USER_AGENT'])) {
+            return Utilities::isTrue(GUI::browser_is_ie11() || GUI::browser_is_edge());
+        } else {
+            return false;
+        }
     }
     
     /**
```
@monkeyiq Do you need another PR for this?

PR #640, opened 2019-09-14
crypto: add new AES-256-GCM encryption modes

monkeyiq:
This adds two new key version modes for AES-256-GCM encryption. These mirror the existing AES-256-CBC modes but use GCM (see https://en.wikipedia.org/wiki/Galois/Counter_Mode) for more details than could easily be described here. The implementation uses a design proposed by a cryptographer where the IV is derived using a combination of entropy and a chunk counter to avoid the possibility of IV reuse across chunks.

This has been tested as working across Firefox, Chrome, and Safari. Tests were performed also uploading in one browser and downloading in the others across these choices. These modes do not work in Edge or IE. 

Some more documentation will follow for the encryption_key_version_new_files settings 2 and 3 which enable these new modes using digest_importKey and importKey_deriveKey respectively. The main difference in modes 2 and 3 is the explicit use of PBKDF2 in version 3 in key derivation for user supplied passphrases. If you have the choice, always use mode 3 over mode 2.

### DIFF
--- a/classes/data/File.class.php
+++ b/classes/data/File.class.php
@@ -241,8 +241,18 @@ class File extends DBObject
         if ($lastChunkPadding == 0) {
             $lastChunkPadding = 16;
         }
-            
-        return $this->size + ($chunksMinusOne * $echunkdiff) + $lastChunkPadding + 16;
+
+        switch( $this->transfer->key_version ) {
+            case CryptoAppConstants::v2018_importKey_deriveKey:
+            case CryptoAppConstants::v2017_digest_importKey:
+                return $this->size + ($chunksMinusOne * $echunkdiff) + $lastChunkPadding + 16;
+            case CryptoAppConstants::v2019_gcm_importKey_deriveKey:
+            case CryptoAppConstants::v2019_gcm_digest_importKey:
+                return $this->size + (($chunksMinusOne+1) * $echunkdiff);
+            default:
+        }
+        // fall through is an error
+        throw new BadCryptoKeyVersionException( $this->transfer->key_version );
     }
 
     /**
--- a/www/js/client.js
+++ b/www/js/client.js
@@ -374,7 +374,8 @@ window.filesender.client = {
                 return uxhr;
             }
         };
-        
+
+        var chunkid = Math.floor(offset / window.filesender.config.upload_chunk_size);
         var $this = this;
         if(encrypted){
             var cryptedBlob = null;
@@ -386,6 +387,7 @@ window.filesender.client = {
             blobReader.readArrayBuffer(function(arrayBuffer){
                 window.filesender.crypto_app().encryptBlob(
                     arrayBuffer,
+                    chunkid,
                     encryption_details,
                     function(encrypted_blob) {
                         var result = $this.put(
--- a/www/js/crypter/crypto_app.js
+++ b/www/js/crypter/crypto_app.js
@@ -24,8 +24,10 @@ window.filesender.crypto_app = function () {
             // newest version first, some metadata about the process
             // taken. The year (and maybe month) should give indication
             // that the later years are also the most desired version.
-            v2018_importKey_deriveKey: 1,
-            v2017_digest_importKey:    0
+            v2019_gcm_importKey_deriveKey: 3, // AES-GCM otherwise same as v2018_importKey_deriveKey
+            v2019_gcm_digest_importKey:    2, // AES-GCM otherwise same as v2017_digest_importKey
+            v2018_importKey_deriveKey:     1, // AES-CBC
+            v2017_digest_importKey:        0  // AES-CBC
         },
         crypto_password_version_constants: {
             // constant values for crypto_password_version_constants
@@ -51,11 +53,45 @@ window.filesender.crypto_app = function () {
             v2019_generated_password_that_is_full_256bit: 2
         },
 
+        createChunkIDArray: function( chunkid ) {
+            var ret = new Uint8Array(4);
+            ret[0] = chunkid>>0  & 0xFF;
+            ret[1] = chunkid>>8  & 0xFF;
+            ret[2] = chunkid>>16 & 0xFF;
+            ret[3] = chunkid>>24 & 0xFF;
+            return ret;
+        },
+
+        /**
+         * Create and return an IV of 16 octets (128 bits) constructed as follows:
+         *    12 octets of entropy
+         *     4 octets containing the chunkid
+         *
+         * This is based on Page 19 of OpenFortress 2018 document:
+         *   "The suggested procedure for the case of FileSender is to combine 96 bits
+         *    of random material with a 32-bit chunk counter to form a 128-bit IV."
+         */
+        createIVGCM: function( chunkid ) {
+            var $this = this;
+
+            // 96 bits of entropy
+            ivrandom  = crypto.getRandomValues(new Uint8Array(12));
+
+            // 32 bits of counter from chunkid
+            ivcounter = $this.createChunkIDArray(chunkid);
+
+            // merge these into return value
+            iv = new Uint8Array(16);
+            iv.set(ivrandom);
+            iv.set(ivcounter, ivrandom.length );
+            
+            return iv;
+        },
         
         generateVector: function () {
             return crypto.getRandomValues(new Uint8Array(16));
         },
-        generateKey: function (encryption_details, callback) {
+        generateKey: function (chunkid, encryption_details, callback) {
             var iv = this.generateVector();
             var $this = this;
             var password    = encryption_details.password;
@@ -71,13 +107,14 @@ window.filesender.crypto_app = function () {
                 hashRounds = encryption_details.password_hash_iterations;
             }
             var saltBuffer     = window.filesender.crypto_common().convertStringToArrayBufferView(salt);
+            var efunc = function (e) {
+                // error making a hash
+                filesender.ui.log(e);
+            };
+
             
             if( key_version == $this.crypto_key_version_constants.v2018_importKey_deriveKey )
             {
-                var efunc = function (e) {
-                    // error making a hash
-                    filesender.ui.log(e);
-                };
 
                 crypto.subtle.importKey(
                     'raw', 
@@ -107,6 +144,45 @@ window.filesender.crypto_app = function () {
                 }, efunc );
             }
 
+            if( key_version == $this.crypto_key_version_constants.v2019_gcm_importKey_deriveKey )
+            {
+                window.filesender.config.crypto_crypt_name = "AES-GCM";
+                this.crypto_crypt_name = window.filesender.config.crypto_crypt_name;
+
+                // IV has a predefined mix of entropy and chunk counter
+                iv = $this.createIVGCM(chunkid);
+
+                crypto.subtle.importKey(
+                    'raw', 
+                    passwordBuffer,
+                    {name: 'PBKDF2'}, 
+                    false, 
+                    ['deriveBits', 'deriveKey']
+                ).then(function(dkey) {
+
+                    crypto.subtle.deriveKey(
+                        { "name": 'PBKDF2',
+                          "hash": 'SHA-256',
+                          "iterations": hashRounds,
+                          "salt":       saltBuffer,
+                        },
+                        dkey,
+                        { "name":   'AES-GCM',
+                          "length": 256
+                          // Note that passing the IV here does nothing as we are not encrypting anything
+                          // tested by passing a random value here that is not the same during a call from
+                          // decryptBlob()
+                        },
+                        false,                   // key is not extractable
+                        [ "encrypt", "decrypt" ] // features desired
+                    ).then(function (key) {
+                    
+                        callback(key, iv);
+                    }, efunc );
+                }, efunc );
+
+            }
+            
             if( key_version == $this.crypto_key_version_constants.v2017_digest_importKey )
             {
                 crypto.subtle.digest(
@@ -129,11 +205,44 @@ window.filesender.crypto_app = function () {
                     filesender.ui.log(e);
                 };
             }
-        },
-        encryptBlob: function (value, encryption_details, callback) {
-            var $this = this;
+
+
+            if( key_version == $this.crypto_key_version_constants.v2019_gcm_digest_importKey )
+            {
+                window.filesender.config.crypto_crypt_name = "AES-GCM";
+                this.crypto_crypt_name = window.filesender.config.crypto_crypt_name;
+
+                // IV has a predefined mix of entropy and chunk counter
+                iv = $this.createIVGCM(chunkid);
+                
+                crypto.subtle.digest(
+                    {name: this.crypto_hash_name},
+                    passwordBuffer
+                ).then( function (key) {
+                    crypto.subtle.importKey("raw", key,
+                                            { "name":   'AES-GCM', "length": 256 },
+                                            false,
+                                            ["encrypt", "decrypt"]
+                                           ).then( function (key) {
+                                               callback(key, iv);
+                                           }, function (e) {
+                                               // error making a key
+                                               filesender.ui.log(e);
+                                           });
+                }),
+                function (e) {
+                    // error making a hash
+                    filesender.ui.log(e);
+                };
+            }
             
-            this.generateKey(encryption_details, function (key, iv) {
+            
+        },
+        encryptBlob: function (value, chunkid, encryption_details, callback) {
+            var $this = this;
+
+            
+            this.generateKey(chunkid, encryption_details, function (key, iv) {
                 crypto.subtle.encrypt({name: $this.crypto_crypt_name, iv: iv}, key, value).then(
                         function (result) {
 
@@ -165,9 +274,11 @@ window.filesender.crypto_app = function () {
             var blobArray = [];
 
             try {
-                this.generateKey(encryption_details, function (key) {
+                var chunkid = 0;
+                this.generateKey(chunkid, encryption_details, function (key) {
 		    var wrongPassword = false;
 		    var decryptLoop = function(i) {
+                        
 		        callbackProgress(i,encryptedData.length); //once per chunk
                         var value = window.filesender.crypto_common().separateIvFromData(encryptedData[i]);
                         crypto.subtle.decrypt({name: $this.crypto_crypt_name, iv: value.iv}, key, value.data).then(
@@ -359,7 +470,7 @@ window.filesender.crypto_app = function () {
             var $this = this;
             var ret = new Object();
             var raw = new Uint8Array(0);
-            
+
             if( $this.crypto_password_version_constants.v2018_text_password == version ) {
                 raw = window.filesender.crypto_common().convertStringToArrayBufferView(value);
             }
--- a/www/js/terasender/terasender.js
+++ b/www/js/terasender/terasender.js
@@ -112,10 +112,13 @@ window.filesender.terasender = {
         if(!file) return null; // Nothing to do
         
         if(!file.endpoint) file.endpoint = this.transfer.authenticatedEndpoint(filesender.config.terasender_upload_endpoint.replace('{file_id}', file.id), file);
+
+        var chunkid = Math.floor(file.uploaded / filesender.config.upload_chunk_size);
         
 	if (typeof file.fine_progress_done === 'undefined') file.fine_progress_done=file.uploaded; //missing from file
         var job = {
             chunk: {
+                id: chunkid,
                 start: file.uploaded,
                 end: Math.min(file.uploaded + filesender.config.upload_chunk_size, file.size) //MD last chunk was too big
             },
--- a/classes/constants/CryptoAppConstants.class.php
+++ b/classes/constants/CryptoAppConstants.class.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * FileSender www.filesender.org
+ *
+ * Copyright (c) 2009-2014, AARNet, Belnet, HEAnet, SURFnet, UNINETT
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * *	Redistributions of source code must retain the above copyright
+ * 	notice, this list of conditions and the following disclaimer.
+ * *	Redistributions in binary form must reproduce the above copyright
+ * 	notice, this list of conditions and the following disclaimer in the
+ * 	documentation and/or other materials provided with the distribution.
+ * *	Neither the name of AARNet, Belnet, HEAnet, SURFnet and UNINETT nor the
+ * 	names of its contributors may be used to endorse or promote products
+ * 	derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS 'AS IS'
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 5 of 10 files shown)


www/js/crypter/crypto_app.js
```diff
@@ -51,11 +53,45 @@ window.filesender.crypto_app = function () {
...
+
+        /**
+         * Create and return an IV of 16 octets (128 bits) constructed as follows:
+         *    12 octets of entropy
+         *     4 octets containing the chunkid
+         *
+         * This is based on Page 19 of OpenFortress 2018 document:
+         *   "The suggested procedure for the case of FileSender is to combine 96 bits
+         *    of random material with a 32-bit chunk counter to form a 128-bit IV."
+         */
```
vanrein: Please also note this remark on the bottom of Page 19:

  * The 96 bits worth of random material in the IV are produced from a secure random generator, and kept constant for one file (or one upload, as desired). This 96-bit value can be used to see if chunks belong to the same file or upload.

If I understand correctly, you are generating random material for each chunk independently.  As a result, you cannot checking the 96 bits to match upon recollection of chunks.  This means that chunks might be taken out of one upload and substituted in another upload.

www/js/terasender/terasender.js
```diff
@@ -112,10 +112,13 @@ window.filesender.terasender = {
         if(!file) return null; // Nothing to do
         
         if(!file.endpoint) file.endpoint = this.transfer.authenticatedEndpoint(filesender.config.terasender_upload_endpoint.replace('{file_id}', file.id), file);
+
+        var chunkid = Math.floor(file.uploaded / filesender.config.upload_chunk_size);
```
vanrein: Please note the remark on Page 19,

  * It is vital for counter mode encryption that the 32-bit counter will never ever wrap around.

I chased your code a bit, but could not find a place where you constrain the size of the upload to the given number of uploads.  Please note that the last chunk may not be full, as that would lead to a need to send another, so the maximum number of chunks is 2^32 and that last chunk must not be full-sized.  (I am assuming a fixed chunk size.)

classes/data/File.class.php
```diff
@@ -241,8 +241,18 @@ private function calculateEncryptedFileSize()
         if ($lastChunkPadding == 0) {
             $lastChunkPadding = 16;
         }
-            
-        return $this->size + ($chunksMinusOne * $echunkdiff) + $lastChunkPadding + 16;
+
+        switch( $this->transfer->key_version ) {
+            case CryptoAppConstants::v2018_importKey_deriveKey:
+            case CryptoAppConstants::v2017_digest_importKey:
+                return $this->size + ($chunksMinusOne * $echunkdiff) + $lastChunkPadding + 16;
```
vanrein: Not sure if your math is completely correct in light of this remark on Page 20:

  * The last chunk is sent, and considered the closing chunk if it is partial. If the last chunk is a full-sized chunk, then a closing chunk with zero bytes of content is added, with the chunk counter incremented once more.

If I have `$this->size=100` and `$upload_chunk_size=10`, I would get `$chunksMinusOne=9` which, if I understand your code correctly, should have been `10` instead.  The number of chunks should match the situation for `$this->size=101`, not `$this->size=99` as seems to be the case here.

www/js/client.js
```diff
@@ -374,7 +374,8 @@ window.filesender.client = {
                 return uxhr;
             }
         };
-        
+
+        var chunkid = Math.floor(offset / window.filesender.config.upload_chunk_size);
```
vanrein: Please cap it off to ensure no more than 2^32-1 chunks are sent, or if 2^32 are sent that at least the last one is not a full chunk.  This is vital for the security of the encryption, as stated on Page 19:

  * It is vital for counter mode encryption that the 32-bit counter will never ever wrap around.

www/js/crypter/crypto_app.js
```diff
@@ -165,9 +274,11 @@ window.filesender.crypto_app = function () {
             var blobArray = [];
 
             try {
-                this.generateKey(encryption_details, function (key) {
+                var chunkid = 0;
+                this.generateKey(chunkid, encryption_details, function (key) {
 		    var wrongPassword = false;
 		    var decryptLoop = function(i) {
+
```
vanrein: This may be the place to verify that all IV values share the same 96-bit random value to indicate that it belongs to the same file, as stated on Page 19:

  * The 96 bits worth of random material in the IV are produced from a secure random generator, and kept constant for one file (or one upload, as desired). This 96-bit value can be used to see if chunks belong to the same file or upload.

vanrein:
Another note: I don't believe I've seen you check the last chunk to be partial, and yet, correctly signed  so it is a proven end of the file.  I also may have overlooked, but have missed checks that the initial chunk is numbered 0 and that the following chunks increment the chunk number by one.  Page 20 says this:

  * The counter in the IV starts at zero, and is incremented by one for every next chunk.

Further down I explain part of why this is necessary and useful,

  * Such an IV achieves cryptographic binding because any change to the IV would be clearly noticeable, as it causes invalid decoding outcomes.

www/js/crypter/crypto_app.js
```diff
@@ -51,11 +53,45 @@ window.filesender.crypto_app = function () {
...
+            iv.set(ivcounter, ivrandom.length );
+            
+            return iv;
+        },
         
         generateVector: function () {
             return crypto.getRandomValues(new Uint8Array(16));
         },
-        generateKey: function (encryption_details, callback) {
+        generateKey: function (chunkid, encryption_details, callback) {
```
vanrein: There is a problem in mapping the (upload-termed) chunk to a (GCM-termed) chunk counter.  This is probably a mistake or unclarity in my 2018 document that I need to correct.

GCM uses the IV as a counter *of 16-byte chunks* and so, if you upload a larger sequence of bytes, you should change the counter to reflect the number of 16-byte chunks instead of incrementing by one.  So, if you upload 4 kB chunks, then your chunk counter in the IV should increment by ceil(4096/16) = 256 on each round.  (I would only accept multiples of 16 bytes by the way, to avoid trouble on that end, though signatures would catch out any problems with that.)

Perhaps we should find a way that I can trace this logic through your code, to help you do this under the constraints of the encryption algorithms used.

www/js/client.js
vanrein: Take note that this means a maximum file size of 16 * 2^32 - 1 -- in light of chunks being the 16-byte strings that pass through AES.

www/js/crypter/crypto_app.js
vanrein: This is also the place to verify that the counter portion rises as expected.

www/js/terasender/terasender.js
vanrein: Given that these chunks are always the 16-byte strings handled by a single AES iteration, the maximum size is 16 * 2^32 - 1.

monkeyiq: I think the chunk term is general enough that it is used in multiple different ways in different contexts, so it might be a good idea to clarify the above use. Normally FileSender sends files in chunks which by default are 4mb in size. This would allow for 2^32 chunks at 4mb a pop which is around 16tb before the chunk id wraps. 

This is also why I have used a new 96 bits of random material for the IV in each "chunk" as they are sort of like little 4mb files. Though I have it earmarked to move to the single 96 bits for the whole file.

There is a current limitation with uploads for encrypted files having to be less than 4tb in size. But the code should also check for the GCM limit in case the 4tb limit gets raised in the future. I'll fix that shortly.

classes/data/File.class.php
monkeyiq: These two are the existing AES-CBC modes and I've left their chunk calculations intact.

www/js/crypter/crypto_app.js
monkeyiq: The chunkid passed in the above cited code is the floor( offset / chunksize ) calculated in the below:
https://github.com/filesender/filesender/blob/master/www/js/terasender/terasender.js#L116

That is eventually passed to createIVGCM() to make an IV.
https://github.com/filesender/filesender/blob/master/www/js/crypter/crypto_app.js#L74

Setting aside me moving the 96 bits of random material to the same 96 bits for a file. The chunkid is encoded by createIVGCM() using createChunkIDArray().
https://github.com/filesender/filesender/blob/master/www/js/crypter/crypto_app.js#L56

At the moment a chunk of 0 will be encoded as 0
A chunk of 1 (4mb into the file) will be encoded as 1
and so on.

To clarify, after moving to chunkid*ceil(chunk-size-in-bytes / 16) from above we should encode 262144 when passed chunkid=1 by FileSender. And for chunkid=2 then 524288 is encoded. And the max_file_size is limited as per your above limit.

www/js/terasender/terasender.js
monkeyiq: I got to your comment at the end of page, I'll limit to 16 * 2^32 - 1

vanrein:
Your patch does not yet incorporate AEAD as a strengthening mechanism.  As written on Page 14,

  * One reason why we prescribed integrity checks for an encryption mode is that these allow the cryptographic binding of data into an encrypted piece of data. This is why we advise to retire the CBC mode and replace it with GCM. Cryp- tographic binding means that it is not just stored in the same data set, but that it cannot be altered without being noticed. This can be done through an AEAD mechanism, which means that extra data can be signed into the vali- dation mechanism without encrypting it; the verifier must have the same data available to be able to decrypt and validate. In the case of GCM, it is possible to use the IV to store the data required, as described below. This fulfils the same role in GCM as a nonce or AEAD string.

This could involve such things as the sender and file name, a random code used for storing it, perhaps the size, a description text and so on.  Binding this into the signature through AEAD causes this data to be part of the validation scheme of AES-256-GCM.

Can you please suggest metadata that belongs with the upload and that is also available during download?  We might place this in a data structure, serialise it in a canonical manner (your code probably does that already) and incorporate that before or after chunk encoding.  Note that recipients are *not* possible in this metadata, as that would require each recipient to know of all recipients for a file which, I suspect, is considered too private.

vanrein:
I suggest a test of the parallel usabilty of the scheme as you implemented it.  This is another strength of the GCM mode.

Basically, the IV contains a counter, which the underlying API should increment by 1 for each continued 16 byte block.  You could test the same by sending in 32 bytes with a given counter value, and the second block of 16 bytes with the same counter value plus one (chunkid, is what you call those in the current code).  Then, ensure that the 16 bytes of overlap encode to the same encrypted data.  The first block is dropped, and so the signature is different of course.  This test only finds the same 16 bytes for the second block.  If that works, you can rest assured that future parallisation of uploads and downloads will work, if you ever decide to go for those.

vanrein:
Can you please point out to me where the last block is encrypted, and where it is decrypted?  I cannot tell exactly how the code hangs together, and this needs checking.  As written on Page 20,

  * The last chunk is sent, and considered the closing chunk if it is partial. If the last chunk is a full-sized chunk, then a closing chunk with zero bytes of content is added, with the chunk counter incremented once more.

vanrein:
Can you please point out to me where the GMAC signature is passed, and where it is validated?  This may be part of the WebCrypto API, but I cannot tell exactly how the code hangs together, and this needs checking.

www/js/terasender/terasender.js
vanrein: Changes such as fresh random material for each "chunk" have serious ramifications for the security of the system.  You need a mechanism to validate that it is all part of the same file.  AEAD might be another mechanism, and I could look into that, but **do not make assumptions of this kind** as they endanger the fine balance under which the cipher mode was designed.

The thing is, I am fencing off cases that are invisible to you, and that involve a rogue party aware of your code structure.  Assumptions like these (and not being able to check that the "chunks" belong to the same upload) seriously endanger protection against such parties.  And you could never tell from tests that these holes existed.  That's my responsibility, and the cause of specifying these validation rules and IV formats in a very particular way.  There can be variations, but implementation choices are the [only] driving force for security code, and such changes must not be made silently, for danger of opening security holes.

classes/data/File.class.php
vanrein: Understood that you leave the existing code intact.  The comment relates to the usefulness of the code somewhat higher up, where I could not add remarks.  Can you please convince me that when the last chunk is full (4 MB is what you mentioned as a chunk size) that there will be one more chunk, sized 0, so that the partial send is a proof that the file submission ends there?

www/js/crypter/crypto_app.js
vanrein: Yes to your clarification!  Multiply the chunkid (0, 1, 2...) by 262144 (0, 262144, 524288, ...) to form the 32-bit counter value.  Indeed, under the max_file_size being 68719476735 or 16 * 2^32 - 1.

classes/data/File.class.php
monkeyiq: I dont think the web crypto api allows you to play around at the "16 byte low level chunk" size. Which we may perhaps refer to as AES-chunks. I would expect trying to encrypt a zero byte array of data on the web crypto api to produce undefined behavior on one or more browsers.

I have tested the code on 4mb and 4mb+1 byte sized files. The former results in a single filesender chunk and the later results in two filesender chunks. Note that the actual byte of the ciphered content is different as per these calculations.

What is your cryptographers opinion of adding a known pad to an exactly 4mb final filesender chunk of data? I could append a known cleartext to force the web crypto api to add a new chunk.

vanrein:
You are supporting a number of possible forms of encryption.  This means that we should be mindful of downgrade attacks.  Within the range of supported options, it is always understandable if the lowest form is negotiated, this cannot be stopped.  The range of options however, should be limited.  Is there a configuration option for administrators to select the minimally acceptable format?  One form you might consider is "enforce option X after data D" though that is perhaps innovative :-)

vanrein:
What you are trying to do with per-chunk random 96 bits is possible, treating the 4MB-chunks as separate files, but not in isolation.  We do need to avoid that chunks can be swapped in and out of a transfer without being detected.

As suggested above, AEAD can give us a hand here.  That's extra information hashed into the checksum of GCM, so the GMAC.  You can think of GMAC as a secure hash, but highly optimised for use with the counter mode in GCM.  Anything that you supply to a hash must be in such a format that the individual bits of information can be extracted separately (otherwise you'd loose information before hashing, and give rise to attacks).  For AEAD, the end of the data should also be clear from the data itself, so we have a proper definition of where the encrypted data starts.

 1. You are creating a unique identity for each file, right?  And it is (at least) 128 bit random?  Is that a source of randomness with good entropy (a secure pseudo-random generator)?  If so, we can use that as AEAD data going into each 4MB-chunk.

 2. You have a counter for the 4MB-chunk, your `chunkid` parameter.  You can pass that into each chunk as AEAD data too.  Doing so means that the order is available, and implicitly checked without you needing to do it explicitly; decryption will fail if you fed it with different data.  You want to use a canonical form (strings "0" and "0000" hash differently, after all) but any form will do, including textual representation with an end marker.  A binary size would also work, but imply limitations to the size that FileSender generally likes to avoid.

 3. The AEAD information must be supplied during both encryption and decryption, so the context should provide it.  The information from the context is only validated through AEAD, which is its purpose.  This is a strong implementation of the validation that you would otherwise have to do by taking the IV apart.  More information, such as the (validated) sender and the filename and perhaps description can be taken along too, given the must-be-able-to-parse-elements-out rule.

 4. Given this, you can now remove the chunk number from the IV.  In fact, your IV can be a 128-bit random number generated specifically for your 4MB-chunk.  You don't need to check its contents anymore, as that is done through the AEAD mechanism.  One major result is that the maximum file size limitation is gone; we can rest assured that a random IV, provided that it is generated by a proper source of pseudo-entropy that cannot be influenced by rogue code (like an adverse advertisement) is unique and so, that the same IV will not occur multiple times.  We should limit the 4MB-chunk to the 32-bit size limit of 16 * 4 GB, but that is not a constraint on the total file size.

Good thinking, Ben.  But please don't rush into such changes by making local implementation choices!  A fine balance is required for this to work.  Yeah, I'm all into details I know, but as explained that's because I try to control for rogue parties that are aware of the FileSender code and protocol and want to use it badly, so I need to focus on things that no test will show, but that should be caught out with checks or invariants in your code.  I am here to listen to your complaints or suggestions.

vanrein:
> I suggest a test of the parallel usabilty of the scheme as you implemented it. This is another strength of the GCM mode.

This suggestion can be dropped when we adopt an IV of 128 random bits, and supplying other information via AEAD.  If we do that, concurrency is implied.

monkeyiq:
I have https://github.com/filesender/filesender/pull/644 to update the IV handling.

classes/data/File.class.php
monkeyiq: I was playing around with this using a upload_chunk_size=`4*1024*1024`
One of my testing sample files is 30091892 bytes. I have been testing in both contiguious and chunked storage. The later is handy because FileSender will store each filesender chunk in a different file.

This results in 7 full sized filesender chunks and one partial chunk. Each chunk has 32 bytes of overhead added by the encryption. The first 7 will consume a fill 4mb of data each and the final chunk will have:
```
30091892 - (7*4*1024*1024)
731764
```

The on disk storage for this encrypted file requires 30092148 bytes. This makes sense when one calculates the padding for the 7 full and 1 partial chunk.

```
(4*1024*1024+32)*7 + 731764+32
30092148
```

I assume in the initial comment above in the part "The last chunk is sent, and considered the closing chunk if it is partial." you are talking about the chunks used by the AES algorithm and not the 4mb filesender chunks.

monkeyiq:
As far as downgrade attacks go the GCM modes are enabled by setting encryption_key_version_new_files to 2 or 3 on config.php. The code is then forcing many options to specific values, see for example:
https://github.com/filesender/filesender/blob/master/www/js/crypter/crypto_app.js#L149

Though you may also have had other downgrade attacks in mind as well?

monkeyiq:
The details of what the WebCrypto API is doing at the lower level are detailed in the NIST document cited in step 6 of the definition https://www.w3.org/TR/WebCryptoAPI/#aes-gcm-operations

classes/data/File.class.php
vanrein: > I dont think the web crypto api allows you to play around at the "16 byte low level chunk" size. Which we may perhaps refer to as AES-chunks. I would expect trying to encrypt a zero byte array of data on the web crypto api to produce undefined behavior on one or more browsers.

If you give me a link, I'll be happy to look for you.  What I also suspect is that you'll be allowed to supply arbitrary sequences of bytes, not just 16-byte multiples.  It would be quite an affront if you cannot pass 0 bytes through all (but then again, this is not a properly designed API for crypto).  If you cannot encrypt 0 bytes on all platforms we should look for another way, with security and all, to signal that the preceding 4MB-chunk was the last.

> I have tested the code on 4mb and 4mb+1 byte sized files. The former results in a single filesender chunk and the later results in two filesender chunks. Note that the actual byte of the ciphered content is different as per these calculations.

Hm, this is not what the report specified; both should produce 2 FS-chunks, the second of which encodes, a payload of 0 bytes, but also a signature to prove the end of transmission.

> What is your cryptographers opinion of adding a known pad to an exactly 4mb final filesender chunk of data? I could append a known cleartext to force the web crypto api to add a new chunk.

My opinion is that everything can work *provided* that you are able to parse the encrypted+signed content into an unambiguous conclusion that the case is different.

  * If you add a trailer, you can never be sure that it is indeed the last bit... the same bytes might occur inside, including at the end of a 4MB-chunk.  So your proposal isn't quite there yet.

  * What *should* work is sending 0 bytes, as specified.  I would be rather shocked if the WebCrypto API is too lame to do that well.  It gives the most consistent code, so I would prefer it.  Having said that:

  * What could work is not sending 4MB but 4MB+1 to indicate that it is the last part, and exactly 4MB in size.  All other payload sizes of end blocks would be <4MB.  As a result, payload sizes != 4MB indicate the end.  The extra byte could be ignored (or might be checked to have a certain value).  In all cases, the signature must be valid (which AES-GCM assures and results in success/failure from the decoding operation).

  * What could work, and would be really simple, is to use AEAD information to share the length.  This length is probably available out-of-bound, that is in headers or another non-secure form.  The AEAD adds security to such non-secure data.  Now you do need to check the length provided against that bit of data, but that's plain and simple and probably already part of what you are doing.

Feed problems with these options if you need me to generate more ;-)

vanrein:
> I have #644 to update the IV handling.

Wonderful.  If AEAD is used however, there are options with and without this.  Please show me how this ends up in calls to the WebCrypto API, I was assuming that you continued the key from one 4MB-chunk into the next and have not checked that yet.  Sorry -- I am acting on partial understanding of the code, especially the trace to WebCrypto API is not easy to understand due to variants and functional constructs.

classes/data/File.class.php
vanrein: > Each chunk has 32 bytes of overhead added by the encryption.

Yes, that would be 16 bytes of unencrypted IV and 16 bytes GMAC/signature.

> I assume in the initial comment above in the part "The last chunk is sent, and considered the closing chunk if it is partial." you are talking about the chunks used by the AES algorithm and not the 4mb filesender chunks.

I am talking of the payload that is 4MB for any but the closing chunk.  I am looking for an invariant, namely != 4MB, and that can be enforced by adding a possible 0 byte chunk.

```
4MB-chunk --> AES --> IV + E(4MB + GMAC) --> 4MB + 32
4MB-chunk --> AES --> IV + E(4MB + GMAC) --> 4MB + 32
4MB-chunk --> AES --> IV + E(4MB + GMAC) --> 4MB + 32
3MB-chunk --> AES --> IV + E(3MB + GMAC) --> 3MB + 32
```

Alternatively, when sending 8MB, it would be (as proposed, and assuming the other options given above are not needed)

```
4MB-chunk --> AES --> IV + E(4MB + GMAC) --> 4MB + 32
4MB-chunk --> AES --> IV + E(4MB + GMAC) --> 4MB + 32
0MB-chunk --> AES --> IV + E(0MB + GMAC) --> 0MB + 32
```

vanrein:
> Though you may also have had other downgrade attacks in mind as well?

I am not very specific, just mentioning that compatibility and security are opposing forces here.  It is trivial to overtake a client role and claim to only support an older version, and if this is below the acceptable level of the server it should be turned down.  The opposite is less likely; the server cannot be easily forged when TLS is used, so its behaviour would be more reliable.  If this sounds like things you already solved, than pardon me for double-checking it ;-)

vanrein:
> The details of what the WebCrypto API is doing at the lower level are detailed in the NIST document cited in step 6 of the definition https://www.w3.org/TR/WebCryptoAPI/#aes-gcm-operations

Thanks, the `encrypt()` and `decrypt()` are what you use then.  I saw distinctions between subtle and supposedly other, and didn't know where to look.

I see a `additionalData` parameter -- that's what I meant with AEAD.  Good, so we can format a string holding things from the context to pin down with the encryption, like

  * Unique (e.g. proper random >= 128 bits) code for the upload
  * Total size in bytes (no more padding or 0-byte-block need)
  * Sequence number 0,1,2,... of 4MB-chunk (your `chunkid` in fact)
  * Metadata, only in the first 4MB-chunk:

      - Sender email address
      - Filename?
      - Description?
      - ...

The former three are really, really helpful because they can simplify several things we've talked about.

In fact, when we use this, I am all in favour of your original modification with a different IV per 4MB-chunk, treating them like separate files.  The unique code allows us to ensure that it is part of the same upload.  The total size means that we can know what the last chunk is without looking at its payload size  to be != 4MB.  The sequence number is different for every 4MB-chunk, and is useful in confirming the block to be in the expected position.

The metadata can be assured too.  It still needs to be sent along, because the signature only hashes it, but the first chunk will now validate the metadata so that it cannot be modified without being noticed.  You can place anything there that you like, as long as it is in a format that could be parsed back into the original fields; that way, you are sure not to loose information prior to the actual hash.

PR #644, opened 2019-09-19
GCM updates: IV stored per file, more checks

monkeyiq:
A random IV is now created client side and stored in the database for
each uploaded file (the fileiv). That IV is 12 octets long for GCM
mode and 16 octets (unused) for CBC.

The fileiv is now used to set the first 12 octets of the IV for a file
during upload for GCM files.

For GCM: The fileiv is verified against the IV for each filesender chunk during
decode to ensure that it is the same for the IV for each filesender
chunk for the file.

The chunkid is extracted from the IV for each filesender chunk to
verify that it has the correct value (it is for this filesender chunk).

When the FileSender chunkid is encoded and decoded to an encryption IV
it is now expaned to cater for the FileSender upload_chunk_size and
contracted back during a decode. The exact value setting of the
chunkid needs to be more relative to the AES chunk size when used in
an IV than the filesender chunk number.

An explicit check that a file does not exceed the maximum size for GCM
is done in many places both on the client and server side. This uses the
internal constant crypto_gcm_max_file_size set in ConfigDefaults.php.

A client_entropy pool of 32 octets is now also uploaded from the
client to the server at transfer creation time in perparation for
future use.

### DIFF
--- a/classes/data/File.class.php
+++ b/classes/data/File.class.php
@@ -94,6 +94,14 @@ class File extends DBObject
             'size' => 60,
             'null' => true,
             'default' => 'StorageFilesystem'
+        ),
+        // The IV used to encrypt the file.
+        // This is 24 bytes long to allow for storage
+        // of a 128bit array in base64 format
+        'iv' => array(
+            'type' => 'string',
+            'size' => 24,
+            'null' => true
         )
     );
 
@@ -146,6 +154,7 @@ class File extends DBObject
     protected $upload_start = 0;
     protected $upload_end = 0;
     protected $sha1 = null;
+    protected $iv = '';
    
     /**
      * Related objects cache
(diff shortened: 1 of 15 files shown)


monkeyiq:
The chunkid expansion above is to handle the case mentioned https://github.com/filesender/filesender/pull/640#discussion_r325435530

The code to do that is at
https://github.com/filesender/filesender/pull/644/files#diff-44c07c50854004295471d34bff92cd45R87
and the related function below that to decode. Basically in most of the code a chunkid is almost always a filesender chunkid (0,1,2,3) until it is encoded into an IV format for use in crypto in which case it becomes(0, 262144, 524288, ...). This avoids doubt about what format a chunkid is in throughout the code.

PR #655, opened 2019-10-10
update script. This needs testing

monkeyiq:
It seems a few method names had changed 4 years ago and this part of the code didn't change too.


monkeyiq:
This will have to be remade against development branch

PR #677, opened 2019-11-06
Write days valid from config into Python API client

peter-:


monkeyiq:
Thanks for this, I made a slight update and merged it as https://github.com/filesender/filesender/pull/682

Issue #675, opened 2019-11-06
i18n: Spanish language files (es_ES) contain Estonian (et_EE) definitions


Phaze-III:
Not all of them but a lot.

PR #678, opened 2019-11-08
i18n: Put Estonian in et_EE directory (#675)

Phaze-III:
Puts Estonian language files in the et_EE directory when bulk importing from POeditor


Phaze-III:
Retracting this one, same commit will be in another PR.

PR #683, opened 2019-11-13
guests: better handling of do not expire

monkeyiq:
This brings out the do not expire option to near the expire date selection. The expires column in the Guests table can now be null to indicate that the guest does not expire. This is explicitly set if does not expire is selected during guest creation. 

Some of the SQL fragments had to be updated to return results where the guest does not expire. Without doing this then some guests might not have been returned, for example ones that had a expires time before now but are set to never expire. IMHO it is better to use null for guests that do not expire to explicitly capture the intent that the guest has no expire time.

Tested on Firefox, will test on other browsers soon.

### DIFF
--- a/classes/data/Guest.class.php
+++ b/classes/data/Guest.class.php
@@ -582,10 +582,6 @@ class Guest extends DBObject
         }
         return false;
     }
-    public function does_not_expire()
-    {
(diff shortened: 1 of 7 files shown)


monkeyiq:
It turns out there were a few other places that were checking if the Guests::expires was still in the future even if the guest should never expire. Still checking for more.

PR #685, opened 2019-11-15
i18n: revert to order and definitions as in release 2.11 plus one new item.

Phaze-III:
The poeditor import did a re-order and overwrote a missed translation
change in #669. This commit restores the order so that recent
definitions are again at the end (including the last added
user_hit_guest_limit). The changed storage_filesystem_file_not_found
definition is now also changed in POeditor.


monkeyiq:
I have the new term being brought into github with https://github.com/filesender/filesender/pull/686. I have been a bit lax with new terms since there hasn't been much contention in adding them, it is mostly me. I normally add new terms to the end of the local en_AU/lang.php and them use the scripts to move them across to poeditor. It would be nice to move to something where there was a closer link between poeditor and github so a PR could bring in terms to poeditor in a transactional way.

As I put up on the https://docs.filesender.org/v2.0/i18n/ page new terms can be exported to a local file to be imported into poeditor with the scripts/language/export-terms-to-new-and-deleted-lists.sh script. The admin rights needed to import new terms en masse on poeditor is something I have not looked at because the account on poeditor I use I think has project wide permissions. If there are only a few new terms found (as is the case in https://github.com/filesender/filesender/pull/686) then they can be added to poeditor manually and the oldterms.txt updated in the same PR to indicate that those terms are added.

When import-langs-directory.sh is run bringing in en_AU it will reorder the language/en_AU/lang.php file alphabetically again. I had hoped to use the "php" export mode via the API in import-langs-directory.sh but that isn't possible or planned on the poeditor side at the moment. So the script  would have to download a different format export and convert it to php or be rewritten to handle a different export format directly (import-translation-for-language.php). I plan to do one of those two things (or something better if there is) because it would make the bulk imports from poeditor much simpler. I would then add the option to the script to git branch first, download, import, commit, and push to github to make the whole process fairly painless.

PR #696, opened 2019-11-23
db: update auditlogs query for pg and mariadb compat

monkeyiq:
It seems removing the table alias makes both MariaDB 10.3 and PostgreSQL 10.6 happy. This relates to the later part of https://github.com/filesender/filesender/issues/630.

Testing has been done only on a single 'left over guest' by creating a reference to a Guest that is not existing in the Guests table. Assuming 404404 is not a valid Guests.id.
```
insert into AuditLogs values (default, 'event', 'guest', 404404, 'user', 'userid','1.1.1.1',now());
```

Then the aliasless SQL seems to run and remove that one row on both PG and Maria
```
delete from AuditLogs where id in
(   select id from (    select al.id from auditlogsview al    left outer join Guests g   on g.id = al.target_id_as_number
    where al.target_type = 'Guest' and g.id is null ) b )
```


monkeyiq:
This will have to be remade against development branch

PR #705, opened 2019-12-02
fixes for ie11

monkeyiq:
This is only for keyversion 0


madsi1m:
Is this all to make filesender work with ie11???

monkeyiq:
It is just for ie11. Using keyversion 0 I get a clean upload / download in ie11 on windows10 with this PR in place. I have something in the pipe to work with keyversion 1 but nothing at the moment for ie11 and other keyversions.

Issue #679, opened 2019-11-08
API access broken when aup_enabled


peter-:
## Problem statement

When trying to access the API using [filesender.py](../tree/master/scripts/client/filesender.py) on a system that uses AUPs...
```php
$config['aup_enabled'] = true;

$config['auth_remote_user_enabled'] = true;
$config['auth_remote_user_autogenerate_secret'] = true;
```
this always fails due to [missing `aup_checked`](https://github.com/filesender/filesender/blob/6d345b457d0c4c3f9cecf3e2f7a2793e38845a42/classes/rest/endpoints/RestEndpointTransfer.class.php#L370-L374) with this message on the client's command line:

```
Exception: Http error 400 {"message":"rest_bad_parameter","uid":"5dc437641651b","details":{"method":"post","endpoint":"transfer"}
```
and the server logging this (giving the only hint as to the reason):
```
[RestBadParameterException details uid:5dc437641651b] parameter = "aup_checked"
```

Since we are running our FileSender instance with `aup_enabled=true` this currently prevents us from offerning API access to our communities.

There are several aspects to this alluded to below and multiple issues may be spun off from this one to track those, if desired.

## Approaches

While it should be easily possible to amend `filesender.py` (and the other client implementations released by the project) to supply the necessary API message signalling that AUPs have been accepted [recent discussions on filesender-dev](https://sympa.uninett.no/lists/filesender.org/arc/filesender-dev/2019-11/msg00007.html) (the whole thread may be worth reading) suggest that maybe the server should not even trust proper API messages that claim that the AUPs where in fact accepted (due to the theoretical possibility of  the existence of "evil clients" unconditionally claiming that, w/o any user interaction) and instead should make issuing of API keys themselfs conditional to the subject prior having accepted the AUPs in the web UI.

If OTOH hypothetical "evil API clients" are not a major concern for the project at this time admending any clients distributed by the project with a flag to signal acceptance of "the current AUPs" via the API is the simplest and most direct fix to this issue.

[...]

Issue #579, opened 2019-06-14
Manage API keys in web interface


peter-:
Filesender now provides a link to the Python API client in the user profile but does not provide any means to create or reset (i.e., revoke and create in one step) the necessary API keys required for that. (Update: If `auth_remote_user_autogenerate_secret=true` is set in the config it will create and show such API keys. AFAICT that's the only method FileSender provides atm.)

As [suggested earlier](https://sympa.uninett.no/lists/filesender.org/arcsearch_id/filesender-dev/2018-03/20180309160039.3yhadbfovt6cwsor%40aco.net) modelling this after Gitlab's "private access tokens" (i.e., not OAuth) would be sufficient, IMO: A bit of UI to "create/reset API key" (in the user profile, right next to the Python client) and code to persist that key (salted and hashed, like a password) to the database, not to be displayed again after creation. (You'd reset/regenerate it, instead, had you forgotten/lost the API key.)

In the initial implementaton a single API could be managed that way (K.I.S.S.). A more flexible implementation might allow to provide "labels" for those keys when creating them and manage multiple API keys, so the subject knows what API key was used where (if the label them accordingly). That avoids having to share the same API across multiple machines and then possibly having to re-key all clients in case *one* of those machines might not be considered sufficiently secure. (But whether people actually make use of the CLI client and from multiple machines is an open question to me, so the simple case should probably be looked at first.)

monkeyiq:
I just merged https://github.com/filesender/filesender/pull/718 which allows the api secret to be deleted and created. The time of creation is now also stored and displayed. There is also support for an aup for the secret if a site wishes to enable that. Though I will mention the aup part elsewhere.

I am not sure what extra security you would get by not displaying the api secret after the initial creation. As soon as a user discovered that the site was doing that they would likely copy and paste it into a text file to mitigate the site hiding it.

There is still only a single api secret per user. Perhaps if we are planning to move to many secrets it might be a good time to first move to oauth or something like that. Another interesting idea might be to allow server side checks on api secret use. For example, a user might know that it is only used from ip address xyz and if requests come from elsewhere might like to know about it or have the key invalidated. 

In the recent updates the api secret is also downloadable in an ini file which includes other information to such as base_url. There are a number of security issues with passing the api secret on the command line as was the only option before.

WebSpider:
@peter- has this been implemented sufficiently?

peter-:
[...]

PR #718, opened 2020-01-21
filesender.py ability to delete and create apisecret, uap

monkeyiq:
The time that the api secret was generated is now tracked in the database. The secret can be deleted and created through the "my profile" page on the site. When creating an api secret, if the config api_secret_aup_enabled=true then a dialog is displayed with a link to the full apisecretaup page that allows the aup to be displayed in full in a single page. A secret is not generated unless the user confirms acceptance of this dialog.

The time of acceptance of the api secret aup is also displayed on this page.

This relates to issue https://github.com/filesender/filesender/issues/679


monkeyiq:
This more relates to issue https://github.com/filesender/filesender/issues/579

Issue #612, opened 2019-08-13
Feature: My Transfers page: details view: show subject (and message)


belnet-ming:
This is a feature request from some of our users:

They want (when an email is sent) the subject, at least, to be displayed in the detailed view of a transfer. If possible, they also want the message.

monkeyiq:
The above PR should improve the formatting of that information and also add in the subject and message.

PR #728, opened 2020-01-25
my transfers as table with extra email md

monkeyiq:
The rows in the my transfers page have been converted to a table for cleaner formatting. If the upload was to email recipients then the subject and message is now also shown in the table.


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/612

Issue #627, opened 2019-09-03
Edge console.log doesnt work


madsi1m:
hard to debug in Edge when console.log doesnt work.

https://www.beyondjava.net/console-log-surprises-with-internet-explorer-11-and-edge

monkeyiq:
Perhaps using a window.filesender.log() which forwards to console.log() is a reasonable solution. It would allow devs to run in a mode that made the collected logs available by a button on the web interface and still allow the "send client logs" feature to work for users of the system on Edge.

monkeyiq:
This may possibly be also causing issues in client logs not showing up (see https://github.com/filesender/filesender/issues/700).

monkeyiq:
It is a good idea to follow up on this with some checking for use of console.log creeping back into the codebase. Such a thing is very likely and would mean those log lines would not be captured on Edge.

monkeyiq:
As the core issue had a fix merged back in may I am closing this issue. Updates to detect console.log calls again should be a future update.

PR #733, opened 2020-01-28
clientlogs: move from console.log to clientlogs.log

monkeyiq:
It was noted in issue https://github.com/filesender/filesender/issues/627 that some browsers ignore console.log() when the developer tools is not open. This PR introduces a new windows.filesender.log() method that adds log lines to the clientlogs even when the browser developer tools window has not been opened. With this PR, code should call window.filesender.log() instead of console.log(). If there is no logging setup (eg, a worker) then window.filesender.log() should still be available and default to falling back to using console.log().

The clientlogs are shown in reverse order (newest first) in the admin/users/show client logs window.

A new export client logs button on the my profile page allowing a user to generate a text file containing a small amount of metadata (that this is a filesender client logs file, and when it was written) and the client logs entries. This offers a fallback to a user who wishes to see the clientlogs and or send them in another way.


monkeyiq:
CI link https://travis-ci.org/github/filesender/filesender/builds/673306243

monkeyiq:
This is closed in favour of https://github.com/filesender/filesender/pull/793

PR #751, opened 2020-02-20
Improve client IP lookup

yent:
Adds support for multi-valuated client_ip_key config parameter (expressed from most significant source to least significant one).

When previously using X_FORWARDED_FOR with chained proxies the returned client IP was wonky since the header may be multi-valuated and needs parsing in that case.

IPs are validated using php's internals, rejecting local/private network IPs and reserved ones.

Configuration example :
  - $config['client_ip_key'] = 'REMOTE_ADDR'; // filesender is its own front server
  - $config['client_ip_key'] = 'X_FORWARDED_FOR'; // there is a proxy in front
  - $config['client_ip_key'] = array('X_REAL_IP', 'X_FORWARDED_FOR'); // there is a proxy in front that uses real-ip, then x-forwarded-for

### DIFF
--- a/classes/utils/Utilities.class.php
+++ b/classes/utils/Utilities.class.php
@@ -316,7 +316,25 @@ class Utilities
      */
     public static function getClientIP()
     {
-        return isset($_SERVER[Config::get('client_ip_key')]) ? $_SERVER[Config::get('client_ip_key')] : '';
+        $ips = array();
+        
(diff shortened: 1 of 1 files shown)


monkeyiq:
I think we would also want to change the config validator to string|array
https://github.com/filesender/filesender/blob/development/includes/ConfigValidation.php#L40

PR #762, opened 2020-03-29
auth: web interface for local password management

monkeyiq:
The https://github.com/filesender/filesender/pull/761 introduced the ability to setup SimpleSAMLphp to authenticate users with password hashes stored in the filesender database. 

This PR offers a web interface to allow creating new users with emails to the new users containing the initial passwords, for admins to explicitly reset user passwords, and for users to modify their own passwords.

These elements are all off by default. To enable them set using_local_saml_dbauth and see scripts/simplesamlphp/passwordverify for details on how to setup SimpleSAMLphp for this scenario.


monkeyiq:
The intention of this and the supporting PR is to have no impact on sites using large scale saml authentication. Thus the default is to not enable these features.

For small sites of 5-30 users the person setting up the service can enable this feature and deploy filesender more easily.

PR #764, opened 2020-04-05
computest-414: increase lockdown that transaction creator is the file uploader

monkeyiq:
To add yet more security to the existing checks this PR creates a 16 byte random string when a transaction is created and returns that to the client. When the client is uploading file chunks (encrypted or not) it has to return that same roundtriptoken or the upload is rejected. This effectively binds the person creating a transfer to be the only person who can upload data to the files associated with that transfer.

Tested on Firefox, Chrome, {Edge, and IE11 on Windows 10}. In case this does something undesired in an environment is can be disabled with the chunk_upload_roundtriptoken_check_enabled config option (off by default).

This feature is off by default so that information stored in the browser that does not contain the roundtriptoken can still be used to upload a file. Also if a transfer is already in progress it will not be rejected by this test. 

Once the code in this PR is being used the roundtriptoken values are still created and sent to the client and back. At some stage, a system admin can say that transfers that were started in the browser that were not completed (say, 1 month after moving to the release) will be considered lost and turn on chunk_upload_roundtriptoken_check_enabled. This is a trade off between enabling the feature right away and annoying some users or enabling it a little later when almost all users will have the roundtriptokens in the browser for a started but abandoned transfer.

Note that the above "not completed transfer" is an abandoned transfer such as if the user navigates away from the upload page and on returning is asked if they want to setup the transfer again. Play, pause, and reconnect and resume will work right away as they will use the roundtriptoken sent when the transfer was created.

The REST filesender.py is updated. Functions such as restartFailedTransfer() are updated.

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -145,7 +145,16 @@ class Transfer extends DBObject
             'size'    => '44',
             'null'    => true,
         ),
-        
+
+        //
+        // This is to hand to a client that is making the transfer and to upload
+        // they have to hand it back to ensure they are the one who made the transfer
+        //
+        'roundtriptoken' => array(
+            'type'    => 'string',
+            'size'    => '44',
+            'null'    => true,
+        ),
     );
 
     /**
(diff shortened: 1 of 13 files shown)


monkeyiq:
Added 'chunk_upload_roundtriptoken_check_accept_before' to allow migration to this option.

monkeyiq:
https://travis-ci.org/github/filesender/filesender?utm_medium=notification&utm_source=github_status

PR #765, opened 2020-04-07
only get frequent email addresses over http post

monkeyiq:
Use HTTP post to get the list of frequent email addresses to avoid a GET request that has a chance that it might introduce some data into logs that is not necessarily desirable.


monkeyiq:
https://travis-ci.org/github/filesender/filesender/builds/672313498

Issue #754, opened 2020-03-02
Feature: More "user friendly" error message


belnet-ming:
A lot of users are asking what to do when they've got an error message like this:
![image](https://user-images.githubusercontent.com/49909162/75664244-73f9cf80-5c72-11ea-8e25-83132b893055.png)

Most of the time, the transfer has expired (or has been deleted by the uploader), but it is not explicitly said in the error message.

So, we should maybe change the error page to a generic "The transfer has expired and files have been deleted from our server." message instead of a "Recipient not found".

madsi1m:
Change the word “server” to the product name eg “CloudStor FileSender” I think we have the name in the config somewhere

monkeyiq:
This has been included in a release. Thanks for reporting the issue!

PR #769, opened 2020-04-10
UX: better error message when you attempt to download expired or deleted transfer

monkeyiq:
This relates to https://github.com/filesender/filesender/issues/754

It was certainly confusing to see the old error message. I have used a new "presumed expired" exception type because we don't really know what happened to the data, if it was once valid it has been removed from the database but it might also be that it never was in there in the first place. We presume it expired because we presume the link was once valid.


monkeyiq:
CI link https://travis-ci.org/github/filesender/filesender/builds/673278959

Issue #531, opened 2019-04-18
Autocomplete of email address should not be case sensitive


baumrasen:
(with autocomplete function for email addresses in To: field enabled)

You have to type in upper case, if the address is stored with upper case to get the email address autocompleted.
Because of email adresses are working with upper and lower case the system should ignore upper and lower case.

monkeyiq:
Thank you for reporting this. Sorry about the delay. It should be addressed with the merge of https://github.com/filesender/filesender/pull/770

PR #770, opened 2020-04-10
email addresses are case insensitive.

monkeyiq:
The user should not have to remember the case used to find a frequent email address. This addresses the issue first reported here https://github.com/filesender/filesender/issues/531


monkeyiq:
CI is at https://travis-ci.org/github/filesender/filesender/builds/673287576

Issue #582, opened 2019-06-26
show number of users in global stats in admin page


madsi1m:
show number of users in global stats in admin page

monkeyiq:
I have put in the estimated number of users from the database to try to not hit the database too much. It seems like the number of active users per week or day would be a nice statistic, but maybe better presented on a page of it's own.

monkeyiq:
... also my above suggestion is more info from the aggregatestats part of the system.

monkeyiq:
Maybe tracking other stats we would like up is best done in another discussion / gh issue.

PR #771, opened 2020-04-10
stats: show user count estimate on statistics page

monkeyiq:
In order to avoid doing a 
```
SELECT count(*) from TABLE
```
a new countEstimate method was introduced in DBObject.class.php. The result might not be 100% accurate but should be fast to compute.

The first example of this new method relates to https://github.com/filesender/filesender/issues/582


monkeyiq:
CI is at https://travis-ci.org/github/filesender/filesender/builds/673293145

Issue #757, opened 2020-03-17
Decrypt download: Show password only when explicity indicated


WebSpider:
Currently, when a file is encrypted upon upload, while downloading the fileset, a password is requested for decryption. 

When using Firefox and Chrome, this password is displayed on-screen. 
This request is to make this behavior more aligned with current web practises: 
- Hide for the enduser
- Button to show the password
- Possibly only show the last entered character for a second

monkeyiq:
It may be an interesting option for the user to be able to set "show passwords by default on this computer" in the my profile page. For example if the PC is located in the user's house and they are happy that the screen is private.

monkeyiq:
Hide password has been released in 2.16.

WebSpider:
Confirmed working, thanks

PR #772, opened 2020-04-11
hide download password text visibility by default, option in dialog to show it

monkeyiq:
As mentioned in https://github.com/filesender/filesender/issues/757 there is a difference between the encrypt and decrypt password input. This PR makes the decrypt password hidden by default with a checkbox option to show/hide it, similar to the upload page.


monkeyiq:
CI is at https://travis-ci.org/github/filesender/filesender/builds/673660045

PR #777, opened 2020-04-25
Move to a new havePermission method for testing access

monkeyiq:
This moves code from being replicated to test access into a new Transfer::havePermission() method to make the code checking for permission simpler. A new exception class was also added to avoid having to pass the user->id around to slipstream that code path.


monkeyiq:
CI is at https://travis-ci.org/github/filesender/filesender/builds/679281962

PR #778, opened 2020-04-25
computest-413: test against attempts to modify settings

monkeyiq:
Prevent guests from trying to change some settings


monkeyiq:
CI is at https://travis-ci.org/github/filesender/filesender/builds/679287398

PR #786, opened 2020-05-05
do not go sifting a non array if there was an exception

monkeyiq:
If there was an exception during processing of the filter operation then do not try to lookup values there as that is not going to be an array and will cause issues for those who have
```
display_errors = On
```
in /etc/php.ini.


monkeyiq:
https://travis-ci.org/github/filesender/filesender/builds/683584426

PR #787, opened 2020-05-07
rate limit guest actions and admin to see who has hit limits

monkeyiq:
This PR uses the auditlog to keep track of how many times the user reminds a guest or tries to create a guest. Note that while there is already a limit on the number of guests a user can create, a user might attempt to be nasty by creating a bunch of guests and then deleting them all so they can remake more. This PR introduces a new 'rate' of creation limit allowing a cap on per day guest creation activity with configuration settings.

To compliment this there is a new section in the admin / users page allowing a site administrator to see who hit guest creation total limit, rate limit, who has deleted the most guests that did not send even a single file, and who has deleted the most guests in all. While a typo would mean deleting a guest that didn't send a file, it is unlikely that a user will delete a whole bunch of unused guests that didn't do anything themselves as that sort of goes against the reason to make the guest in the first place.

From a coding perspective the new Logger::logActivityRateLimited() method is a single point of call to throw an exception if the rate is too high when taken against a configuration setting. This new method acts like Logger::logActivity() but will also log a "rate limit hit" if called too frequently and throw an exception. This effectively rolls a whole control block to verify rate limiting into a single function call.

DBObject now has a count() method to compliment the all() method if you only want a count of results from a query.


monkeyiq:
This is the new addition to the admin / user query section...

admin-users-query

Issue #745, opened 2020-02-14
Admin transfers should have some way to adjust the sorting


monkeyiq:
Perhaps by transfer id, size, Recipients, expires.

monkeyiq:
This is a duplicate of https://github.com/filesender/filesender/issues/148

monkeyiq:
It was also requested today on the mailing list

monkeyiq:
It might also be a good idea to allow search like on the users tab. If you know a transfer ID already then paging around to find it isn’t going to make things any more fun.

monkeyiq:
The size field is calculated as the sum of the size of the files for a transfer.

monkeyiq:
The 'downloads' is calculated from the auditlogs.

monkeyiq:
This has been merged with https://github.com/filesender/filesender/pull/794.

monkeyiq:
My Transfers page can now also be sorted with https://github.com/filesender/filesender/pull/796

PR #794, opened 2020-05-14
admin transfers page allow sorting

monkeyiq:
This relates to https://github.com/filesender/filesender/issues/745 so the admin section the transfers listings can be sorted by clicking on the table headers.

A follow up PR might also use this base code for the my transfers page to allow the user to sort their transfers view too.


monkeyiq:
I notice going through the pages forgets the sort param. have to fix before merge

PR #823, opened 2020-08-01
Release 219binaries

monkeyiq:
some image files needed explicit updating.


monkeyiq:
against master instead

PR #831, opened 2020-09-01
Change PASSWORD to PASS in documentation to match script

bepstein:
I noticed in the documentation for v2.0 installation that the PASSWORD variable set does not match the PASS variable that is actually used.  Here is the fix for the documentation.

### DIFF
--- a/docs/v2.0/install/index.md
+++ b/docs/v2.0/install/index.md
@@ -296,7 +296,7 @@ config file itself.
 * **NOTE**: Replace the PASSWORD line with a choice of your own!
 
 ```
-PASSWORD=something-really-long-and-wonderful-written-near-the-keyboard
+PASS=something-really-long-and-wonderful-written-near-the-keyboard
 
 cd config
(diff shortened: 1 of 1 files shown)


monkeyiq:
Thank you for this PR! 

There were are few rather minor changes and I created https://github.com/filesender/filesender/pull/834 with them which I will merge shortly. I thought it better to do that than just leave the feedback.

The contributor docs should mention that all PRs are now made against development. It is likely a good idea for me to find and update (or create) some overview there to help. At release time those changes are applied in bulk to the master branch and a tag made there. I also thought I'd update the PASS to PASSWORD instead of the other way around. The NOTE in the line above mentions "the PASSWORD line" so I thought it was best for all references to just use that variable.

bepstein:
Sounds good thanks!
ep

-- 
Brian Epstein <bepstein@ias.edu>                     +1 609-734-8179
Manager, Network and Security, CISO     Institute for Advanced Study
Key fingerprint = A6F3 9F5A 26C5 5847 79ED  C34C C0E5 244A 55CA 2B78

----- Original Message -----

monkeyiq:
I have now merged https://github.com/filesender/filesender/pull/834 which has the core of this change. Thanks again for the update to the documentation.

PR #838, opened 2020-09-20
test selenium

monkeyiq:


monkeyiq:
It seems Selenium has some issues with CI at the moment

Issue #841, opened 2020-09-24
Translation for Serbian language available


imilos:
The colleague of mine, Ivan Vojvodic, created a translation for the Serbian language using POEditor. It's available on https://poeditor.com/join/project/RqXr9WBJwU. Please consider accepting it into master.

monkeyiq:
Excellent work! 

I took a look at the ISO references I will soon add this to the import script to produce the sr_RS locale in FileSender. And also add it to the locale.php to be found.

ISO_639-1 = sr
ISO 3166-1 alpha-2 = RS

Info in https://github.com/filesender/filesender/tree/development/language

imilos:
Thanks a lot @monkeyiq ! the number of users who will benefit from the translation is not negligible :) 

Perhaps you have a suggestion for [this issue](https://github.com/filesender/filesender/issues/840). It's a show stopper for my institution's deployment but seems to be something easily manageable. Thx in advance!

monkeyiq:
With the above PR merged the translations are updated. The import script is also updated in that PR to include Serbian as a language that is imported during bulk translation imports, so it should also continue to be updated in the future.

monkeyiq:
Let me know if there are other updates needed. Thanks again!

PR #849, opened 2020-10-07
2020/oct/i18n serbian

monkeyiq:
This adds Serbian to the import all from poeditor scripts and also imports the latest translation for that language. By extending and using the scripts to do this the translations for this language will be updated when all imports from poeditor are performed in the future.


monkeyiq:
This was raised in issue https://github.com/filesender/filesender/issues/841

PR #850, opened 2020-10-07
i18n add new terms to memory

monkeyiq:


monkeyiq:
These (well this as there is 1) are up on poeditor now too.

Issue #844, opened 2020-10-02
filesender.py breaks on dependency if not installed


flickerfly:
The dependency on `requests` does not appear to be documented and when not installed keeps the script `filesender.py` from running. Requirements for installing should probably be documented somewhere or the script should catch the missing dependency and instruct the user to run `pip3 install requests`.

monkeyiq:
Thanks for the report. I have added some detection in the script and also updated the docs to point to pip or distribution packages where it makes sense. Please reopen or add more issues if there are other things that could be improved.

PR #859, opened 2020-10-18
filesender.py: a more friendly message if deps are not installed

monkeyiq:
This was reported in issue https://github.com/filesender/filesender/issues/844. This PR tries to give more information about an import failure for two libraries that might need to be installed by pip or through platform packages.


monkeyiq:
The main CI has completed (sans selenium) and the other two commits are docs only.

Issue #840, opened 2020-09-22
Expiration email


imilos:
Is there a possibility to suppress sending emails such as

> Transfer n°32 has expired and is no longer available for download. 

to receivers on the expiration date? I couldn't find the option to do that.

monkeyiq:
When a file is uploaded there should be the option "Notify me when expired" which should allow the user to select if that are wanting that feature. The default for that should be able to be set in your config.php.

A while ago it was requested that the "Send me daily statistics" option might be able to be turned off after an upload (through the My Transfers page). Are you after the option to turn this off after an upload is complete or just be able to turn it off + disable it in general for all users?

imilos:
Dear @monkeyiq , thx for your quick reply! No, I'm not talking about the feature that enables sender to be notified upon expiration. That's clear and can be enabled or disabled in the UI. What I'm talking about is notifying the recepient about the expiration. I cannot see a way to turn it off.

WebSpider:
@imilos We have disabled this in our install, but I have forgotten how we did it :)
I'll look it up.

monkeyiq:
It is very useful to know if sites are disabling things like this. It makes it much easier to turn it into something that can be turned on/off with a config.php setting so that everybody can benefit from that update and no longer have to patch things out.

monkeyiq:
The PR https://github.com/filesender/filesender/pull/860 should add this feature if you set your config.php as described in the PR setting your $config['transfer_options'] / email_recipient_when_transfer_expires to false. I've also put some stuff in that PR as to how I have been testing things.

imilos:
Many thanks for your support! IMHO this is a very useful option! We will check your commit ASAP!

monkeyiq:
Please reopen if the issue persists.

PR #860, opened 2020-10-18
transfer expire: make email to recipient on expire optional

monkeyiq:
This is to address issue https://github.com/filesender/filesender/issues/840. I have used the transfer options to possibly allow this feature to be set per transfer in the future. At the moment setting the below fragment in your config.php should disable emails to recipients that a transfer has expired.

```
...
        'email_recipient_when_a_transfer_expires' => array(
            'available' => false,
            'advanced' => false,
            'default' => false
        ),
...
```


monkeyiq:
I have been testing the PR using the following to get the max ID and updating that upload to have expired.
```
select * from Transfers order by id desc limit 3;
update Transfers set expires = '2020-9-27 14:00:00' where id = 48;
```
then using the cron.php to expire the transfer and seeing what happens. Some Logger::error() calls in Trnasfer.class.php inside close() help to narrow things down. The really interesting part is if sendToRecipient() is called in the part that this patch touches.

PR #867, opened 2020-10-28
Always show page header when an exception is thrown

mdusher:
The page header template is not printed to the user if an exception is thrown in the Auth or Security functions before it. This results in a "plain text" error message being printed (see below).

image

This pull request stores the exception in a variable and re-throws it after the page header is printed.

I was able to replicate this by:
1. Creating a new voucher
2. Modify the expiry date in the database to be in the past
3. Try and load the voucher link.

You will get a `GuestExpiredException`  from `Auth::isAuthenticated()`.

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -2,13 +2,13 @@
 
 /*
  * FileSender www.filesender.org
- * 
+ *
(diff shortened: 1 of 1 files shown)


madsi1m:
Looks Good

monkeyiq:
Thank you for the detailed report, good reproduction information and patch :+1:

PR #877, opened 2020-11-25
Typo

yent:
Fix french translation


monkeyiq:
I am happy to merge the PR. Though when the next bulk import of translations from poeditor happens these updates will be clobbered with the data from poeditor. So changing there too will be useful. 

If there are a few changes it may be easier to update things in poeditor and I can just use the script to update the translations in github from poeditor.

PR #879, opened 2020-11-26
StorageFileSystemChunked: Chunk and file integrity verification

mdusher:
We've been encountering issues where chunks have not been written to disk correctly, so we have added in extra validation for chunks.

This pull request adds extra verification of each chunk file on upload by confirming:
1. number of bytes written matches the expected size of the chunk
2. the file actually exists on the storage after the write
3. the on storage file size matches the number of bytes written
4. an md5 hash of the on storage file matches an md5 hash of the data provided to write

I've also done a bit of refactoring to improve the readability and add in some phpdoc.

### DIFF
--- a/classes/storage/StorageFilesystemChunked.class.php
+++ b/classes/storage/StorageFilesystemChunked.class.php
@@ -10,12 +10,12 @@
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
  *
- * *    Redistributions of source code must retain the above copyright
+ *     Redistributions of source code must retain the above copyright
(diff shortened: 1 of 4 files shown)


mdusher:
Added in resetting a files upload state and retrying the entire file when a FileIntegrityCheckException is thrown.

Still having a weird situation where the file gets marked as uploaded (the row goes green and gets a tick) until the automatic retry kicks in.

monkeyiq:
I will be having a look at updating and merging this in soon.

mdusher:
I've actually moved away from this particular change and created a new storage class for the EOS storage. I'll close this one for now and look at creating a PR for that one instead.

PR #884, opened 2020-12-03
Don't set readonly on generated password

peter-:
This change adds/restores the ability to edit/change/replace the password after having it generated, without having to restart the transfer from scratch (or changing that attribute in your browser using developer tools).

As long as people can set trivial passwords before/without ever selecting "generate password" preventing them from changing the generated password doesn't achieve much security-wise, I think.

There may of course be other reasons for setting the input field `readonly` I'm not aware of right now.


peter-:
Sorry, branched from master (not development) and didn't notice the completely b0rked changes before submitting. Re-done in #885

Issue #880, opened 2020-11-28
disable or force encryption


helmut72:
Please allow disable or force encryption.

It's impossible to type an own password if 'generate password' was clicked accidentally. Also it's impossible to see encrypted filesize limits if differs from unencrypted. Both would be raise calls to an helpdesk.

Any option is one option too much for users. ;)

Thank you.

helmut72:
Oh, it looks like there is an option 'encryption_enabled'

But it's not documented in:
https://docs.filesender.org/v2.0/admin/configuration/

I will let this issue open, because there is no 'force_encryption'.

peter-:
>> Please allow disable or force encryption.
>
> Oh, it looks like there is an option 'encryption_enabled'

To rephrase: `includes/ConfigDefaults.php` comes with `encryption_enabled` out of the box (note that it's not actually _encryption_ that's _enabled_ that way: it's controlling whether the option for enabling encryption is shown in the UI!) and deployments can disable showing of the option for encryption by setting:
```
$config['encryption_enabled'] = true;
```
in their config.php. There's no way currently to have the checkbox preselected/enabled by _default_ nor to have it enabled but not shown in the UI (enabled but not _available_, in "transport options"-speak).

> It's impossible to type an own password if 'generate password' was clicked accidentally. Also it's impossible to see encrypted filesize limits if differs from unencrypted. Both would be raise calls to an helpdesk.

Those are two separate issues (one trivial, the other possibly impossible):

### Unchangeable generated passwords

Once you click "generate password" the HTML form input element for the password get's changed by adding _readonly_:
```html
<input class="encryption_password" id="encryption_password" ... readonly="readonly"
```
via
```js
filesender.ui.nodes.encryption.password.attr('readonly', true);
```
TBH I don't understand the reasoning for that. That way you can't simply go back (you decided you've changed your mind, e.g. given the length of the generated password once you see it) and instead type in your own. Disabling and re-enabling encryption again via the checkbox also doesn't re-set this (which is fine with me). Simply not setting `readonly` this all still fulfills its purose (it generates a password for people who can't be bothered to generate or come up with one themselfs, prefilled into the input form) but leaves the rest up to the subject, including changing her mind or changing the generated password.

### Showing file size limit for encrypted files

[...]

PR #885, opened 2020-12-03
Don't set password input field readonly after generating password

peter-:
This change adds/restores the ability to edit/change/replace the password after having it generated, without having to restart the transfer from scratch (or changing that attribute in your browser using developer tools).

As long as people can set trivial passwords before/without ever selecting "generate password" preventing them from changing the generated password doesn't achieve much security-wise, I think.

There may of course be other reasons for setting the input field readonly I'm not aware of right now.

Should take care of one of the several issues raised in #880

### DIFF
--- a/www/js/upload_page.js
+++ b/www/js/upload_page.js
@@ -1467,7 +1467,6 @@ $(function() {
         var crypto = window.filesender.crypto_app();
         var encoded = crypto.generateRandomPassword();
         password = encoded.value;
-        filesender.ui.nodes.encryption.password.attr('readonly', true);
         filesender.ui.nodes.encryption.password.val(password);
         filesender.ui.transfer.encryption_password_encoding = encoded.encoding;
         filesender.ui.transfer.encryption_password_version  = encoded.version;


monkeyiq:
Generated passwords and user selected passwords are treated slightly differently by FileSender. The basis for this is that the generated password is created using entropy and uses the whole scope of each byte. On the other hand most user selected passwords use only a range of characters that can be easily entered on a keyboard. This has follow ons for password hashing and possibly other treatments.

Back to this change, in order to allow the user to edit the password manually again FileSender would also need to monitor for that and probably set change the following to again treat the password as user entered:
```
filesender.ui.transfer.encryption_password_version = crypto.crypto_password_version_constants.v2018_text_password;
```
I have put an update in https://github.com/filesender/filesender/pull/886 which includes some other tweaks such as making the show/hide password checkbox visible again. I have not tested 886 to my normal level as yet.

peter-:
> Generated passwords and user selected passwords are treated slightly differently by FileSender. 

Fine with me. Then another UI option to switch back to subject-provided passwords -- that does not involve canceling the whole transfer and starting another from scratch -- would be needed. E.g. the link "generate password" could have its text changed to  "enter a password of your chosing" (or whatever) once it has been seleted, reversing the change to the input field.  
I.e., click once to get a password generated and filled, click again to go back to the normal input field.

I don't care much about the implementation, I just think there's needs to be a way to back out of the generated password, as requested by the OP in #880. ("It's impossible to type an own password if 'generate password' was clicked accidentally.")

monkeyiq:
895 which cites this PR above should allow going back to user nominated passwords again.

PR #886, opened 2020-12-05
ui: allow a generated password to be converted back to a user entered password

monkeyiq:
This should allow a generated password to be taken over and reset to a user generated password by just altering the password. 
This is an update with additional updates for the update in https://github.com/filesender/filesender/pull/885.

### DIFF
--- a/www/js/upload_page.js
+++ b/www/js/upload_page.js
@@ -1303,6 +1303,9 @@ $(function() {
     filesender.ui.nodes.encryption.password.on(
         'keyup',
         function() {
+            // plain text passwords have a specific version
+            // and encoding which may be used by key generation
+            // so we must reset that here if the user starts modifying the password.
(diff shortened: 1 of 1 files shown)


monkeyiq:
With just the generate button this is a bit fiddly. Using arrows or control-c to copy the generated password has to not register an attempt to change the password (or localized versions or osx copy etc). Otherwise copying the generated password will turn the password into a "user entered" password instead of remaining registered as generated and thus with the entropy that comes with that.

monkeyiq:
Rather than fighting the input box I think I will make generate password a checkbox that shows a regenerate button. If the user unchecks generate password then the code explicitly knows when they are wanting a user entered password again.

monkeyiq:
This is being deprecated and the replacement is at https://github.com/filesender/filesender/pull/895

PR #890, opened 2020-12-07
Add configurable email headers

### LINKED ISSUE #871: Add email headers to help with auto-responses
When FileSender's `$config['admin_email']` parameter is set to an address where incoming email is being processed by humans (personal/functional email address, internal mailing/distribution list, ticket system address, etc.) outgoing messages/notifications from FileSender (e.g. "upload completed" or whatever) may cause auto-responses from some recipient to end up at that address.

I.e., the admin/support team would recieve auto-responses such as "I'm out of office until ..." as replies to system notifications sent to FileSender users or external recipients. Personally I have no use for or interest in recieving "out of office" messages from random people that may have sent or received files from out FileSender instance as a consequence of other FileSender users' actions.  
As such I'd suggest to add the following headers to all generated emails:

```php
diff --git a/classes/utils/ApplicationMail.class.php b/classes/utils/ApplicationMail.class.php
index d061de13..d2d38b0f 100644
--- a/classes/utils/ApplicationMail.class.php
+++ b/classes/utils/ApplicationMail.class.php
@@ -317,6 +317,8 @@ class ApplicationMail extends Mail
             $this->addHeader('X-Filesender-Context', $context);
         }
         
+        $this->addHeader('Auto-Submitted', 'auto-generated');
+        $this->addHeader('X-Auto-Response-Suppress', 'All');
         // Send email
         return parent::send();
     }
```

The first one is standardised by RFC 5436. The second one seems to be a proprietary header only M$ mail systems use. Neither should prevent actual bounces from being returned by the email systems receiving FileSender's emails with this header set.

If you don't want to hard-code then these yet another config option (additional email headers) would be needed, I suppose.

Note that I don't know what happens when FileSender's `$config['admin_email']` parameter is set to an address processed by FileSender's own bounce handling (`scripts/task/emailfeedback.php`), i.e., when it's set to the value of `$config['email_return_path']` --- whether that script would process these auto-replies correctly (meaning it would ignore these except to `(re)move_after_processing` them) or whether it might show false positives in the UI from random status notifications.

peter-:
Fixes #871 by making this a deployer choice, not a code decision (meaning we don't have to agree on the content of the additional headers). Though I'm not sure how you'd approach that (e.g. I'd default to an empty `array()`, not `false`, but `include/ConfigDefaults.php` had no such examples, etc.) and whether the proposed code/documentation suffices.  
My own config looks like this, though that may need some more tuning, depending on the brokenness of some mail systems:

```php
$config['email_headers'] = [
  'Auto-Submitted' => 'auto-generated',
  'X-Auto-Response-Suppress' => 'All'
];
```

### DIFF
--- a/classes/utils/ApplicationMail.class.php
+++ b/classes/utils/ApplicationMail.class.php
@@ -312,9 +312,15 @@ class ApplicationMail extends Mail
         }
         
         // Set context in headers so that it is returned along with bounces
+        $add_headers = Config::get('email_headers');
         if ($context) {
             $this->msg_id = '<'.$context.'-'.uniqid().'@filesender>';
             $this->addHeader('X-Filesender-Context', $context);
+            if (is_array($add_headers)) {
+                foreach ($add_headers as $key => $value) {
+                    $this->addHeader($key, $value);
+                }
+            }
         }
         
         // Send email
--- a/docs/v2.0/admin/configuration/index.md
+++ b/docs/v2.0/admin/configuration/index.md
@@ -76,6 +76,7 @@ A note about colours;
 * [email_return_path](#email_return_path)
 * [email_use_html](#email_use_html)
 * [email_newline](#email_newline)
+* [email_headers](#email_headers)
 * [relay_unknown_feedbacks](#relay_unknown_feedbacks)
 
 ## General UI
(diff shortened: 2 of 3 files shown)


monkeyiq:
Since the default config of 'email_headers' => false is already there for the admin and will only really be changed if they wish to add headers I don't think it is a big deal if that is array() or false.

Issue #888, opened 2020-12-07
Improve admins' ability to find transfers


peter-:
Providing support for a FileSender instance is currently being made difficult because
* `Admin` -> `Transfers` cannot be searched by sender email (only by min/max transfer id numbers),  
even though once you find and expand ("Show details") a transfer you can clearly see the sender email address,
* under `Admin` -> `Users` -> `Search for user (based on user ID)` you can only search for the (federated, in our case) userid but not for the sender email address (which may not be the same thing, depending on deployment details and configuration).

E.g. we're getting support requests from email address X but I can't search for transfers from that sender in the transfers page, nor can I search for that email adress in `Search for user (based on user ID)` (when the IDP has sent a different userid as SAML attribute).

In our current setup (using the Shibboleth SP) we'd have to perform the following steps to find the transfers for a given person (known only by email at that time):
* Ask the person for an specific transfer id/number (or for their userid, which not only sounds phishy, the subject also has no idea what identifier the IDP and the SP agreed on using to identify the subject with, i.e., the subject may not even know their own userid as far as what FileSender persistend for them. Also FileSender does not display it anywhere within the UI -- not that asking the subject to read their own userid off a web site and tell it to the support sounds like a very elegant way to provide support).
* Then try to find the transfer id in the web server access logs (`fgrep /transfer/12345 /path/to/access.log`) where we'll find the userid of the subject that's being passed to FileSender from the web server (because we've configured the Shib SP to set `REMOTE_USER` from selected SAML attributes and httpd writes out `REMOTE_USER` into the access logs, next to `REMOTE_ADDR`.).
* Then head back to FileSender's admin UI and `Search for user (based on user ID)`, now that we know the userid.

That seems (needlessly) cumbersome, to say the least.

peter-:
That's the kind of thing [you](@monkeyiq ) were looking for in https://github.com/filesender/filesender/issues/747#issuecomment-647286751, I guess.

monkeyiq:
I will take a look at adding search for sender email to the admin/transfers page shortly. The "Show details" page is using the column transfers.user_email from the database as the sender email so I think if we can find transfers based on that then it might help this admin situation. To assist for larger databases we might like to add a secondary index for the transfers.user_email column which would make this search for exact match very quick.

monkeyiq:
Feel free to reopen if there is still an issue.

PR #894, opened 2020-12-14
admin: allow search for transfer by sender email

monkeyiq:
This should allow you to find a transfer given the sender email address. This is to help administration as highlighted in https://github.com/filesender/filesender/issues/888.

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -245,6 +245,9 @@ class Transfer extends DBObject
     protected static $secondaryIndexMap = array(
         'userid' => array(
             'userid' => array()
+        ),
+        'user_email' => array(
+            'user_email' => array()
         )
     );
 
--- a/language/en_AU/lang.php
+++ b/language/en_AU/lang.php
@@ -612,3 +612,6 @@ $lang['you_can_report_exception'] = 'When reporting this error please give the f
 $lang['you_can_report_exception_by_email'] = 'You can report this error by email';
 $lang['you_can_send_client_logs'] = 'In order to help your support team to find out what happened you can send the last log entries from your user interface by clicking this button :';
 $lang['you_generated_this_auth_secret_at'] = 'You generated this auth secret at: {datetime}';
+$lang['sender_email_search'] = 'Sender email';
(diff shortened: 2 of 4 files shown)


peter-:
Thanks, but after updating to 2.21 and running the database.php script without any errors ...
```
...
All core code worked (leaving foreign keys), commit to database starting...                                                                                                                                                                    
Commit went well, those changes are now permanent                                                                                                                                                                                              
Performing FOREIGN KEY creation...                                                                                                                                                                                                             
Everything went well                                                                                                                                                                                                                           
Database structure is up to date
```
unfortunately I can't get this work:

* Unlike the search for userids, which also finds users based on partial string matches (e.g. "foobar" for user "baz.foobar@example.org") this seems to require something resembling a full email address. That's OK as a start (but slightly less helpful than the search for userids).
* When I just tested this it didn't find any transfers, though, whether closed or yet unexpired:
  * In the case of something less than the full email address it didn't seem to do anything, other then reload the currently displayed list of all transfers.
  * In the case of searching for the literal and full sender email adress of existing (unexpired or closed) transfers it only ever showed an extra section above all transfers without any results, but including the text "No transfers" (and the paging indicators to page trough all existing transfers for this "no result" list of transfers, too).

This is on Postgresql 11 using the standard Debian packages.  
(Also let me know whether I should report that elsewhere, I didn't know where would be most appropriate.)

peter-:
Hm. Here too I have to qualify that report some more: I can find some transfers using the literal, full sender email address, but not others. I'm still searching for a pattern.

monkeyiq:
I had made this search for exact equality on the email field rather than partial match. Thinking some more I can see a number of reasons one might like to perform a search by pattern. For example, only having part of the email address for privacy reasons, or not knowing the exact display name the user might have used.

My rationale for exact matching was performance; it could use the index for quick query resolution. Though that is not really the main factor if it doesn't allow easy use due to only knowing part of the email address and wanting to use a pattern match to get the job at hand done.

I will add pattern search shortly and retain the exact match search option as an option (explicitly saying it is "full match" in the web UI) for cases where it can work and the administrator may want the performance gain from the index.

PR #895, opened 2020-12-14
ui: allow going back from generated to user selected password

monkeyiq:
Allow the user to go back to user selected password when they currently have a generated password on the upload page. Generated passwords are treated slightly differently to user entered passwords as the generated ones are more likely to use the full entropy of each byte. This can have an impact when password hashing is used on user entered passwords.

To get around the user modifying a generated password they are set to read only. To go back to a user entered password the selection has become a checkbox with the generation of new passwords moved to a second button which is only available when generated passwords are used. I think retaining the generate button is good because the user might feel better about the appearance of some passwords over others.

This is related to https://github.com/filesender/filesender/pull/885.

### DIFF
--- a/language/en_AU/lang.php
+++ b/language/en_AU/lang.php
@@ -614,4 +614,4 @@ $lang['you_can_send_client_logs'] = 'In order to help your support team to find 
 $lang['you_generated_this_auth_secret_at'] = 'You generated this auth secret at: {datetime}';
 $lang['sender_email_search'] = 'Sender email';
 $lang['search_transfer_by_sender_email_description'] = 'Search transfers by sender email address';
-
+$lang['generate_a_different_password'] = 'Generate a different password';
--- a/templates/upload_page.php
+++ b/templates/upload_page.php
@@ -199,7 +199,14 @@ if(Auth::isGuest()) {
                             {tr:file_encryption_password_too_short}
                         </div>
                         <div class="fieldcontainer" id="encryption_password_container_generate">
-                            <a id='encryption_generate_password' href="#">{tr:file_encryption_generate_password}</a>
+                            <input id="encryption_use_generated_password"  name="encryption_use_generated_password" type="checkbox">  
+                            <label for="encryption_use_generated_password" style="cursor: pointer;">{tr:file_encryption_generate_password}</label>
+                            
+                        </div>
+                        <div class="fieldcontainer" id="encryption_password_container_generate_again">
+                            <a href="#" id="encryption_generate_password" class="">
+                                <span class="fa fa-refresh"></span>&nbsp;{tr:generate_a_different_password}
+                            </a>
                         </div>
                         <div class="fieldcontainer" id="encryption_password_show_container">  
                             <input id="encryption_show_password" name="encryption_show_password" type="checkbox">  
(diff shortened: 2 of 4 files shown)


peter-:
Thanks. I was only now able to test this and I'm sorry to report have a few issues with this:

* `Generate a different password` is displayed even when not having selected `File Encryption` (which I think doesn't make much sense), so this should be moved into the section that only becomes visible once `File Encryption` has been checked.
* I don't see how the issue from #885 and #880 is now handled: There is now an extra checkbox `Generate password` but once a password has been generated (including unintentionally) there still is no way to undo that / to change your mind and supply a password of your own?
* TBH, the `Generate password` checkbox seems like a noop to me:
  * If I generate a password _before_ checking that box (including by clicking `Generate a different password` before even checking `File Encryption`) then nothing I do allows me to supply a password of my own (other than restarting the whole "create a transfer from scratch" process).
  * If I don't select `Generate password` I can enter or remove or change passwords as I desire, but once I click `Generate a different password` I'm back in the same position as previously (about which the aforementioned issues were raised).

I.e., other than introducing more UI elements (and therefore complexity, esp. wrt their interaction and permutation) it seems the reported issue has merely moved from the "Generate password" checkbox to a `Generate a different password` link: Once you click the latter (same as before with the the former) you're stuck and can't undo what you just did (which is not good UI/UX, IMO: If I can do/select something with a click I expect to undo that with another click or two, no)?

Here's an attempt to illustrate how this looks for me. Sorry about the bad quality (GH doesn't allow certain file extensions and I didn't try to smuggle in a proper video file with a `.gif` file extension, so I actually converted the video to animated GIF) and the lack of clear indication when I clicked on something:

fs-encpwd

What happens there: I start this off by clicking `Generate a different password` before ever checking `File Encryption` (to see what happens), then once I do that, too, I see that I'm now stuck with generated passwords -- same as before this fix. Clicking around (incl. checking and unchecking `Generate password`) doesn't change that.  
At some point I restart the transfer from scratch (no other way to restart "clean") and find out that `Generate password` doesn't do anything obvious in that mode either: In checked or unchecked state I can enter passwords as desired. In checked or unchecked state I **cannot** enter passwords as desired once `Generate a different password` has been clicked.

peter-:
Not sure where else I should have reported that? I can't re-open my original one, should I have create a completely new one (when there's already #880 and #885 and this PR)?

peter-:
Interesting. Deactivating `encrytption_enabled` in `config.php`, saving, reloading the browser window and then undoing the change (i.e., re-enabling `encrytption_enabled`) and reloading the browser window again made everything work and all of the things I reported above untrue:

* `Generate a different password` is no longer shown before checking `File Encryption`,
* In fact it's not even shown before also checking `Generate password`
* And once `Generate password` is unchecked it allows to change/delete the generated password, as was requested.

So thanks and **sorry for the noise**. I cannot explain the reported (and at least documented, so there's some evidence it's not just in my imagination) behaviour, it must have been some caching issue with the browser having been logged in since before the upgrade. (I'd rule out something server or process-based as I did perform a reboot after the update from 2.20 to 2.21 to change unrelated hardware settings on the VM.)

PR #901, opened 2020-12-21
admin: allow transfer search by partial sender email

monkeyiq:
This is a follow up to https://github.com/filesender/filesender/pull/894 which moves the default search to a LIKE and retains the previous exact match search as an explicit option (not enabled by default).

This should give flexibility by default and possible improved performance when you have the exact full email at hand.


monkeyiq:
Initially I made partial match the option but later thought that the checkbox option should be for exact match and the default should be for a LIKE match. If you think you have the entire email address then you will likely know it.

Issue #916, opened 2021-02-17
Filename containing multi-byte character(s) is shorten incorrectly


hironobu-nagaya:
Hi,

Sometimes, shorten file name(s) on transfer table becomes empty.
I noticed it occurs only when filename contains multi-byte character(s).
I traced code and I thought that multi-byte character(s) was not included in the consideration.

https://github.com/filesender/filesender/blob/7877cfb5dfa4adb3fb6017edcd06e042731e1206/templates/transfers_table.php#L245-L261

Here is script to check on Bash.

```
$ php << 'EOF'
<?php
// prepare dummy values.
class Lang
{
  public static function tr() {
    return 'See all';
  }
}
$file = new class{};
$file->path = "マルチバイト文字を含むファイル名.txt";
$transfer = new class{};
$transfer->downloads = [];

// FileSender 2.21 - templates/transfers_table.php - line 248-253
// BEGIN
$name = $file->path;
$name_shorten_by = (int) (strlen((string) count($transfer->downloads))+strlen(Lang::tr('see_all'))+3)/2;
    if(strlen($name) > 28-$name_shorten_by) {
        if(count($transfer->downloads)) $name = substr($name, 0, 23-$name_shorten_by).'...';
        else $name = substr($name, 0, 23).'...';
}
// END

echo $name . "\n";
echo htmlentities($name, ENT_QUOTES, 'UTF-8') . "\n"; // same as `Template::sanitizeOutput($name)`
?>
EOF
```

( "マルチバイト文字を含むファイル名" means "filename which contains multi-byte characters(s)" in Japanese )

And it outputs,

```
マルチバイト文...

```
( line 2 is empty )

Environment
-------------

* CentOS 7
* PHP 7.4 ( from [Remi's RPM repository](https://rpms.remirepo.net/) )

```
$ cat /etc/centos-release
CentOS Linux release 7.9.2009 (Core)
```

```
$ php -v
PHP 7.4.14 (cli) (built: Jan  5 2021 10:45:06) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
```

----

I feel sorry I am not good English speaker, and have little knowledge of PHP.
Furthermore, this is first time to submit issue to open source project.

Thank you.

monkeyiq:
[...]

PR #919, opened 2021-02-18
handle multibyte characters in strings better here

monkeyiq:
Directly using substr() might truncate a string in the middle of a multibyte character giving a poor experience for strings containing mbc. There is a follow on in that strlen() needs replacing with mb_strlen() if one expects the length of actual characters in the string when working out if truncation is required.


monkeyiq:
This should address https://github.com/filesender/filesender/issues/916

Issue #918, opened 2021-02-17
data protection: provide a summary/copy of personal data stored


peter-:
As discussed [on the list](https://sympa.uninett.no/lists/filesender.org/arc/filesender-dev/2021-02/msg00008.html) it might be a useful extension to optionally allow subjects to get a summary (and/or copy) of the personal data the system currently holds about them, as required by [GDPR Article 15](https://gdprinfo.eu/en-article-15) (i.e., which I think applies not only to deployers in the EU but also to any deployer of the software that might also serve citizens of EU member states).

> The controller shall provide a copy of the personal data undergoing processing.

I'm not convinved this has to be done as a UI element enabling self-service. A first implementation as a script to be called by an admin might be fully sufficient.

(FWIW, the data to be collected should be able to be derived from the data that would be deleted if the subject used the "delete all my data" button currently offered on their profile page.)

PR #923, opened 2021-02-18
Command line script to export a user's data

monkeyiq:
I would have liked JSON format instead of var_export but the later is used to avoid having to maintain methods that export all the fields to JSON if any new fields are added.


monkeyiq:
This is related to https://github.com/filesender/filesender/issues/918

peter-:
Thanks! I'll need to re-build my test machine so I can try this.

FWIW, no idea how reliable this would work (esp. since transfers may include user-provided data), but I've seen people [parse the outout of `var_export()` and convert it ito JSON](https://github.com/klesun-misc/dev_data/blob/master/common/php/VarExportParser.php). (Note that the referenced project doesn't currently have a license statement applied, but if that'd be something we could use one can always ask them to slap on compatble license terms.)  
[Another, historic, example.](https://www.phpclasses.org/package/4529-PHP-Unserialize-values-exported-with-var-export.html)

monkeyiq:
I was initially using SQL to directly search. But the objects all have matching data members for anything stored in the database and also have various accessor methods. I may again look at SQL and how to format the database query results into individual json streams.

Another option would be to use introspection on the objects and generate one or more html tables in a single html file. Though then you start bumping into how much you are just mirroring much of the existing tables and what is presented in the normal UI.

PR #956, opened 2021-03-16
Docker CI workflow

WebSpider:
Adds a CI workflow that runs upon release.
The workflow builds dockerimage defined in /docker/Dockerfile (such as the one in #955 ), and then uploads it to dockerhub

Credentials for dockerhub are assumed to be available as a secret in the organisation or the repository

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -19,5 +19,5 @@ jobs:
           tag_with_ref: true
           tag_with_sha: true
           path: docker
-          build_args: BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%Sz'),VCS_REF=${GITHUB_SHA:8}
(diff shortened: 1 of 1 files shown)


monkeyiq:
It seems that the dockerfile will create an image from FILESENDER_VERSION using the released tarball. Does this change if the dockerfile is run from the github release action?

WebSpider:
Those two are not integrated right now. 

I'll push some commits to this branch to make the Dockerfile accept a version during the build process, and the CI to detect/export the version.

monkeyiq:
So I am guessing this is ready to merge now / soon?

monkeyiq:
I will have a discussion with the board as to which account level is desired for running these scripts.

WebSpider:
Well. That wasn't supposed to happen. 
I'll re-implement once we have the board's decision.

Issue #966, opened 2021-03-20
Ability for admin to extend the expire time for guest vouchers.


monkeyiq:
Some method for an admin to extend the expire time of an existing guest voucher. 

Perhaps the admin can also set a default for a user so they have vouchers that last $x days by default instead of the system default. Handy to allow long term known users to have vouchers that will cover a period of time that makes sense in their setting, for example two semesters at a uni.

monkeyiq:
This is a follow up to https://github.com/filesender/filesender/issues/875

peter-:
> Perhaps the admin can also set a default for a user so they have vouchers that last $x days by default instead of the system default. Handy to allow long term known users to have vouchers that will cover a period of time that makes sense in their setting, for example two semesters at a uni.

That actually closer resembles our current use-case: As part of a research project a user of the system needs to allow several external cooperators to upload files to her over an extended period of time. Currently the vouchers constantly expiring makes this somewhat impractical, having to send out new vouchers every week or so. (And the institution isn't keen on issuing guest accounts to all external project partners to access filesender as users themselfs, due to their remoteness, potential number, and all other reasons that guest accounts suck.)  
So allowing certain users of the system to issue (multiple-use) vouchers with a much extended validity would be great -- but seems like additional functionality on top of making it possible to extend the validity (by the user and/or the admin) in the first place?

monkeyiq:
The above allows the extension of the guest validity (983) and the ability to set specific users to have higher defaults for the guests they create (984). I have then moved the existing code to use the abstracted out code I made in the above in 985. 

Allowing some users to be able to create guests with longer than default expire times did expand the scope of things. Though if it captures scenarios that are likely to be common in many installations I think it is a good extension to make.

WebSpider:
This feature was shipped with filesender 2.23

PR #985, opened 2021-03-28
move transfers to use the new shared code for expire updates

monkeyiq:
Now that there is a common use version of this code in DBObject and ui.js the old version of this code should really be removed from use in Transfer management. This has both Guest and Transfer use the same new code that is based on and extended from the old code. Having two code paths around is a bad idea for maintenance.


monkeyiq:
This moves the existing code path to the new more common code in DBObject that was done while addressing https://github.com/filesender/filesender/issues/966.

PR #987, opened 2021-03-31
Add file integrity validation to StorageFilesystemChunked class

mdusher:
This may need some review, but this commit adds:
- Hashing of the incoming stream and compare to what ends up stored on disk
- Verification of the upload size
- Update javascript to handle FileIntegrityExceptions and add a waiting state while a file is hashed.
- Refactor StorageFileSystemChunked


monkeyiq:
Thanks for the update!

PR #988, opened 2021-04-05
core: encryption: do not base64 encrypted data during upload

monkeyiq:
This is a huge bottleneck. Some empirical data to show just how large the impact is. The upload speed would ramp up slowly during an encrypted upload on a local connection and for large files (4gb) might get to 85mb/s at the end. Without the base64 encoding and decoding the speed goes over 200mb/s on upload. I am running with 12 terasender workers and 4mb chunk size for these numbers.

I have made a temporary config setting to allow enabling base64 for both local speed testing but mainly for backward compatibility if there are issues discovered in some installations and we need to support base64 on the PUT contents for some reason.


monkeyiq:
This will have a positive impact on all encrypted uploads not just local ones.

PR #994, opened 2021-04-13
responsive menu and favicons for all the menu items added

zivel:
Hi Ben (I assume you're reading this)

I made the menu a tad more responsive and added favicons to all the menu items. When the width of the screen gets too narrow, the menu becomes a 'hamburger'. 

Cheers
- Renato


madsi1m:
I like this also

monkeyiq:
Hi Renato, thanks for the update!

Issue #1000, opened 2021-04-22
Behaviour of user_can_only_view_guest_transfers_shared_with_them


monkeyiq:
This was reported on the mailing list. I am duplicating here to allow interested parties to track it.

```
To be exact, when this option is turned off, and a user creates a guest voucher, who then creates a transfer that is not shared, this transfer:

1- Is visible in the mytransfers overview of the user who created the guest voucher
2- Causes email notifications to be sent to the user
```

monkeyiq:
An update is available at https://github.com/filesender/filesender/pull/1001

WebSpider:
This fix seems to be only in development and filesender-3. Will this be backported to filesender-2.x as well?

monkeyiq:
The above update (1001) should be in 2.25 and later: https://github.com/filesender/filesender/releases/tag/master-filesender-2.25

If this is still not working as expected in 2.25 and over I will have another dig at it.

I was tempted to rename development to development2 but left development and master as the version2 trees for now to avoid everybody having to rename branches. In the future once FileSender 3.x is released and the 2.x series is more 'parked' I will likely rename these 2.x branches to development2 and master2 so that people do not accidentally use the 2.x series for a PR.

PR #1001, opened 2021-04-22
Update to user_can_only_view_guest_transfers_shared_with_them for wider scope

monkeyiq:
This is to address https://github.com/filesender/filesender/issues/1000. I have also looked at the emails that are sent though it is possible that this PR does not catch all "guest emails" that might also trigger an email to the user who invited the guest.

Because this test is needed in multiple locations I have created a new `$guest_transfer_shown_to_user_who_invited_guest` column in the Transfers table. This allows SQL to check and also the code to directly check if an email should be sent.

The `$guest_transfer_shown_to_user_who_invited_guest` is set in RestEndpointTransfer.class.php when a transfer is created and remains the same thereafter. The default is to show transfers to the user who owns them. Guest transfers are owned by the user who created them. So this whole config option is trying to hide some transfers from a user with the userid who owns some transfers because those transfers are from a guest of the user.

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -683,6 +683,16 @@ class Transfer extends DBObject
         
         Logger::info($this.' deleted');
     }
+
+    public function userCanSeeTransfer()
+    {
+        if( !$this->guest_id )
+            return true;
+        if( $this->guest_transfer_shown_to_user_who_invited_guest ) {
+            return true;
+        }
+        return false;
+    }
     
     /**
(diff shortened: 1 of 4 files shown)


WebSpider:
Will this be backported into filesender-2.x as well, or is this only filesender-3+ ?

monkeyiq:
This should have been released as part of [Release 2.25]( https://github.com/filesender/filesender/releases/tag/master-filesender-2.25).

PR #1002, opened 2021-04-24
NO MERGE: testing selenium in docker

monkeyiq:
making a container


monkeyiq:
This seems like maybe it could generate videos in GHA but I don't know if the experience for finding regressions would be inviting.
https://github.com/SeleniumHQ/docker-selenium

PR #1009, opened 2021-04-27
i18n: auto import from poedtor on 2021-04-27-1619502721

monkeyiq:


monkeyiq:
I am bringing in changes in existing languages before adding another language.

PR #1055, opened 2021-05-27
cookie: updates to default config and install guide

monkeyiq:
The SimpleSAMLphp part of the installation did not take advantage of some of the options for SimpleSAMLphp for the default configuration settings. This introduces a recommendation to use SimpleSAMLphp configuration options that will result in SameSite=Strict and HttpOnly being used to protect the auth cookies.

The OWASP CSRF file was updated to use SameSite=Strict but can not use HttpOnly as the cookie must be explicitly accessed from javascript.

The apache overrides recommendations have also been updated if that is the method an admin wishes to add these cookie options to their environment.

This relates to 20100861-L1 and 20100861-N1


monkeyiq:
This also required php 7.3+ for the more explicit cookie settings.

Issue #999, opened 2021-04-21
Generated password feature does not respect encryption directive.


michaellegal:
Hi, I have that kind of config directives and it seems the generate password feature does not respect the values:

$config['encryption_mandatory'] = true;
$config['encryption_min_password_length'] = '12';
$config['encryption_password_must_have_upper_and_lower_case'] = true;
$config['encryption_password_must_have_numbers'] = true;
$config['encryption_password_must_have_special_characters'] = false;
$config['encryption_generated_password_length'] = 12;
$config['encryption_generated_password_encoding'] = 'base64';

The generated password button still offer something like that: PZPP59uA86FAxMTDisgk2VeO+5FlhmIVEbQjDrxQWaM=

Best regards,

Michael LE GAL

monkeyiq:
Thank you for raising this issue, Some of the documentation is stale. I will update the documentation soon to reflect why this behaviour happens. 

The FileSender project is lucky enough to have input from a professional cryptographer. In line with their suggestions generated passwords are made based on how much entropy is desired for good crypto. Those random bytes are then put into a format that can be displayed as text and presented as the generated password. The goal is to make the generated password as secure as we can for file encryption. Many of the encryption configuration directives such as password_must_have_numbers do not apply to generated passwords. I will update the documentation to outline which options have no effect for generated passwords.

monkeyiq:
I have updated the docs to deprecate encryption_generated_password_length in https://github.com/filesender/filesender/pull/1068.

For now you could set encryption_random_password_version_new_files=1 to use encryption_generated_password_length. However it is highly recommended to retain the system default for generated password length.

PR #1068, opened 2021-06-11
docs: update that encryption_generated_password_length is deprecated

monkeyiq:
Since FileSender 2.9 (released Oct 5, 2019) encryption_random_password_version_new_files has been 2 which is v2019_generated_password_that_is_full_256bit (from crypto_app.js). This password handling has many advantages such as generating a string that contains the entropy that the system desires and also being a password format known to be generated this way it may skip password hashing.

As the encryption_generated_password_length only applies to v2018_text_password I am deprecating it here. Unless you are setting encryption_random_password_version_new_files to 1 explicitly and then looking to set the max length this setting would have no effect. It is strongly recommended to use the default encryption_random_password_version_new_files=2 (or leaving that setting unset to get the default) as introduced in FileSender 2.9 (released Oct 5, 2019).

### DIFF
--- a/docs/v2.0/admin/configuration/index.md
+++ b/docs/v2.0/admin/configuration/index.md
@@ -1256,7 +1256,7 @@ If you want to find out the expiry timer for your SAML Identity Provider install
 
 
 ### encryption_generated_password_encoding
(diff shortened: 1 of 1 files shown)


monkeyiq:
This documentation update relates to https://github.com/filesender/filesender/issues/999.

PR #1097, opened 2021-07-01
docs: Setup initial v3.0 docs tree

WebSpider:
Taking 2.0 docs as base

### DIFF
--- a/docs/_includes/menu.html
+++ b/docs/_includes/menu.html
@@ -30,4 +30,8 @@
 	<ul>
 		<li><a href="{{site.baseurl}}/v1.6/">Index</a>
 	</ul>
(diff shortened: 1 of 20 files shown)


monkeyiq:
I have included some more information about the 3.x series in the docs with https://github.com/filesender/filesender/pull/1170. I am trying to delay copying the 2.0 docs to the 3.0 docs tree to a bit closer to the release. I would hope for a 3.0 release some time in 2022. Though maybe with beta releases first. The decision is in the hands of the FileSender board.

Maybe something like markdown include could be used to create docs/3.0 which just links to the docs/2.0 until it needs to diverge.

monkeyiq:
I think with the above comment and the updated that have happened to the docs/v2 tree in the meantime I might close this PR for now.

PR #1099, opened 2021-07-01
added encryption symbol (key) to transfers table

baumrasen:
We use this to show encrypted transfers beside the transferid in the transfers table. Maybe it finds the way to the origin filesender.
image


WebSpider:
This is a great addition!

monkeyiq:
I like the idea, thanks for the PR :)

I have a slightly modified version of this PR that uses the font awesome icons that are used elsewhere on the page to show the same thing.
https://github.com/filesender/filesender/pull/1101

PR #1100, opened 2021-07-01
Release automation: Initial setup

WebSpider:
This adds an initial setup for release automation, meant to speed up the release cycle of filesender by automating parts of it.

The initial setup currently only drafts the release body, from a template extracted from the filesender-2.25 release.

It does this by checking if a new tag is pushed to the repository. If this happens, the action gathers data from the current state, the previous release, and commits in between, and fills the template, which it then attaches to a draft release.

The template is a seperate file `release-body.md`, with variables that are to be filled.

This obviously needs to be expanded upon, for instance by:

1. Agreeing on a commit format, so we can automatically add bits from commits into the release notes (I just picked an action that creates something now, so you get an idea)
2. Removing the draft marker, so that a new tag automatically means a new release
3. Finding a way to handle deprecation notices and documentation updates. These sections are generated in draft, but currently not filled.
4. Whatever seems like a good thing to do whenever a release is made

### DIFF
--- a/.github/workflows/release-body.md
+++ b/.github/workflows/release-body.md
@@ -1,0 +1,27 @@
+## Release Version {{ .to_tag }}
+
+Release date: {{ .date }}
+
+## Distribution
+Source snapshots are attached to this announcement and the git tag `{{ .to_tag }}` contains the base that these snapshots were created from.
+
+## Installation
+Documentation is available at http://docs.filesender.org/v2.0/install/
+
+## Major changes since {{ .from_tag }}
+{{ .db_changed }}
+{{ .templates_changed }}
+
+{{ .changelog }}
+
+## Configuration changes
+
+
+These options are detailed in the docs/v2.0/admin/configuration/index.md file as usual.
+
+## Deprecations
+
+
+## Support and Feedback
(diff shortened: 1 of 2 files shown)


monkeyiq:
This looks like a very handy starting point for release notes drafts! 

My current plan when doing a release to start looking for changes that need database updates are to check the `$dataMap` for classes in the data directory for meaningful updates. For the purposes for automation we might just say that any change to that php variable is meaningful as most non white space only changes are likely to be meaningful. The secondary check is if `$secondaryIndexMap` has changed. The third check is that `getPrimaryViewDefinition()` and `getViewMap()` have changed.  Normally for the later two I am the main one to change them so I tend to also have a feel for if and where I should look to find that change. It seems to me that the rest of the code the the classes in data/*php may be able to change without changing the database schema.

Another check for if the database script must be run is if the `ensureFK()` function changes in scripts/database.php.

These database checks can be updated post merge and it is probably a good plan to capture that in the action and document in the dev docs that it is a good place to look for what the system considers when doing a release.

I might change the on/push trigger to only consider master and master3. I normally do master3 updates via a push and master updates via a PR (eg https://github.com/filesender/filesender/pull/1122).

PR #1114, opened 2021-07-08
updating the files and scripts

kerins:
There is an issue with AWS and our S3 provider where there is a limit of 100 buckets. 
For this reason we feel its better to run the files from one bucket which gets defined in the config
rather than creating a bucket per file. These are the changes we are using in production.

### DIFF
--- a/classes/storage/StorageCloudS3.class.php
+++ b/classes/storage/StorageCloudS3.class.php
@@ -98,8 +98,8 @@ class StorageCloudS3 extends StorageFilesystem
             $offset=$offset/Config::get('upload_chunk_size')*Config::get('upload_crypted_chunk_size');
         }
 
-        $bucket_name = $file->uid;
-        $object_name = self::getObjectName($offset);
+        $bucket_name = Config::get('cloud_s3_bucket');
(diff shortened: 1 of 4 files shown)


monkeyiq:
I am wondering if it would be an idea to pull this code into a method so that the old and new storage formats both work. Perhaps making `$bucket_name` and `$object_name` object variables which are set by a method at the start of calls. This would at least allow people with systems running the old format the ability to update FileSender without running into trouble. Maybe the data/File class could have a storage_class_parameters to match the storage_class_name database variable. So the S3 class could save how the file was stored and allow future live migrations (new files to new layout, old files still accessible).

I may take a look at making and testing an update to do this in the next weeks.

monkeyiq:
I have created a new PR https://github.com/filesender/filesender/pull/1175 which should allow this layout and the old layout and migration between the two while still allowing already uploaded files to continue to be downloaded.

monkeyiq:
Let me know if I missed anything in the update in PR https://github.com/filesender/filesender/pull/1175 otherwise things should continue to work for you in the next and future releases. Thanks for the PR, I thought it best to get in and rework it to 1175 and merge.

PR #1115, opened 2021-07-09
enable fill on graphs (lost/new setting when updated chart.js)

madsi1m:


monkeyiq:
I think those might have got side tracked in my CSP header chart update. Thanks for the PR!

Issue #1052, opened 2021-05-26
Cron + database: an secondary index on 'Transfers.expires'


monkeyiq:
The cron job looks to trim Transfers that have expired based on a date comparison on `Transfers.expires`. 
Adding a secondary index on `Transfers.exires` will help this and other cases that are specifically looking for aged transfers.

PR #1129, opened 2021-08-22
add secondary index on transfers.expires to help cron execution times

monkeyiq:
This relates to https://github.com/filesender/filesender/issues/1052


monkeyiq:
The next release has some db additions, so it is a good time to also add this sec idx

Issue #1140, opened 2021-09-10
Run Tests without Sausage Package. 


vinnyalvs:
Hello. I work with the Brazilian National Research Network implantation of Filesender (https://filesender.rnp.br/), and we do some little customizations and we wanted to test them with Selenium. That said, we don't want to use the current tests implementations because:

- We don't want to use Sauce Labs
- We don't want to use the Sausage Package
- The tests use an old version of PHPUnit 

Hence, we build a new Base class for Selenium Tests, updated the tests to run with newer PHPUnit versions, and changes the requirements in composer.json File. We made a pull request with these modifications, and we want to discuss what is best for the project.  We did not want to break what currently exists, so we created new files instead of replacing old ones. 

I don't know if this is the best place to ask this, but we want to know **why** did you choose to use the Sausage Package and **why** do you run the selenium tests in another repository.

monkeyiq:
Hello and welcome!

This is a good place to ask the above. Either here on the dev mailing list. The selenium tests are run in another repository because they are managed by savage which checks if somebody is trying to update the test suite to do nasty things. The selenium test suite needs to contact sauce labs so it is a good idea to do some checks before running tests from PRs from unknown sources.

The sausage package was there when I arrived. Not a great answer but the only one I have. I tried at one point to update to more recent versions for the test suite but met with resistance on the sauce labs connection so left it where it was for now. I would be happy to use a newer version of phpunit.

A partial fragment of my /opt/filesender/config/config.php on the selenium test VM:

```
  $config['testsuite_run_locally'] = true;
  $config['auth_sp_type'] = 'fake';
  $config['auth_sp_fake_authenticated'] = true;
  $config['auth_sp_fake_uid'] = 1;
  $config['auth_sp_fake_email'] = 'memoryhole@localhost.localdomain';
```

This VM runs the tests locally without Sauce Labs. I also set a sauce username though that shouldn't be used because a local Chrome in the same VM is used to run the test pass.

```
export SAUCE_USERNAME=ben
```

PR #1141, opened 2021-09-10
New Base Class to run Selenium Tests Without need of sauce/sausage

vinnyalvs:
Please view issue #1140 . 

This Pull Request brings in a new class called BasisTest which permits the run without the need of Sauce Labs or the Sausage package.  

 We did not want to break what currently exists, so we created new files instead of replacing old ones.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,8 +2,11 @@
     "require": {
         "sauce/sausage": ">=0.15.2",
         "appium/php-client": "@dev",
-        "phpunit/phpunit-selenium": "^4.1",
         "squizlabs/php_codesniffer": "3.*",
-        "friendsofphp/php-cs-fixer": "^2.13.0"
+        "friendsofphp/php-cs-fixer": "^2.13.0",
+        "php-webdriver/webdriver": "^1.11",
+        "phpunit/phpunit": "7.5.20",
+        "phpunit/phpunit-selenium": "7.x"
     }
 }
+
--- a/composer.lock
+++ b/composer.lock
@@ -4,19 +4,19 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "62dce9b10d07f5c53fa690124356385a",
+    "content-hash": "d8bc785181f5aead33b0b4f4c4589cd3",
     "packages": [
         {
             "name": "appium/php-client",
             "version": "dev-master",
             "source": {
                 "type": "git",
-                "url": "https://github.com/appium/php-client.git",
+                "url": "https://github.com/appium-boneyard/php-client.git",
                 "reference": "cb9049b380beb5dec07bedbfbcce8800bc7af1b4"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/appium/php-client/zipball/cb9049b380beb5dec07bedbfbcce8800bc7af1b4",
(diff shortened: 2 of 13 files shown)


monkeyiq:
Thanks for the PR. 

I have a local VM setup to run Selenium tests against using the same code that is used with sauce labs. This uses the older phpunit and the other bridge stuff that sauce uses. I was tempted in the recent github actions update to try to run selenium in docker and do the test suite locally on a GHA. The current downside to doing that was getting failure feedback. It is handy to be able to playback a video some times to see what was failing in the test. The older phpunit currently used in development is less than optimal, but IIRC it was locked by the version of the bridge code used to contact sauce labs.

I have recently updated half of the selenium tests to work on the development3 branch https://github.com/filesender/filesender/pull/1157. I have the other half slated for action soon. As part of that update many tests were updated to check for elements and visibility before progressing rather than just sleep() blindly. Also I made some methods in the SeleniumTest class to do specific actions or to busy wait on some css class becoming visible for example. Long story short there was a fair bit of code updates for the `SeleniumTest.php` base class and the tests themselves. So we may want to avoid duplicates of those files if possible.

Looking at a diff of SeleniumTest.php and BasisTest.php the major change seems to be the base class being `Sauce\Sausage\WebDriverTestCase` and `\PHPUnit_Extensions_Selenium2TestCase` respectively. It seems most of the other blocks could be updated to be conditional, such as the sauce code in `__construct()`.

Would it be reasonable to try to have the github code inherit from `WebDriverTestCase` and a small script which could be run after a git checkout to update that php file to use `PHPUnit_Extensions_Selenium2TestCase` instead? If the script was fairly small it is probably less of a maintenance issue than trying to keep copies of the tests in sync.

Looking at the differences in ConfigurationOptionsTest.php (taken at random) it seems the biggest change is a method name going from byCss() to byCssSelector(). I am wondering if we couldn't use a common method in the base SeleniumTest which uses reflection to forward to the right place (parent class byCss() if a child of WebDriverTestCase and byCssSelector() if a child of PHPUnit_Extensions_Selenium2TestCase.

vinnyalvs:
Thank you for your answer!

What do you think is the best way to conduce this update?  Do you think the best is to do the modifications in the development3 branch with your updates in the SeleniumTest.php base class and the tests? Of course, we could do your suggestions like creating the common method in the base SeleniumTest which uses reflection and the git checkout script. 

But do you think the best is to do this on the 2.X version or with the 3?

monkeyiq:
One question that is probably best to address first is if the goal of running tests locally (without Sauce Labs) is the main objective.  If the use of sauce and an older phpunit can be lived with then we might be best to direct attention towards documenting what is needed to setup a local execution environment. 

I have one such environment but cobbled it together and once it worked I did not remake another one again from scratch and document the process. If this local execution environment is acceptable then I could turn my attention to making public docs for making such setups.

vinnyalvs:
If you could write down the steps to run locally, it would be great. I think a good place for the documentation it's [here](https://docs.filesender.org/filesender/v2.0/dev/testing/). Also, it's needed to update certain parts of the documentation related to tests like [the reference page in FrontEnd tests section](https://docs.filesender.org/filesender/v2.0/admin/reference/#:~:text=real%20unique%20one.-,Frontend%20tests,-Requirements) where it explicitly says that a SauceLabs API key is a requirement.

monkeyiq:
I will propose this documentation update to the FileSender board for the next MoU. The local VM setup documentation will need to have multiple runs from scratch to make sure that I don't accidentally leave anything out that is needed to get fully up to speed.

monkeyiq:
I think since these tests can be run locally already I will close this PR. There are now also tests against 3.x series which makes splitting this way a little less practical.

PR #1180, opened 2021-11-24
Fixes to allow running under PHP8

TheMysteriousX:
This isn't comprehensive by any means, but to get FileSender to run under PHP8 we needed to make these changes.

No behaviour should be changed, as it's just a formatting change, and dead code removal.

### DIFF
--- a/classes/utils/Utilities.class.php
+++ b/classes/utils/Utilities.class.php
@@ -87,8 +87,8 @@ class Utilities
         // Generate 16 bytes of random data (128 bits)
         $bytes = random_bytes(16);
         // Set bits required for a valid UUIDv4
-        $bytes{8} = chr((ord($bytes{8}) & 0x3F) | 0x80); // Eat 2 bits of entropy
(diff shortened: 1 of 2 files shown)


monkeyiq:
Thank for the PR! 

The {} to [] is a great update. 

I am still investigating if `get_magic_quotes_gpc()` could have returned anything other than `false` for php 7.0 through 7.4.

monkeyiq:
The `get_magic_quotes_gpc()` body is basically `RETURN_FALSE` in php-7.0.32 source.

monkeyiq:
It is false in 7.0.0 (released in 03 Dec 2015) as well and php 7.3. Since 7.3 is the minimum version of php cited on the installation of filesender [1] I think it has been a very long time since `if(get_magic_quotes_gpc())` has ever returned true.

[1] https://docs.filesender.org/filesender/v2.0/install/

monkeyiq:
For completeness I wanted to see when `get_magic_quotes_gpc` actually did something other than return false. In php 5.3.0 it returned a configuration value, in 5.4.0 it was hard coded to false.

PR #1181, opened 2021-12-13
Patch 1

liedekef:
It seems to me functions are called that don't exist during install and upgrade. I wonder how this ever even worked ...
Also the version number is not being maintained.


monkeyiq:
the version.txt being in the /scripts/... path should protect it from being world read but it is still something that I would like to clean up so I will make a new issue for it.

Issue #1223, opened 2022-03-16
Feature request: Store and reproduce folder layout


WebSpider:
When uploading a complete folder, using either encrypted or unencrypted transfer options, as an enduser I would like to be able to ensure the folder structure is retained when the transfer is downloaded as a .zip file.

monkeyiq:
This seems to be done at the upload page. 

If I click to send an entire directory then the path information is lost. The files list does not show paths and the resulting download of tar/zip doesn't have then either.

If I drag and drop a directory to the "Drag or choose file(s) to send" part of the web page then the files to upload list shows directory paths intact and the resulting transfer can be downloaded with paths included.

I will take a look at the upload page javascript as it is a real shame that this difference exists.

monkeyiq:
I just tested https://github.com/filesender/filesender/pull/1224 with encryption using "select a directory" and the path information was retained. I am still in the process of rebuilding a Edge-non-chromium env for testing that browser.

WebSpider:
We will be testing this PR this week. Thanks for the quick response!

WebSpider:
@monkeyiq unfortunately this breaks sending single files with a valid filename

monkeyiq:
Update added, sorry about that :/

monkeyiq:
I have now merged https://github.com/filesender/filesender/pull/1224 bringing the subdirectory name preservation into what will be the next release.

WebSpider:
We will be testing this as we prepare for 2.36 release

WebSpider:
This is fixed in 2.36, thanks a lot!

PR #1224, opened 2022-03-19
allow select directory button to store paths too

monkeyiq:
As reported in https://github.com/filesender/filesender/issues/1223 the functionality is different between adding a directory by drag and drop and adding by "select a directory". Specifically the drag and drop would retain the relative paths of the files in the directories added and the "select a directory" would only store the file names.

This patch allows the relative paths to be stored where the browser supports it for "select a directory" mode. This will also allow the files to be downloaded in a zip containing the paths.


monkeyiq:
Tested on Chrome, Firefox, Safari, and Edge-Chrome. Note that safari does not support the select a directory option currently.

WebSpider:
This PR breaks uploading of single files with a valid filename in Firefox (not tested in other browsers)

monkeyiq:
grr, too focused on testing out the select a directory path. The above push should fix the single file to work again.

WebSpider:
This seems to work on firefox!

PR #1225, opened 2022-03-19
no merge: initial version of mediumint to int4 for mariadb

monkeyiq:
The main failing of this is that foreign keys are dropped before the database.php changes mediumint to int4. This is a big brute force and also means that the database script really can not be run at the same time as the system is in use.

I am mainly pushing this up at this stage to make it easy to see the update.


monkeyiq:
This has been superseded by https://github.com/filesender/filesender/pull/1295

PR #1229, opened 2022-03-30
Add missing libraries and replace deprecated npm packages

Veredior:
As i was trying to test the new bootstrap version, i found out that the scripts of the www-lib-update folder was missing several packages. With this PR we should be able to update all javascript dependencies.

- Bootstrap 5.x uses popperjs version 2, so the version 1 package is switched for the version 2 one;
- flag-icon-css is deprecated, in favor of flag-icons which is the same package with a minor change of the CSS classes
- Deletion of the reset.css file as Bootstrap includes its own based on normalize.css


monkeyiq:
I noticed recently that somehow parts of scripts/dev/www-lib-update/package.json had been rolled back. I have brought back the versions there and will update development3 with your changes now too. 

First up is the popper.js update which I have put into https://github.com/filesender/filesender/pull/1252

monkeyiq:
Bootstrap 5.1 update in metadata in https://github.com/filesender/filesender/pull/1253

monkeyiq:
Thanks for the PR. I have put these in a small series of updates and merged them. Picking the updates out and doing it this way allowed the 1000 or so updated files in flag-icons to be updated via the scripts locally which made the reviewing easier. Hopefully I got everything in the smaller updates!

PR #1232, opened 2022-04-06
Correct English for user_id

liedekef:


WebSpider:
Translations are typically managed using poeditor.com, check out https://poeditor.com/projects/po_edit?id_language=198&id=48000

monkeyiq:
As a direct update to the lang.php will be overridden the next time a bulk import from poeditor is performed I am going to close this. As WebSpider mentioned above the best way to perform translation updates is by editing them on poeditor.com so they will be updated in the git repo by the scripts.

PR #1233, opened 2022-04-06
No CSRF check for remote users

liedekef:


monkeyiq:
Wouldn't it be better to leave the validateAgainstCSRF() call intact and inside the validateAgainstCSRF method at the top do something like the following. IMHO it is better for this exception to be inside the method itself so that it is explicit and in only one place.
```
    public static function validateAgainstCSRF( $canReturnJSON = false )
    {
      if ( Auth::isRemote() ) {
         return;
     }
     ...
}
```

liedekef:
Well, I did it like on line 294 in the original version (where an exception is thrown concerning xsrf), but your remark is also valid. I'm currently on holiday though and can't quickly create a new PR :-)

monkeyiq:
I also noticed the line 294 and wondered the best way to handle things. 
I might take a look at bringing in some of the changes in two smaller PRs myself shortly.

monkeyiq:
It seems from the loading code that `self::$translations['fallback']` is not an array of langcode but the codes are directly written into the fallback? 

https://github.com/filesender/filesender/blob/dc61226dd5d63820cbc942cf736e5ac2fb56f3fe/classes/utils/Lang.class.php#L415

I'm happy to dig more into this, it may help if I knew which language you are having an issue with so I can reproduce the issue you are/were seeing.

monkeyiq:
It seems from the loading code that `self::$translations['fallback']` is not an array of langcode but the codes are directly written into the fallback? 

https://github.com/filesender/filesender/blob/dc61226dd5d63820cbc942cf736e5ac2fb56f3fe/classes/utils/Lang.class.php#L415

I'm happy to dig more into this, it may help if I knew which language you are having an issue with so I can reproduce the issue you are/were seeing.

liedekef:
Is this still relevant in the latest version or should I create a new PR?

monkeyiq:
We could start with a PR to avoid the CSRF check for Remote clients. At the top of `validateAgainstCSRF()`. Explicitly falling out of an always called `validateAgainstCSRF` is much safer than a 3 clause if() being tested before the call to `validateAgainstCSRF`.

I would have to try to work out the differences in the Lang.class.php updates with respect to an array being present at one site and not the other.

monkeyiq:
Sorry about the delay. The FileSender board paused the project for 6 months.

monkeyiq:
I have put the CSRF part into https://github.com/filesender/filesender/pull/1353

I am not sure about the Lang.class.php update as I found I didn't have an array where it was expected in the PR. I guess some more digging is needed for that part.

monkeyiq:
If the Lang.class.php update is still desired please put that into another PR and we can discuss the array issue there. Thanks.

liedekef:
That's weird, as the code still matches for Lang.class.php ... I'll create a new pull request.

PR #1252, opened 2022-05-23
dev3: update popper.js handling code and version

monkeyiq:
This is a partial application of https://github.com/filesender/filesender/pull/1229.

In particular this PR updates the popper.js handling code and version only.


monkeyiq:
It seems that bootstrap was updated as part of this too.

PR #1262, opened 2022-05-31
doc fixes

peter-:
Mostly no longer working URLs (e.g. due to the SimpleSAMLphp web site changes), http links where https could be used, broken relative documentation URLs, etc.
(I started out with very targeted commits but ultimately lumpted together more changes in cf60ad2f5bf9284fe1f71c619fbb1961997ad6dc since you may want to squash all those on merge anyway.)

The docs are still a big mess IMHO but now with fewer broken links.

Also note that https://filesender.wordpress.com/ still exists but hasn't been updated in years.  
(The same goes for https://twitter.com/filesender which I left in `docs/socialmedia.md` for now, otherwise the page would essentially be empty after having removed references to the no longer functional http://blog.filesender.org/.)


docs/v2.0/install/index.md
```diff
@@ -232,28 +232,28 @@ authentication for development and testing. When you move to a
...
+* **NOTE**: you will of course remember to check [the sha256 hash of the tar file](https://github.com/simplesamlphp/simplesamlphp/releases), right?
 
 Extract SimpleSAMLphp in a suitable directory and create symlink:
 
 ```
 mkdir -p /opt/filesender
 cd /opt/filesender
-tar xvzf /root/src/simplesamlphp-1.15.4.tar.gz
-ln -s simplesamlphp-1.15.4 simplesaml
+tar xvzf /tmp/simplesamlphp-1.17.1.tar.gz
```
WebSpider, line 252: We're at 1.19.5 already :)

monkeyiq:
Thank you for the PR :)

Only one thing caught my attention while reading over things... I am not sure why session_cookie_path was removed in this PR. It seems to be still referenced in https://github.com/filesender/filesender/blob/development/includes/init/init_web.php#L45
It may be that the existing docs for that are slightly off but perhaps leaving that directive in for now and changing the docs for it separately is a good idea.

Other than that (which I am happy to bring back post merge) everything looks good.

monkeyiq:
Hmm, I see now that we went from 2 docs for session_cookie_path down to the correct one doc stanza for session_cookie_path :)

PR #1264, opened 2022-06-09
SAML logout update

monkeyiq:
Moving to using a POST request and allow CSRF check if enabled with
```
$config['owasp_csrf_protector_enabled'] = true;
```


monkeyiq:
From a technical point of view it was recommended that logout be done over POST and protected by CSRF.

thijskh:
The deployer needs to remove a file from SSP. And remove that file every time after they upgraded SSP?

monkeyiq:
This has now been seen as an acceptable risk as it was so I am reverting this update. It will always be available should the desire to remerge and move to this solution be seen as desirable by the board or one of it's members.

Issue #1276, opened 2022-06-30
FileSender cookie incompatible with draft-west-cookie-prefixes


WebSpider:
FileSender always sets the "domain" flag of the session cookie, looking at 

https://github.com/filesender/filesender/blob/20f141b8d1bcb1fd8798c2f8ca1a9ef5a3caa6f3/classes/utils/GUI.class.php#L282-L290

This is incompatible with https://datatracker.ietf.org/doc/html/draft-west-cookie-prefixes, which uses cookie prefixes to increase security of the cookie
This standard is implemented by all major browsers, and security auditors are starting to recommend using it.
The `domain` flag in this case has to be absent, because of its unintended side-effect, as per the [PHP manual on `setcookie()`](https://www.php.net/manual/en/function.setcookie.php):
```
The (sub)domain that the cookie is available to. 
Setting this to a subdomain (such as 'www.example.com') will make the cookie available 
to that subdomain and all other sub-domains of it (i.e. w2.www.example.com).
```
(this turned out in the end to be the problem I was encountering in #1275)

This is a request to remove the `domain` part of the cookie that FileSender sets, since using the flag means an inherent opening for other sites in the same (sub) domain to use the cookie, and leaving it empty makes the browser fall back to sending the cookie only to the exact domain of filesender, as per https://www.rfc-editor.org/rfc/rfc6265#section-4.1.2:

```
Unless the cookie's attributes indicate otherwise, the cookie is
returned only to the origin server (and not, for example, to any
subdomains), and it expires at the end of the current session (as
defined by the user agent).
```
which seems to be the more secure option here.

monkeyiq:
I was moving this way with https://github.com/filesender/filesender/pull/1248 which introduces the cookie_domain setting with a default of ```''``` . As cookie_domain was introduced in that PR it should present the default value. I made it a configuration setting in case FileSender was used as part of a larger offering and a admin might like to set it. Though the chance of that is probably slim.

monkeyiq:
I am happy to merge https://github.com/filesender/filesender/pull/1277 and remove the docs for the cookie_domain setting and see if that breaks for anyone.

WebSpider:
Since once of the pillars of the project is to be secure, I prefer removing the domain in 3.0
If it must be configurable, I think breaking the default setting in 3.0 to be `''` instead of the current setting in 2.x is the best solution.

monkeyiq:
[...]

PR #1277, opened 2022-06-30
Remove domain flag from cookie

WebSpider:
This removes the domain flag from the cookie to be compatible with [draft-west-cookie-prefixes](https://datatracker.ietf.org/doc/html/draft-west-cookie-prefixes)

As further explained in #1276, compatibility with this draft, which is implemented by major browsers, should improve protection of the session cookie we use


WebSpider:
Duplicate of #1248

PR #1279, opened 2022-07-06
restore docs for session_cookie_path

monkeyiq:
As mentioned in https://github.com/filesender/filesender/pull/1262 I am bringing back the docs for session_cookie_path. I am happy to change the docs if a better text is offered. This config value is used at https://github.com/filesender/filesender/blob/development/includes/init/init_web.php#L45


monkeyiq:
no needed

Issue #578, opened 2019-06-14
Clean URLs


peter-:
Filesender should support them. ;)

This was [mentioned on the dev list earlier](https://sympa.uninett.no/lists/filesender.org/arc/filesender-dev/2018-10/msg00010.html) but also relates to a few issues here, e.g. #98, #137, #444 and #577. Clean URLs would avoid whole categories of issues (such as the first few mentioned) due to the complete lack of "unusual" characters in URLs/emails/chat messages.

It might help to have an issue here that's clearly and only about clean URLs.

PR #1335, opened 2022-11-23
Add a way to use nicer URLs

yent:
Add a configurable way to use nicer, path based, URLs for the main features (upload, downloads, guests) as requested in #578, mainly to avoid URL transmission issues due to bad/weird email/IM clients behavior.

This requires some URL rewriting on the server side, configuration examples are provided for apache (tested) and nginx (should work, don't have a test platform for it though).

This is done through Utilities::http_build_query, had to enforce its use here and there to get (more)  consistent results.

Had to tweak some path builders in js because they depended on the current path, which is now context-dependent.


monkeyiq:
Looks good. This needs to be made against development to suit the FileSender board's method of updates. I don't see that I can change the target branch from here :/

I have a few very little updates to make. For example the prev button on the transfers table does not work as the offset is set too high. As these are very small I'm happy to make the updates in a small PR after meringing this into development.

templates/transfers_table.php
```diff
@@ -123,12 +98,13 @@ function clickableHeader($displayName,$trsortcol,$trsort,$nosort) {
...
            $prevPage = max(0,$offset-$limit);
-           echo "<td class='pageprev0'><a href='$base&$cgioffset=0&$cgilimit=$limit&transfersort=$transfersort&as=$cgias$cgiuid$cgiminmax'><span class='fa-stack fa-lg'><i class='fa fa-square fa-stack-2x'></i><i class='fa fa-angle-double-left fa-stack-1x fa-inverse'></i></span></a></td>";
-           echo "<td class='pageprev'><a href='$base&$cgioffset=$prevPage&$cgilimit=$limit&transfersort=$transfersort&as=$cgias$cgiuid$cgiminmax'><span class='fa-stack fa-lg'><i class='fa fa-square fa-stack-2x'></i><i class='fa fa-angle-left fa-stack-1x fa-inverse'></i></span></a></td>";
+           echo "<td class='pageprev0'><a href='".transferHref([$pagerprefix.'offset' => 0, $pagerprefix.'limit' => $limit])."'><span class='fa-stack fa-lg'><i class='fa fa-square fa-stack-2x'></i><i class='fa fa-angle-double-left fa-stack-1x fa-inverse'></i></span></a></td>";
+           echo "<td class='pageprev'><a href='".transferHref([$pagerprefix.'offset' => $prevPage + $limit, $pagerprefix.'limit' => $limit])."'><span class='fa-stack fa-lg'><i class='fa fa-square fa-stack-2x'></i><i class='fa fa-angle-left fa-stack-1x fa-inverse'></i></span></a></td>";
```
monkeyiq, line 107: Adding $limit makes the prev button ineffective...

`$pagerprefix.'offset' => $prevPage + $limit`
=>
`$pagerprefix.'offset' => $prevPage`

monkeyiq (approved):
will merge against development when updated.

PR #1339, opened 2023-01-02
Fix storable exceptions template injection

yent:
Clean potential template injections from user-provided data upon de-serializing storable exceptions


monkeyiq:
I am wondering if you have deployed this. If there were any regressions in other parts of the system.

madsi1m:
We've deployed this a few weeks ago, no issues so far.

yent:
Same here, deployed in production before I even did the pull request 😄

WebSpider:
We have a functionally equivalent fix in production since end 2022 as well

PR #1340, opened 2023-01-12
Fix possible SQL injection in admin page

madsi1m:
Sanitise email field from admin form and pass it through as a placeholder so PDO uses prepare function + improved UX and search emails in lowercase.

### DIFF
--- a/templates/admin_transfers_section.php
+++ b/templates/admin_transfers_section.php
@@ -24,8 +24,10 @@ $transfers_page = function($status) {
             $selector .= " AND id >= $idmin AND id <= $idmax ";
         }
     }
+    $placeholders=array();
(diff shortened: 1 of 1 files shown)


liedekef:
Hum ... what I see here is not a change to a possible SQL injection at all. If the only concern is to allow an email to have "%" before or after it to allow wildcard searches:

- check if it starts with a "%" and set a boolean var
- check if it ends with a "%" and set another boolean var
- sanitize the email
- re-add the beginning and/or ending "%" to the sanitized value based on the boolean vars

The way it is now coded it seems easy to abuse an admin session in place by sending a link with as get-param a very "bad" email ... While PDO prepare is difficult to hack, it is not impossible:
https://stackoverflow.com/questions/134099/are-pdo-prepared-statements-sufficient-to-prevent-sql-injection

Issue #1314, opened 2022-09-09
Show logged-on user in "My Transfers - See the transfer logs" when "user must login" and "get link instead of sending" are selected in "Upload"


dxdemetriou:
It could be usefull if it's possible instead of "Recipient Anonymous started downloading" to show the username of the logged-in user if the user is logged-on

monkeyiq:
I'll do some more validation on https://github.com/filesender/filesender/pull/1354 but it is likely to get merged soon :)

monkeyiq:
This has some privacy concerns because it exposes the downloading user's email to the person who created the transfer. The option is off by default but is now available for sites who wish to do this.

PR #1354, opened 2023-01-30
New Option: log downloads from authenticated users as recipients

monkeyiq:
This relates to https://github.com/filesender/filesender/issues/1314

This is mostly to be able to see which authenticated users on the same system have downloaded your "get a link" transfers. This is a little tricky as the logging system revolves around the non "get a link" work flow. In that way, each email that is added to a transfer is added to the recipients table for this transfer. Then when a user visits the download page their token uniquely identifies that guest for that transfer as a single tuple in the recipients table. That recipient tuple/object is then used by the system to log their download activity.

Rather than specialise the paths for both recipient and user to allow logging the authenticated user who is downloading, this PR adds a new tuple to the recipients table for the authenticated user. This effectively makes the user a recipient of the transfer when they download. They will have had to have a valid link to the transfer with a valid token in order to download. This will have had to have been shared by the transfer creator as the token is very hard to guess.

As this has privacy implications it is off by default and must be enabled with the new `log_authenticated_user_download_by_ensure_user_as_recipient` config option.

### DIFF
--- a/classes/auth/Auth.class.php
+++ b/classes/auth/Auth.class.php
@@ -450,6 +450,14 @@ class Auth
         return self::$isAdmin && !self::isGuest();
     }
 
+    /**
+     * Current user is not an admin, not remote, not guest.
+     */
+    public static function isRegularUser()
+    {
+        return !self::isAdmin() && !self::isRemote() && !self::isGuest();
+    }
+
     public static function canViewAggregateStatistics()
     {
         if (is_null(self::$canViewAggregateStats)) {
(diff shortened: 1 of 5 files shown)


monkeyiq:
This can also be handy if you select "User must login to FileSender to download file(s)" as with this option enabled you will be able to see which user logged in to download the files. Assuming that if the option is enabled then users are informed that this will happen and have already explicitly consented to it as part of the terms of system use for that site.

Issue #1319, opened 2022-09-19
Download page allow subsequent download with previous password


monkeyiq:
Consider the download page when there are multiple files. The user clicks to download the first file and is prompted for a password. The download completes (password was fine). 

When the user clicks a second file then they have to reenter the password. Perhaps if the password was correct it could be cached for the download page to allow subsequent files to be downloaded without the password prompt.

PR #1356, opened 2023-01-31
crypto-ui: when a password has worked offer it for other files on the same download page

monkeyiq:
This relates to https://github.com/filesender/filesender/issues/1319

The PR allows the successful password to be offered across the individual download buttons and the archive download button. Much nicer than forcing the poor user to retype things to download 2+ files.


adminnr1:
To be honest, this seems like a workaround, or a temporary solution.
It is possible for a user to encrypt multiple files with one password in one go.
The same should be possible when downloading the files.

Issue #1375, opened 2023-02-07
i18n the langauge chooser should use nice langauge names instead of the codes


monkeyiq:
Some are already there but converting the rest and presenting a sorted list would be nice. An open question remains as to if the language name should be written in English or a manner specific to that language.

monkeyiq:
For now the language name is in English. I have based many of those on the names from poeditor.

PR #1380, opened 2023-02-11
Update for a log entry in language fallback mode and cleaner language selector.

monkeyiq:
lang: the self reference there wasnt good. Only log the id to avoid super verbose logs
lang: Consecutive entries with the same language name are avoided in the language drop down.


monkeyiq:
This is also related to https://github.com/filesender/filesender/issues/1375

monkeyiq:
This contains a slight update to the update in https://github.com/filesender/filesender/pull/1377

PR #1396, opened 2023-02-22
Concurency and resilience changes to filesender.py

PaulLocksley:
This Pull request covers two changes:

- Allowing chunks to retry when something goes wrong on either the client or the server.
Errors on the file-sender server or a brief interruption in internet service will currently cancel the whole transfer. 
Simply retrying the request after a delay will resolve this issue the majority of the time.
It will try up to either terasender_worker_max_chunk_retries or a user flag. 

- Adding concurrent workers to increase performance dramatically.
For this we are using a thread pool to try and improve network utilization / remove downtime between chunks.
If terasender is disabled it will revert to a single threaded chunk by chunk transfer.  
The number of workers is user configurable via the --thread argument up to the terasender_worker_count defined by the server.

From testing on a fast internet connection uploading a single 1gb file has lowed from 1min 25 second average to just 22 seconds!

One main problem with this is currently all the servers arguments are being grabbed via filesender-config.js.php instead of just via the info endpoint.

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -137,7 +137,7 @@ except requests.exceptions.SSLError as exc:
     print(exc)
     print('Running with --insecure flag, ignoring warning...')
     info_response = requests.get(base_url+'/info', verify=False)
-    config_response = requests.get(base_url[-9]+'/filesender-config.js.php',verify=True)
+    config_response = requests.get(base_url[-9]+'/filesender-config.js.php',verify=False)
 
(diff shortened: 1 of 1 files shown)


madsi1m:
@monkeyiq this pull request comes from @PaulLocksley (one of our team), a number of our users have been asking for parallel uploads for some time via the client and this achieves it.

monkeyiq:
Cool. Thanks for the intro @madsi1m :)

PR #1414, opened 2023-03-28
Make default date format, and that of en & nl, less ambiguous

thijskh:
The default date format of Filesender is Australian. This means that a date will be written as 05/04/2023. Receiving an English language email with that date is very ambiguous to me since there's different interpretations between English speaking countries about which part is the month in that date. Most notably the current format is different from what is used in the US which some might assume, justifiably or not, to be the default for things.

To disambiguate it once and for all, change the default to 05 Apr 2023, which is instantly clear and requires no interpretation. The exact month abbreviation is based on the locale which is set right before interpolation.

I would leave it up to translators of the other languages to make any changed they think are appropriate.


WebSpider:
Thanks for this PR!

Please be aware that any changes to languages will be overwritten by [POEditor changes](https://poeditor.com/projects/shared?id=48000), where the translation for filesender is managed.

thijskh:
I don't see anything at that link, just the option to create a new project.

meijer:
Does this work for you? https://poeditor.com/join/project/RqXr9WBJwU

Taken from the docs, but perhaps those are not the easiest to find anymore? https://docs.filesender.org/filesender/v2.0/i18n/#how_to_contribute_to_the_filesender_2.0_translation_effort

thijskh:
Thanks, I've submitted the translation part of this PR also there.

monkeyiq:
I notice that the poeditor has been updated, thanks! I just updated datetime_format for the 'en' translation to match. I think updating the base 'en' is a good idea as it gets suggested to translators.

I tend to update date formats for personal use for the same reason. Seems like a really good idea to remove ambiguity in the date formats of FileSender by default.

Thank you for the update and those who commented on the PR :)

PR #1415, opened 2023-03-28
New reformatted filesender.py with additional options

KenRoy-Octo:


scripts/client/filesender.py
```diff
@@ -29,427 +29,429 @@
...
+_PARSER = None
+
+
+################################
+# If error, print usage and exit
+################################
+def print_usage_and_exit(err_message):
+    global _PARSER
+    print("\n" + err_message + "\n")
+    _PARSER.print_help()
```
WebSpider, line 69: This is undefined, there is no help text in your PR

monkeyiq:
I will take more of a look once the missing help text is resolved :)

KenRoy-Octo:
> I will take more of a look once the missing help text is resolved :)

_parser. print_help is a function within _parser that prints out arguments added by ArgumentParser.  It already exists.

See:  https://docs.python.org/3/library/argparse.html

scripts/client/filesender.py
KenRoy-Octo: _parser. print_help is a function within _parser that prints out arguments added by ArgumentParser. It already exists.


https://docs.python.org/3/library/argparse.html

KenRoy-Octo:
_parser.print_help() is being called before it is created.  Fixing now...

KenRoy-Octo:
Fixed.

monkeyiq:
It seems that the updates from https://github.com/filesender/filesender/pull/1396 and https://github.com/filesender/filesender/pull/1358 are removed by the PR?

peter-:
@KenRoy-Octo: Could you integrate more recent changes (see above) and update this PR here?

Issue #1411, opened 2023-03-24
Issue with guest uploads with Cloud Backend S3


flowdnb:
Hello,

We are facing an issue with v2.39 when using guest uploads while the file store is connected to an S3-compatible cloud backend (IBM Cloud Object Storage on-premises).

- Configuration we have is:
```
$config['storage_type']      = 'CloudS3';
$config['cloud_s3_endpoint'] = 'https://MyHiddenS3Endpoint.local';
$config['cloud_s3_key']      = 'MyHiddenS3Key';
$config['cloud_s3_secret']   = 'MyHiddenS3Secret';
$config['cloud_s3_bucket']   = 'MyHiddenS3Bucket';
```

- FileSender tries to create a bucket for the file uploaded by guests instead of putting it into the configured bucket (at the same time, uploads work for authenticated users).

```
exception 'Aws\S3\Exception\S3Exception' with message 'Error executing "CreateBucket" on "https://MyHiddenS3Endpoint.local/6c556b43-73a8-47dd-b086-65884b0df4b7"; AWS HTTP error: Client error: `PUT https://MyHiddenS3Endpoint.local/6c556b43-73a8-47dd-b086-65884b0df4b7` resulted in a `403 Forbidden` response:
<?xml version="1.0" encoding="UTF-8" standalone="yes"?><Error><Code>AccessDenied</Code><Message>The provided account is  (truncated...)
 AccessDenied (client): The provided account is not authorized to create vaults through the Provisioning API - <?xml version="1.0" encoding="UTF-8" standalone="yes"?><Error><Code>AccessDenied</Code><Message>The provided account is not authorized to create vaults through the Provisioning API</Message><Resource>/6c556b43-73a8-47dd-b086-65884b0df4b7/</Resource><RequestId>669a5787-685d-441c-be2b-e2afcce3707c</RequestId><httpStatusCode>403</httpStatusCode></Error>'

GuzzleHttp\Exception\ClientException: Client error: `PUT https://MyHiddenS3Endpoint.local/6c556b43-73a8-47dd-b086-65884b0df4b7` resulted in a `403 Forbidden` response:
<?xml version="1.0" encoding="UTF-8" standalone="yes"?><Error><Code>AccessDenied</Code><Message>The provided account is  (truncated...)
 in /opt/filesender/filesender/optional-dependencies/s3/vendor/guzzlehttp/guzzle/src/Exception/RequestException.php:113
Stack trace:
[...]

PR #1416, opened 2023-04-03
fix use custom S3 bucket for Guest Uploads

flowdnb:
Hello, I have created a fix for my issue #1411 and would like to ask for a review.


monkeyiq:
I can see where in `RestEndpointTransfer.class.php` this gets carried to set transfer options. Looks good.

PR #1419, opened 2023-04-08
Return to original page upon session timeout

makinog3:
When a session times out while a non-upload page is open and I log back on, it takes me to the upload page instead of the originally opened page.


monkeyiq:
This seems to be a larger patch than was intended?

WebSpider:
I think it was meant to be an issue instead of a PR

makinog3:
Sorry, I had a little trouble while checking.

makinog3:
I can’t fix the trouble soon, and  I'll close.

PR #1424, opened 2023-04-11
Return to original page upon session timeout

makinog3:
When a session times out while a non-upload page is open and I log back on, it takes me to the upload page instead of the originally opened page.


makinog3:
Thank you for merging!

PR #1425, opened 2023-04-11
Display page navigation on Admin_Transfers page

makinog3:
Display page navigation depending on the number of transfers on Admin_Transfers

* Cannot confirm all transfers on Admin_Transfers page if the transfers are over 8 page.

### DIFF
--- a/templates/admin_transfers_section.php
+++ b/templates/admin_transfers_section.php
@@ -4,6 +4,7 @@ $auditlogs = Config::get('auditlog_lifetime') > 0;
 
 $transfers_page = function($status) {
     $page_size = 15;
+    $display_page_num = 4;
 
     switch($status) {
         case 'available': $selector = Transfer::AVAILABLE_NO_ORDER; break;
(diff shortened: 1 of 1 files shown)


templates/admin_transfers_section.php
```diff
@@ -97,6 +102,14 @@
     for($o=0; $o<$total_count; $o+=$page_size) {
```
monkeyiq, line 102: total_count was limited before. If this is a large installation it might become quite large and so the for loop will be over many, many tuples. I have local update soon to be a PR to work around this by starting and ending at more specific indexes so that only `$page_size * $display_page_num` before/after should be considered in the loop.

templates/admin_transfers_section.php
```diff
@@ -66,7 +67,11 @@
     // FIXME: move the code away from wanting to know the total.
     //       if the user has 1000 tuples do we really want to show 1000/15 direct page links
     //       or should we instead allow queries on timeframe etc.
-    $total_count = 100;
+    $total_count = Transfer::count(array(
```
monkeyiq, line 70: I would also like to short this out if `$offset` is zero and there are no 'real' conditions in the where clause. There are three tables shown by default and calculating the exact total_count for those cases might be quite expensive on a very large database. Especially since one condition starts simply as status='available' which will likely be quite a number of tuples. I have an update to do this so admins with large databases might see the page right away at the expense of maybe seeing page 2, 3 etc for closed tables.

PR #1428, opened 2023-04-11
Ignore error if a bounced email with broken Content-type field

makinog3:
When task/emailfeedback.php receives a bounced email with broken Content-type field, it fails to the process. For example, the end of Content-type field is ";".

```
--------------I305M09060309060P_297316237437599
Content-Type: text/plain; charset=UTF-8;
Content-Transfer-Encoding: 8bit
```
=> PHP Notice:  Undefined offset: 1 in /opt/filesender/filesender-2.21/classes/utils/FeedbackMail.class.php on line 151


monkeyiq:
Cool, thanks! I see that if there is a trailing space after the `;` then the code could still crumble in a slightly different way. I'll add a PR for that possibly odd and rare case next :)

makinog3:
Thank you for your review!

PR #1430, opened 2023-04-11
Zip archive for file names containing multibyte characters

makinog3:
When generating a ZIP archive on the client side(JS), file names containing multibyte characters are incorrectly handled.

### DIFF
--- a/www/js/crypter/crypto_common.js
+++ b/www/js/crypter/crypto_common.js
@@ -33,12 +33,7 @@ window.filesender.crypto_common = function () {
         },
         convertStringToArrayBufferView: function (str)
         {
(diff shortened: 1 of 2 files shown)


monkeyiq:
I have started testing this.. on Linux unzip does now make the correct extracted files.  I spent some time trying to remove the following warning to no avail. The files extract ok though. This warning might be a future thing to dig into.

```
multibyte-Á-filename.txt:  mismatching "local" filename (multibyte-├Б-filename.txt),
         continuing with "central" filename version
```

I will try to get some testing on osx and windows next.

monkeyiq:
Thank you for the PR!

PR #1432, opened 2023-04-15
email handling: improved handling for content-type with errors in it

monkeyiq:
This allows erroneous spaces at the end of the content-type to be handled.


monkeyiq:
This builds on the check in https://github.com/filesender/filesender/pull/1428

PR #1433, opened 2023-04-15
admin transfers page: db paging update

monkeyiq:
This is an update to https://github.com/filesender/filesender/pull/1425 

Thanks @makinog3 for the above PR which inspired me to tinker on these updates!

It allows a fake `$total_count` if the offset is zero and there are no real selectors in the SQL query. This is the case for when you first view the admin / transfers page and the totals are likely to be large, somewhat uninteresting, and maybe slow (seq scan on the transfers table).

I have also limited the scope of the `for` clause to only consider pages near the current offset rather than from 0 to  `$total_count`.

### DIFF
--- a/templates/admin_transfers_section.php
+++ b/templates/admin_transfers_section.php
@@ -67,11 +67,20 @@ $transfers_page = function($status) {
     // FIXME: move the code away from wanting to know the total.
     //       if the user has 1000 tuples do we really want to show 1000/15 direct page links
     //       or should we instead allow queries on timeframe etc.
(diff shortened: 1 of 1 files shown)


monkeyiq:
On further thought the `$total_count` might not be needed. That query really just wants to know if there are entries before and after the `$offset`. Something like the following but with bounding for ensuring that offset does not go negative. Seems like a good idea for a subsequent PR at some stage.

```
        'count'  => $page_size * $display_page_num,
        'offset' => $offset - $page_size * $display_page_num/2,
```

Issue #1406, opened 2023-03-02
Feature Request - Per domain personalisation


charcoalbe:
Add possibility for personalisation (intro text / logo / custom css (?) ) per domain. 
Possibility to also link guest-uploads to domain of requester (and have personalisation work there)

liedekef:
If the text/logo/css-url were to be options, we could use the new feature  auth_config_regex_files already for this.
So I think this could be a request to be able to have a config option for intro text/logo/css-url .
Btw: a left/right logo would be even more interesting (so people wanting to have their logo per domain could still also have the logo of the organisation running filesender visible).
While I think about this, I guess using a different theme per customer is already possible, but personally I also think the text/logo/css could merit a config option on their own :-)

WebSpider:
Closing this issue, as the remaining TODO items will be adressed in #1547

PR #1437, opened 2023-04-18
allow css/logo to be set in config (maybe using auth_config_regex_files)

monkeyiq:
This relates to https://github.com/filesender/filesender/issues/1406

This allows setting and additional css file (and top logo) using config settings. Using auth_config_regex_files allows that config setting to be performed differently for different sites.


WebSpider:
Really looking forward to this, thanks!

PR #1450, opened 2023-04-25
Pull req: Tiny fixes related to Filesender cloud backend support

juhassi:
Quick fixes related to Filesender's cloud backend support.

$written is not returned by the cloud storage functions when we're saving data. PHP is logging warnings of this in the error log. I'm assuming $chunk_size can be used instead of $written.

(edit 26.04.2023) Also additional bugfix: When using StorageCloudS3 backend with custom bucket name (all files go to the same bucket instead of creating a new bucket for every file), deleting one file would empty out the entire bucket.


juhassi:
Also additional bugfix: When using StorageCloudS3 backend with custom bucket name (all files go to the same bucke instead of single-bucket-per-file), deleting one file would empty out the entire bucket.

PR #1466, opened 2023-05-10
Allow filenames contain special characters in Unicode

makinog3:
* add '+' in ASCII
    * add special character areas
      ex) MIDDLE DOT : U+30FB
    * convert from PHP to JavaScript for surrogate pairs
      ex) CJK UNIFIED IDEOGRAPHS : U+29E3D, Emoji : U+1F637

### DIFF
--- a/includes/ConfigDefaults.php
+++ b/includes/ConfigDefaults.php
@@ -187,7 +187,7 @@ $default = array(
     'report_format' => ReportFormats::INLINE,
 
     // Note that this must not have a fixed end of string '$' as the last character in the match 
-    'valid_filename_regex' => '^[ \\/\\p{L}\\p{N}_\\.,;:!@#$%^&*)(\\]\\[_-]+',
+    'valid_filename_regex' => '^['."\u{2010}-\u{2027}\u{2030}-\u{205F}\u{2070}-\u{FFEF}\u{10000}-\u{10FFFF}".' \\/\\p{L}\\p{N}_\\.,;:!@#$%^&*+)(\\]\\[_-]+$',
     'message_can_not_contain_urls_regex' => '',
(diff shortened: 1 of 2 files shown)


WebSpider:
IIRC, in some configurations filenames are used in the storage on the server filesystem. 
Will this also work if the filesystem doesnt properly support Unicode files?

monkeyiq:
The `$config['storage_type'] = 'filesystem'` will store the file contents in a single file using a uuid as the file name. The filesystemChunked will make a directory with that file uuid as its name and use several files with offset and chunk size in that directory. A quick look at the cloud storage it seems that the uid and offsets are the main locators there too.

monkeyiq:
I am not sure what sort of reason people would do it for but  U+2044 fraction slash might be confused with a `/` and U+205F is a math space might be confused with a regular space. I guess the open question is if we want the default to be accepting wider ranges or if we should suggest this as a config.php option for sites.

makinog3:
Thank you for your comments.
So, in my PRs, the default setting of `valid_filename_regex` in `ConfigDefaults.php` should be dropped and defined in `config.php` to adapt to individual system requirements.
Another one, a fix in `filesender-config.js.php` will correctly convert `valid_filename_regex` when it contains surrogate pairs, and I would like to take the fix.
How about do you think?

madsi1m:
You would be better off setting it in your config.php
Not 100% sure you need to strreplace in the .js.php file

monkeyiq:
I split out the suggestion for updating valid_filename_regex into https://github.com/filesender/filesender/pull/1476 and the surrogate pairs update into https://github.com/filesender/filesender/pull/1477 and merged them. This should cover the suggestions made in this PR.

Please let me know if there are other translation updates to make :)

Issue #1325, opened 2022-09-26
Feature request: verify e-mail-address on download


baumrasen:
It would be nice to have an option on upload, that the downloader should enter an additional verification code. This should be send on download-request to the e-mail-address, which is associated with the download-token.

The expiration time for this code should be set in the config. Default 5 min for example.

There should be the possibility to send a new verification code.

There could be an option to "keep the downloader signed in" (by cookie, which is associated with the a-mail-address of the downloader)

WebSpider:
This has been added in 2.41

PR #1480, opened 2023-05-21
new option for email verification code

monkeyiq:
This allows for email verification codes as suggested in https://github.com/filesender/filesender/issues/1325. When enabled a user must verify that they have access to an email account before they are allowed to download a file. When enabled, this option is selectable on the upload page. The option is not available for "get a link" uploads as they do no have email address associated with them.

There is still some more validation to do before I merge this.


language/en_AU/transfer_email_verify_to_download.mail.php
```diff
@@ -0,0 +1,41 @@
+<?php 
+// WARNING, this is a read only file created by import scripts
+// WARNING
+// WARNING,  Changes made to this file will be clobbered
+// WARNING
+// WARNING,  Please make changes on poeditor instead of here
+// 
+// 
+?>
+subject: FileSender: One time verification code for download
```
WebSpider, line 10: Possibly better to use the cfg:site_name here instead of static string FileSender

monkeyiq: Cool. I did a few passes to catch if I still had statics in the html/php but hadn't really worked on the email text too much. Updated.

PR #1497, opened 2023-06-15
Fix parameter of SystemMail::quickSend()

makinog3:
Fatal error at `scripts/task/cron.php` when calling `SystemMail::quickSend()` because the parameter is different from that of the inherited `ApplicationMail::quickSend()`

```
PHP Fatal error:  Declaration of SystemMail::quickSend($translation_id) must be compatible with ApplicationMail::quickSend($translation_id, $to) in /opt/filesender/filesender-2.39/classes/utils/SystemMail.class.php on line 65
```

### DIFF
--- a/classes/utils/SystemMail.class.php
+++ b/classes/utils/SystemMail.class.php
@@ -62,7 +62,7 @@ class SystemMail extends ApplicationMail
      * @param string $translation_id
      * @param mixed ... additional translation variables
      */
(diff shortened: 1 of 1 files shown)


monkeyiq:
It is probably a good idea to update the method comment for SystemMail::quickSend() to mention that $to is not needed or used and is only there are a synthetic place holder and the ... "additional translation variables" are taken from $to onwards.

PR #1505, opened 2023-07-04
drag and drop files eval the Send button

madsi1m:
When drag and drop files the Send button doesn't enable, this change fixes this issue. I am not sure if it breaks other things

### DIFF
--- a/www/js/upload_page.js
+++ b/www/js/upload_page.js
@@ -223,7 +223,7 @@ filesender.ui.files = {
                 file_name = files[i].webkitRelativePath;
             }
             
-            var latest_node = filesender.ui.files.addFile(file_name, files[i], false, source_node);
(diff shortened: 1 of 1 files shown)


monkeyiq:
I made https://github.com/filesender/filesender/pull/1490 which makes the send button run again. I'm not 100% happy with it as yet but works in quick testing.

monkeyiq:
The isSingleOperation was added in https://github.com/filesender/filesender/pull/1490 to allow huge file lists to be added a little bit quicker. Some folks have been adding directories which result in 1000s of files being listed. Somehow testing D&D slipped through that gaps when testing out the 1490 PR code.

monkeyiq:
I merged in 1507. Thank you for reporting the issue and also going the extra mile of making a PR.

PR #1508, opened 2023-07-06
i18n: import Thai from poeditor

monkeyiq:


monkeyiq:
This is also added to the regular import script

PR #1518, opened 2023-07-26
Update ui.js popup buttons text

madsi1m:
Update ui.js popup buttons text, strip out html tags from lang.tr() on button text


madsi1m:
Idea is to avoid things like this:
image

PR #1525, opened 2023-08-08
Update guzzlehttp/psr7 and libs around it (CVE-2023-29197)

madsi1m:
Attempt to update guzzlehttp/psr7 for CVE-2023-29197.


monkeyiq:
For others reading this, the version in lib/vendor/guzzlehttp is only used by the AVProgram function in FileSender and 
even then only if you setup a remote (REST) avprogram. In that case FileSender  makes a http call during AV runs on files. In this case both the client (the cron job) and server (your AV program running as a REST service to inspect a  file) are fully under the control of the installation.

```
$ grep -Ri guzzle classes/
classes/avprograms/AVProgramURL.class.php:use GuzzleHttp\Client;
classes/avprograms/AVProgramURL.class.php:use GuzzleHttp\Psr7\Request;
```

PR #1527, opened 2023-08-08
Added Albania to list of known-installs.md

klention:


monkeyiq:
Thank you for the update!

PR #1532, opened 2023-08-18
fix for php8

liedekef:
not sure if "message" as string value is correct, but it needs to be an array, otherwise you get this php 8 error : 

`2023/08/18 16:06:03 [error] 8771#8771: *131 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught TypeError: in_array(): Argument #2 ($haystack) must be of type array, string given in /opt/filesender-3.0.beta2/templates/upload_page.php:501
`

### DIFF
--- a/templates/upload_page.php
+++ b/templates/upload_page.php
@@ -576,7 +576,7 @@ if(Auth::isGuest()) {
                         if( $name == 'add_me_to_recipients' ) {
                             $upload_options_handled[$name] = 1;
                             $forcedOption = false;
-                            $displayoption($name, $cfg, Auth::isGuest(),$forcedOption,"message");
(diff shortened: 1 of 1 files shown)


monkeyiq:
Thanks!

monkeyiq:
Inline with SURF & the filesender boards requirements this should go into development3 to later be rerolled in bulk into master3.

monkeyiq:
Changing the base to dev3 seemed to produce a non exciting result. Likely because dev3 is branched from development a while back and master3 is not directly related in git.

monkeyiq:
To save a bit of hassle I created https://github.com/filesender/filesender/pull/1534 against development3 and merged that.

PR #1544, opened 2023-09-15
Add option for download_show_download_links

madsi1m:
Add new option download_show_download_links where default is false. This allows FileSender to show direct download links on the download page (where possible)

How it looks with default filesender settings:
image

How it looks with download_show_download_links set to true:
image


monkeyiq:
Thanks for the PR! 

I think having the option for direct links is very useful when people are using download managers or for more advanced users who are looking to copy and paste links around. Perhaps they are looking to use wget/curl from the command line on another machine to download to a remote site.

Issue #1549, opened 2023-10-03
feature request: filesender cli client to accept folders and maintain the file structure?


madsi1m:
Hi team, is it possible to add a feature where filesender cli client to accept folders and maintain the file structure?
Cheers

monkeyiq:
I like this idea.

WebSpider:
Yeah I do as well. 
Looking at #1553 it seems that more people are after feature improvements of the CLI client

madsi1m:
Hi here are some improvements on @PaulLocksley's merge: https://github.com/filesender/filesender/pull/1556 hope you all don't mind

PR #1554, opened 2023-10-06
Allow Filesender.py to preserve directory structure

PaulLocksley:
In response to https://github.com/filesender/filesender/issues/1549
This allows the user to specify the -d / --directories flag to automatically try and preserve folder structure. 

I opted for the approach of automatically working out what folders to keep based on a common root for all files. 
We use filesenders automatic directory insertion to create the files, because of this the names returned in transfer aren't actually usable to identify the file locations, to get around this I keep track of the file name we gave filesender and the on disk location. 

I've tested this on windows with 1000 files and several subfolders deep and it worked well.

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -99,7 +99,7 @@ parser.add_argument("-i", "--insecure", action="store_true")
 parser.add_argument("-p", "--progress", action="store_true")
 parser.add_argument("-s", "--subject")
 parser.add_argument("-m", "--message")
-parser.add_argument("-d", "--subfolder", action="store_true")#name.
+parser.add_argument("-d", "--directories", action="store_true")
 parser.add_argument("-g", "--guest", action="store_true")
 parser.add_argument("--threads")
 parser.add_argument("--timeout")
@@ -133,7 +133,7 @@ guest = args.guest
 user_threads = args.threads
 user_timeout = args.timeout
 user_retries = args.retries
-preserve_directories = args.subfolder
+preserve_directories = args.directories
 
 if args.username is not None:
   username = args.username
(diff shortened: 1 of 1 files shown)


monkeyiq:
Thank you for the PR and details on the high level of testing you have applied. I will look at this today/tomorrow.

monkeyiq:
I initially tested this thinking the `-d` in filesender.py was going to work like `-R` in cp. 
```
./filesender.py  -v   -r tester@localhost -d /tmp/test
```

which results in 
```
Root path identified : /tmp/
putChunks: /tmp/test
<class 'IsADirectoryError'>
(21, 'Is a directory')
[Errno 21] Is a directory: '/tmp/test'
deleteTransfer
```

Once I read the patch I got uploads going....

```
 ./filesender.py -v   -r tester@localhost -d /tmp/test/df1.txt /tmp/test/sub1/sub1file.txt
```

 I think maybe most users will just want to upload an entire directory tree  rather than listing the files explicitly? The proposed `-d` is very useful if you want to pick explicit files though... I am happy do do that though it seems having and recommending a `-R` to compliment this `-d` might be useful. Perhaps even asking if the sole path given is a directory if the user wants to switch on `-R` to upload the whole tree.

PaulLocksley:
Thanks, I agree with the -R flag, I noticed thought that would be useful, I have a [branch](https://github.com/filesender/filesender/commit/37028786e5c3b0af3ede35651aaf780362f4c43f) where I was testing it on Friday. 

I may clean that up and send it through later this week.

madsi1m:
I also think a -R type flag would be super useful. Keep up the good work :)

PR #1556, opened 2023-10-09
FileSender Python Client: recursive dir 

madsi1m:
Here is my improvements to @PaulLocksley Python Client:

- removed -d
- able to just give it folder(s)/file(s) in args and python works it out
- sort file list
- include filesender max_transfer_files, so we don't spam filesender

should solve https://github.com/filesender/filesender/issues/1549


monkeyiq:
Thank you for the update.

PR #1566, opened 2023-10-13
Update known-installs.md

monkeyiq:


monkeyiq:
This was merged in https://github.com/filesender/filesender/pull/1565

The downside to the edit button on github I used to make this is that it makes branches in the main repo.

Issue #1555, opened 2023-10-09
API for listing file IDs using only a download token


multimeric:
I'm trying to automate the download process using the API. The issue is that, if you have a download URL such as `/?s=download&token=2d335494-b624-5942-a152-a5579eeccff2`, there is no way to programatically list all of the files that are available. 

Once you have the list of file IDs, you can easily use `/download.php?token=2d335494-b624-5942-a152-a5579eeccff2&files_ids=ABC,DEF,GHI` to download the files programmatically.

Therefore, it would be great if there were a REST API that did this. The authentication should be similar to `download.php` in that it doesn't require authentication but does require a download token. The endpoint would simply return a JSON list of file IDs.

monkeyiq:
This is almost there. At the moment if you go a GET for a transfer id `$tid` on
```
https://localhost/filesender/rest.php/transfer/$tid
```
then you can see all the IDs. It is probable that a bridge to get the transfer id $tid from the token is needed. Or that REST api expanded to allow $token -> file id list.

The below pull request allows the fileids to be obtained making a REST GET with `.../rest.php/transfer/fileids?token=2d335494-b624-5942-a152-a5579eeccff2`

https://github.com/filesender/filesender/pull/1569

multimeric:
Yeah that was part of the issue. As a download user with a download link, all you have is a download token, but this doesn't authenticate you to list transfers and therefore get the file IDs. The *sender* of the files will have no problem, but the recipient will.

multimeric:
Just noting here that it would also be nice to get the full file name from this API, because currently there's no way to recreate the full filename with its directory names just via the `/download.php` endpoint.

madsi1m:
@multimeric, perhaps you would like to create a pull request for consideration?

monkeyiq:
I think this is there but remains behind a wall where you have to be user authenticated. The user authentication is not really needed if you have the token as that will mirror the security of the web interface. I think perhaps the user authentication was put in there to err on the side of caution. But this access control discussion is secondary to the information offered.

In particular this call will give all metadata about a file including it's name. If that name does not include the path then I am happy to include that in this API call as well.
https://github.com/filesender/filesender/blob/f2a2f9570e9129d600739c39cdc7eb256ea6020c/classes/rest/endpoints/RestEndpointTransfer.class.php#L202C1-L202C15

multimeric:
[...]

PR #1569, opened 2023-10-20
rest allow the fileids to get obtained if you know the token and are authenticated

monkeyiq:
This relates to https://github.com/filesender/filesender/issues/1555.

### DIFF
--- a/classes/rest/endpoints/RestEndpointTransfer.class.php
+++ b/classes/rest/endpoints/RestEndpointTransfer.class.php
@@ -148,7 +148,6 @@ class RestEndpointTransfer extends RestEndpoint
      */
     public function get($id = null, $property = null, $property_id = null, $filtertype = null, $filterid = null )
     {
-
         // Special case when checking if enable_recipient_email_download_complete option is enabled for a specific transfer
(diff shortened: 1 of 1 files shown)


multimeric:
Can you explain how to use this? A recipient has only a token, but if I hit this new endpoint, it says I need to authenticate

```
$ curl 'https://filesender.aarnet.edu.au/rest.php/transfer/fileids?token=my-download-uuid-token'                                                                                                             {"message":"rest_authentication_required","uid":"6683a5db7c2e3","details":{"method":"get","endpoint":"transfer"}}
```

PaulLocksley:
Hey @multimeric, 
I was having a play with this earlier today, this endpoint still requires authentication but you _should_ have no issues getting files from any valid download token.

Reading the related file changes, this seems like the desired behavior.

If you sign the request like normal do you have any issues?

multimeric:
> this endpoint still requires authentication

Yes but the person who has been sent a transfer potentially has no credentials at all. They are neither a FileSender user, not a guest (which is only for voucher uploads). So possibly my problem is still outstanding?

Issue #1607, opened 2023-11-02
UI3 My invitations page: options not saved


monkeyiq:
(inv4) 

selecting all the options results in a guests tuple with
     options = {"email_guest_created":true,"email_guest_created_receipt":true,"email_guest_expired":true}
       and 
     transfer_options = {"get_a_link":false}

     Should be for example:
      {"email_upload_started":true,"email_upload_page_access":true,
       "valid_only_one_time":false,"does_not_expire":true,"can_only_send_to_me":true,
       "email_guest_created":true,"email_guest_created_receipt":true,
       "email_guest_expired":true}
           and
      {"email_me_copies":true,"email_me_on_expire":true,"email_upload_complete":true,
       "email_download_complete":true,"email_daily_statistics":true,
       "email_report_on_closing":true,"enable_recipient_email_download_complete":true,
       "add_me_to_recipients":true,"get_a_link":false,"must_be_logged_in_to_download":true,
       "web_notification_when_upload_is_complete":true,"hide_sender_email":true}

monkeyiq:
This needs more verification with pr1658. I might have put some things in the wrong guest or transfer slot but at least there is something being saved now.

WebSpider:
This seems to work fine on latest 3.0 beta

PR #1658, opened 2023-11-05
Allow saving new guest options again

monkeyiq:
dev3: As long as new_invitation_page.js remains looking for these by the parent class it is important for the template to keep putting that expected format in place. Either change both things, neither of them, but certainly not just one of the two.


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/1607

Issue #1638, opened 2023-11-02
UI3: Code details - Guest table with open downloads


WebSpider:
templates/guests_table.php
(code15) if there are downloads for a guest then the table is broken trying to read
    Attempt to read property "downloads" on array

monkeyiq:
I guess too the "downloads" is strange is the page is clearly showing how many uploads a guest has done.

PR #1659, opened 2023-11-05
Allow the page to render when a guest has performed some uploads

monkeyiq:
That collection is of transfers not a single transfer object. I dont recall this one being in the review itself. Just an organic one that prevented the page from rendering.


monkeyiq:
I guess it is a good thing that the review had found this one. https://github.com/filesender/filesender/issues/1638

PR #1681, opened 2023-11-08
Download flow consolidation

yent:
Upon first access to a download link a transaction identifier is generated (using filesender's uuid generator) and user is redirected to the current download along with the identifier.

AuditLogs get an additional transaction identifier property that automatically logs any (valid) transaction identifier contained in the request.

This makes it so subsequent download events (resumes, ends) are flagged as part of the same transaction in the auditlogs, and thus makes it possible to identify bogus clients and compute more reliable statistics.


monkeyiq:
Nice. I will probably follow up with a PR to add a secondary index on the `transaction_id` as that should make grouping queries go a bit better.

monkeyiq:
As per the FileSender board's decision pull requests must be made against `development` for the 2.x series. The master branch evolves as a series of discrete patches taken between master and development at the time a release is being made. While the code is similar this does make the two branches have completely different histories and thus one can not easily redirect a PR from master to development on the github UI. The easiest resolution is to manually make a patch to master, checkout development, branch from development, apply the patch and rename a new PR.

monkeyiq:
I am also happy to do the code bounce to a new PR & merge it.

yent:
Redid the PR on filesender:developement : #1699

PR #1690, opened 2023-11-11
UI3 guest handling update: email shown, no get a link

monkeyiq:
The full email fields are shown for a guest upload. They are removed for guests who can only send to the inviting party. That needs retesting.

For now I have disabled the ability for a guest to perform an upload and "get a link" to the file. It is a code path that needs revalidation to allow a guest to upload a file and get a link to it instead of sending emails etc.


monkeyiq:
This relates to https://github.com/filesender/filesender/pull/1690

Issue #1720, opened 2023-12-01
3.0b5: Allow configurable transfer expiry settings


WebSpider:
Ever since #1709 the minimal transfer validity is 45 days.
For some installs this almost doubles the currently configured maximum, without the possibility to lower this, without causing FileSender to report an unrecoverable error.

Please make this more flexibily configurable (both UI and back-end), so moving to FS3.x doesn't cause the application to grow in storage footprint

monkeyiq:
It might make more sense to remove items from the fixed choices in the drop down if they are not valid for the configured minimal transfer validity.

PR #1727, opened 2023-12-12
d3ui3 go back to default max_transfer_days_valid and respecting it

monkeyiq:
This is an update to
https://github.com/filesender/filesender/pull/1709 in that the config setting is what dictates what the UI offers rather than the other way around.

This was reported in
https://github.com/filesender/filesender/issues/1720 and should make migration to 3.x simpler for sys admins.


WebSpider:
Thanks! 🎉

Issue #1733, opened 2023-12-22
users cannot download multiple files as a single zip/tar file in filesender 2.43


StCyr:
It seems that you ran composer on php-8 and that you've dropped the following in filesender 2.43:lib/vendor/composer/platform_check.php:

if (!(PHP_VERSION_ID >= 80100)) {
    $issues[] = 'Your Composer dependencies require a PHP version ">= 8.1.0". You are running ' . PHP_VERSION . '.';
}

for some reason it prevents users from downloading multiple files as a single zip/tar file

monkeyiq:
It seems that the first issue can be gotten around with a blunt axe using the following. 
```
 composer update  --ignore-platform-req=php
```

As there are two fairly distinct paths in the code for downloading multiple files as a single zip/tar I assume we are talking about the case here for not encrypted files. Archives for not encrypted files are generated on the server and will likely be effected by this php 8.x issue if the code has troubles running on 7.x.

I have made a branch with the php version check ignored at https://github.com/monkeyiq/filesender/tree/2023/dec/php8-server-archives I have not tested that on a php7 system. I am wondering if that branch solves the archive problem in your environment?

If not if there are any messages in `/opt/filesender/log` that might shine some light on a next move.

madsi1m:
I don't think it is a php version bug as i see it in php 8.3, above MR fixes it for me.

StCyr:
> composer update  --ignore-platform-req=php

we don't have composer installed on our filesender server

>I don't think it is a php version bug as i see it in php 8.3, above MR https://github.com/filesender/filesender/pull/1735 fixes it for me.

I don't think your PR solves this particular issue: Our filesender server works perfectly well with php7 after we've replaced the lib/vendor/composer/platform_check.php file with the one from version 2.42

The problem of this particular issue is that some part of your code say that they require php8 while your documentation doesn't say so. 

The solutions are either (a) filesender requires php8 and you should update your documentation accordingly, or (b) filesender doesn't require php7 and you should fix your lib/vendor/composer/platform_check.php file

madsi1m:
Looking at that code it seems it was added when composer updated guzzle. Looking at guzzle lib it says it needs php: ^7.2.5 || ^8.0 so i guess in theory we could remove that if statement >= 70250 or revert it back to 70300 ?

monkeyiq:
>> composer update  --ignore-platform-req=php

> we don't have composer installed on our filesender server

[...]

PR #1735, opened 2023-12-27
seems this should be REQUEST not a GET

madsi1m:
Fixes: https://github.com/filesender/filesender/issues/1733
Please double check!

### DIFF
--- a/www/download.php
+++ b/www/download.php
@@ -134,12 +134,12 @@ try {
     
     // Ensure transaction id
     $transaction_id = '';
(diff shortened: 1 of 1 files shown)


monkeyiq:
This makes sense. Non encrypted files for archives use POST for a little while in order to avoid length limitations in GET when you have 1000s of explicitly selected fileids.

monkeyiq:
This feels like something that makes a 2.44 release a good idea. I will probably do one within the next week if not sooner.

PR #1775, opened 2024-01-15
S3: Add bulk delete option

WebSpider:
This adds bulk delete for S3 storage to filesender.

Bulk delete allows deletion of multiple chunks simultaneously, up to a configured maximum. The purpose of this feature is to increase speed while deleting chunks from S3, due to lower overhead on the S3 side.
In tests, with default values, the deletion speed of a file with an exact multiple 1000 chunks was 8x faster than normal.

Bulk delete is controlled by two configurables:

The feature is toggled by boolean `cloud_s3_bulk_delete`, which defaults to false, for backward compatibility
There is one option, called `cloud_s3_bulk_size`, which defaults 1000, and is the number of objects that is deleted in one request.

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -797,6 +797,7 @@ class Transfer extends DBObject
             // In case we keep audit data for some time only delete actual file data in storage
             foreach ($this->files as $file) {
                 try {
+		    Logger::debug('Attempt to call Storage::deleteFile for ' . $file);
                     Storage::deleteFile($file);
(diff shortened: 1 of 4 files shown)


monkeyiq:
Cool. ConfigDefaults, docs updates, and a PR from somebody who is wanting the feature! Thank you.

efef:
hi @WebSpider I notice this feature has been added to V2, will it also be added in the V3 release? Do you have an estimate when it will become available? thanks!

PR #1776, opened 2024-01-16
Fail to parse "To:" header at emailfeedback.php

makinog3:
Failed to parse received email with "<" or ">" in the "To:" header and did not provide feedback.
Some mail systems generate "To:" header with "<" and ">" when replying to the "Return-Path:"  of the original email in our case.

Ex)
`To: filesender+recipient-1234@sample.com` -> OK
`To: <filesender+recipient-1234@sample.com>` -> NG

### DIFF
--- a/scripts/task/emailfeedback.php
+++ b/scripts/task/emailfeedback.php
@@ -93,10 +93,9 @@ while($inputs) {
         
         $return_path = Config::get('email_return_path');
         if($return_path && preg_match('`^(.+)<verp>(.+)$`i', $return_path, $match)) {
-            $addr = explode('@', (string)$headers->to);
-            $variable = substr($addr[0], strlen($match[1]));
-            
-            if(preg_match('/^(recipient|guest)-([0-9]+)$/', $variable, $m)) {
+            $regex = '`' . preg_replace('`([.+])`', '\\\\${1}', $match[1])
+                . '(recipient|guest)-([0-9]+)' . preg_replace('`([.+])`', '\\\\${1}', $match[2]) . '`';
+            if(preg_match($regex, (string)$headers->to, $m)) {
                 $target_type = $m[1];
                 $target_id = $m[2];
                 


monkeyiq:
I will be taking a look at this soon. Sorry for the delay.

monkeyiq:
Unfortunately the `email_return_path` is underdocumented in FileSender. Plus, I don't know as much about verp as a sys admin would :/

With that in mind I have been playing with the following setting:
```
$return_path = "filesender+<verp>@sample.com";
```

It is quite strange that the existing code is breaking apart the config setting with a regex and then using fixed string lengths to get do a substring on the `$headers->to`.  

Does it work when you simply remove the encasing <> characters? Such as in the code here https://github.com/monkeyiq/filesender/blob/c50fbeedc5c4f53d8476fa455d214ccb560e061a/scripts/task/emailfeedback.php#L98

My main concern with how this PR solves the issue is if the email_return_path contains something other than the escaped `+` and `.` characters that does not allow the build `$regex` to match.

makinog3:
Sorry for the late reply.
The code that you suggested is fine when only the "<" ">" are contained.
According to [RFC 822](https://datatracker.ietf.org/doc/html/rfc822), the "To:" header can contain characters other than "<" ">", so my code uses regular expressions to identify them.
`+` and `. ` are escaped because they make sense as regular expressions and can be included as email addresses.

Ex)
`To: File Sender Admin <filesender+recipient-1234@sample.com>`
`To: (File Sernder Admin) filesender+recipient-1234@sample.com`

However, I think this email is most likely sent by a regular SMTP server, so a strange email address will not be set as To, and your code will be fine.

monkeyiq:
I have merged the above update for the `<>` brackets. I think the next move might be to update for the `+` and `.` too.

monkeyiq:
I think I will close this. If there are strange `To` being sent by an SMTP server then we could look back at this and I can try to update the regexes in use to cater for that. Thank you for raising the issue and the PR. Unfortunately it is always a bit of a balance between being able to review the regex changes and the functionality gained.

PR #1779, opened 2024-01-18
Fix french typo

yent:


WebSpider:
Hi Yent,

Changes to languages in github will be overwritten automatically by the content of POEditor by the next release. Could you make sure the translation is updated in POEditor?

monkeyiq:
Since it was only a single update I have done that for FileSender 2.0 and FileSender 3.0 in poeditor. 

A smoother update would be to perform the updates on poeditor and allow the import scripts to clobber the `language/` files with the data from poeditor.

monkeyiq:
Thanks for the update.

Issue #1038, opened 2021-05-12
Possible to use encrpytion via API?


Trefex:
Hi,

I had a look at the Rest API and I cannot see how to add encryption when creating a transfer.

Could you let me know if creating a transfer with encryption can be automated at all?

Thanks

monkeyiq:
The encryption feature uses the web crypto API to work. I had an idea to write a rest client using nodejs if there was a suitable web crypto implementation that could be used from there https://github.com/filesender/filesender/issues/861. Another option might be to use a web browser in headless mode. A bit resource heavy but it would be nice to be able to upload/download encrypted content from an automated pipeline.

If we can't reuse the same javascript code for client side encryption then there may be a fairly big maintenance overhead making sure that the REST client code and javascript used by the browser still match up correctly.

WebSpider:
It would seem that the basic calls that the code uses (deriveKey, getRandomValues and asmcrypto) are present in nodejs after v15. This would open up the way for a nodejs implementation.
For someone like me who knows python and not nodejs, this seems like ... quite the task :)

monkeyiq:
There is now a proof of concept to upload from the command line with encryption using nodejs

https://github.com/monkeyiq/filesender/tree/2023/apr/jsrestclientpoc

Though there is also a laundry list of required updates before that can become a real rest client and be merged.

WebSpider:
With encryption now being present in the python client and a nodejs equivalent, I am going to assume the way to incorporate encryption in an API client is sufficiently available.

PR #1780, opened 2024-01-18
Encryption support in python client

PaulLocksley:
Hi Team, 
I've been working on the start of encryption support in the python client. This can also work as an example for issues such as [1038](https://github.com/filesender/filesender/issues/1038 ). 

Currently it only supports AES-GCM but I tried to write it in such a way that is easy to expand in the future. 

The user simply specified -e with their desired password much like they do for email address already.

### Changes

There are 3 new imports required, base64 and Enums are from the standard library but the [Cryptography library](https://github.com/pyca/cryptography) is 3rd party

As it is a 3rd party library and may not be present it has been moved so it only impacts users who haven't installed required library but also want to use Encryption. 

There should be no code changes for users who do not use the -e argument. 

### Testing

I've tested several transfers with large files and multiple files with and without encryption. 
The encrypted transfers behave as expected are decrypted by the web client without issue.
The non encrypted transfers behaved as usual.

If you would like anything to change or don't think this is appropriate for the client let me know.

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -30,22 +30,25 @@
 
 import argparse
 try:
-  import textwrap #used to format help description and epilog
-  import requests
-  import time
-  import re
-  from collections.abc import Iterable
-  from collections.abc import MutableMapping
-  import hmac
-  import concurrent.futures
-  import hashlib
-  import urllib3
-  import os
-  import sys
-  import json
-  import configparser
-  from os.path import expanduser
-  from math import ceil
+    import textwrap #used to format help description and epilog
+    import requests
+    import base64
+    import time
+    import re
+    from collections.abc import Iterable
+    from enum import Enum
+    from collections.abc import MutableMapping
+    import hmac
+    from cryptography.hazmat.primitives.ciphers.aead import AESGCM
+    import concurrent.futures
+    import hashlib
+    import urllib3
+    import os
+    import sys
+    import json
+    import configparser
+    from os.path import expanduser
+    from math import ceil
 except Exception as e:
   print(type(e))
   print(e.args)
@@ -76,6 +79,8 @@ if 'user' in config:
   username = config['user'].get('username')
   apikey = config['user'].get('apikey')
 
+
+
 #argv
 parser = argparse.ArgumentParser(
     formatter_class=argparse.RawDescriptionHelpFormatter,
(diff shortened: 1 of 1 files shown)


WebSpider:
Is it your goal to make this compatible with the webclient crypto?

PaulLocksley:
>Is it your goal to make this compatible with the webclient crypto?

As in support all the same cryptography types? If there is a desire, I wanted to put this forward with just a single type before I put more effort into it. 
 
This AES-GCM implementation is able to be decrypted by the web crypto client.
If that is what you mean so its compatible in that way. 

I had a quick look and it looks like AES-GCM is less popular with 5* servers nominating it in filesender-config.js.php compared to the 31 using AES-CBC so that would probably be a good next step.

<sub>* from servers listed in the [known installs list](https://docs.filesender.org/filesender/known-installs).</sub>

WebSpider:
I guess my question could use a bit more explanation, sorry for that :-)

My main question is: When you upload a file using the python client in this PR, will the webclient be able to decrypt it?

A quick review of this PR shows there doesn't seem to be a PDKDF2 implementation in here, so how would the password be communicated, for instance?

PaulLocksley:
>My main question is: When you upload a file using the python client in this PR, will the webclient be able to decrypt it?

Yes. 

The password is communicated via the -e or --encrypted argument, you simply add the password just like you do for recepiant so -e=Password1234!

From the file recipients point of view it just looks like a normal transfer, it works if they put in the correct password and fails if they put in the wrong password with a warning.

 ---------------
PBKDF2 is used in the key generation step, its part of Pythons standard library so that was fairly straight forward. 

I validated the encryption with the following steps:

1. I took snapshots of the IV , Salt and encrypted data from a real transfer. 
2. I saved the undecrypted first chunk of the same transfer. 
3. I then performed the encryption and confirmed the binary data of the file was identical to the expected data.  

For testing I validated with transfers less than chunk size, and 60mb / 100mb multifile transfers to validate chunking encryption behaviour. 

On download the md5 hash's all matched the uploaded files.

--------------------------


Let me know if I have misunderstood any of the requirements.

WebSpider:
Wow, thanks a lot for this contribution, amazing! 🙏

I will take some time this week and test this on filesender!

monkeyiq:
This is looking cool. I like that `encryption_password_hash_iterations_new_files` is respected too!

I will have more of a dig into things in the coming days. I have performed an upload in script and download in browser test. There are a number of tests in the javascript which are in place at the request of an NREN and reviewed by a cryptographer to check the sanity of the assertions from that point of view. I'm not too sure what to do there, if replicating them is needed. One involving overflow by an upload being too large in GCM is definitely worth adding in the python code to ensure no bad things happen. Having another command line method to upload encrypted files is always welcome.

monkeyiq:
It may also be a nice idea to include a more user friendly message if `encryption_key_version_new_files` is not == 3. At the moment it fails with the following if I set the key version to 1:
```
Failed to parse encryption requirements, encryption cannot be supported.
```

Issue #1782, opened 2024-01-30
Move to a more recent UUID version


monkeyiq:
Later UUID versions maintain a portion of random data to make guessing and collisions difficult but are prefixed with time information. This setup allows databases and btrees to more effectively index a set of UUIDs. This will help as the files are stored using the UUID from the database as file names or directory names and file systems may in general use btrees to index the file/directory names. This should lead to shallower btrees and less wasted cycles on the server.

WebSpider:
These collisions are exactly the reason why daily buckets are used on s3, so awesome enhancement!

monkeyiq:
Thinking about this a bit more today, maybe daily buckets would be a good enhancement for the filesystem based storage too. On the paper napkin test, as one site I know has 50k active users per day and perhaps they allow 30 days holding that would make a 1.5 million entry directory. Having 30 directories of 50k+ files would be much more efficient.

PR #1789, opened 2024-02-05
uuidv7 use and per day and hour buckets for filesystem storage

monkeyiq:
This brings in two features: use of uuidv7 for uids and the ability to store files in subdirectories based on the day and hour they were created. This will reduce the number of entries in the `/opt/filesender/files` directory to one new directory per day and inside that 24 hour directories (if both are enabled). The v7 uuids have a stable prefix in them to allow better indexing in databases. As the uuid are used as directory and file names and many kernel filesystems use btrees to index the names of entries in a directory the uuid7 format used should also aid in quick lookup of files by the kernel.

I have used the php 8 requiring version of the uuid library.  I think since php 8.0 is itself out of security support moving to requiring php8+ in FileSender is not a huge ask.  I think the test suite etc should be updated to using php 8 at least as that is now old enough to be out of support https://www.php.net/supported-versions.php

This relates to https://github.com/filesender/filesender/issues/1782

Toward the end of the PR I added forward and backward migration support for the bucket settings. This should allow existing transfers to continue to be able to be downloaded when buckets are enabled. Also if you don't want the buckets any more they can be disabled and new transfers will not use them but the existing ones that do can still be accessed.

```
$config['storage_filesystem_per_day_buckets'] = true;
$config['storage_filesystem_per_hour_buckets'] = true;
```


monkeyiq:
It seems that some of the issues are because the test suite assumes there is no subpath

WebSpider:
Wow, big PR :)

Could you see if you can make these changes in `development3` as well, so there is some feature parity?

monkeyiq:
Some of the deps updates made it much bigger than it could be. The lib/vendor stuff could be split into another PR or ignored in this one to an extent.

monkeyiq:
I'll be digging into this a bit more in the next days with an eye to merge. I held it off until after the latest 2.x release so as not to conflate those changes with these.

monkeyiq:
I will look to move all  the stuff in development to development3 again at some stage. After the  sonarqube updates in development3 it will likely be more difficult than usual due to merge conflicts.

yent:
Hum, not so thrilled about this change, while it seems useful for storage it poses a significant problem for recipient and guests tokens generation.

The previous code was generating uuids with 122 bits of entropy, the new one only has 62 bits of entropy (time part cannot be considered random).

According to the french authority for data security (ANSI) an identification key is strong if it has > 100 bits of entropy and weak if it has < 65 bits, so we likely won't pass security audits with this (that is, we won't be able to upgrade anymore).

Why not use uuid7 for storage specifically but still use a more robust uuid scheme for recipients and guests token ?

yent:
I just did a pull request for those changes : #1810

PR #1810, opened 2024-03-07
Use uuidv7 for storage and uuidv4 for key-based identification

yent:
Recent use of uuidv7 for all uids led to a significant loss of entropy for key-based identification (recipients, guests ...)

While the use of uuidv7 for storage is not an issue, its use for identification may be because of its weakness, these changes keep the use of uuidv7 for storage names but use uuidv4 with more entropy when generating identification keys (recipient tokens, guests tokens, email translation tokens, web interface security tokens ...)

### DIFF
--- a/classes/data/File.class.php
+++ b/classes/data/File.class.php
@@ -339,8 +339,8 @@ class File extends DBObject
         $file->transfer_id = $transfer->id;
         $file->transferCache = $transfer;
 
-        // Generate uid until it is indeed unique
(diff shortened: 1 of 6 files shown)


monkeyiq:
I have moved all uuids to version 4 again for now https://github.com/filesender/filesender/pull/1812

I think we would have to specifically create a storageuuid in the files table and use that on the server only to ensure uuidv7 is only used locally to the server.

Issue #1781, opened 2024-01-26
Attribute requirements for Filesender


surfnet-niels:
The install docs (https://docs.filesender.org/filesender/v2.0/install/) talk about the IdP attribute requirements and point to https://docs.filesender.org/filesender/v2.0/admin/reference/#idp_attributes. However that page does not have a section on IdP requirements.

monkeyiq:
Thanks for the report. I will update the docs soon now that I have more of an understanding of IdP attributes.

monkeyiq:
Before I get the docs update... the AuthSPSaml.class.php file processes the attributes. It looks for uid, name, and email. The uid and email must be present. The name will be taken from the email if it is not present. 

These uid, name, and email are found using the filesender config.php settings `auth_sp_saml_email_attribute` https://docs.filesender.org/filesender/v2.0/admin/configuration/#auth_sp_saml_email_attribute and the like. So by default email is taken from the SAML attribute mail, name from the SAML cn, and the uid from the SAML eduPersonTargetedId. The filesender config.php settings let you change what SAML attribute to inspect (or a list of them) to obtain these values.

In a default FileSender install the SAML attributes `eduPersonTargetedId` and `mail` are required to login.


The `auth_sp_additional_attributes` config.php setting can be used to allow access to other SAML attributes as well.

monkeyiq:
If the above makes sense I will use that as the core of a docs update

monkeyiq:
I have merged the above information into the reference docs.
https://github.com/filesender/filesender/pull/1795

monkeyiq:
Please reopen the issue or submit new ones if find the docs lacking. Thank you.

peter-:
> So by default email is taken from the SAML attribute mail, name from the SAML cn, and the uid from the SAML eduPersonTargetedId. The filesender config.php settings let you change what SAML attribute to inspect (or a list of them) to obtain these values.
> 
> In a default FileSender install the SAML attributes `eduPersonTargetedId` and `mail` are required to login.

Note that the [eduPersonTargetedID attribute has finally been deprecated 4 years ago](https://wiki.refeds.org/display/STAN/eduPerson+2020-01#eduPerson202001-ChangeLog) in the relevant specification. As such that makes for a bad default because it really should be avoided for any new services/deployment (with `pairwise-id` from the [SAML 2.0 SubjectID Attributes Profile](https://wiki.oasis-open.org/security/SAMLSubjectIDAttr) being the most direct replacement).

[...]

PR #1822, opened 2024-03-13
SAML: update default id attibute to the more modern pairwise-id

monkeyiq:
This updates the default SAML user ID attribute from the deprecated `eduPersonTargetedId` to the newer `pairwise-id`.

Interested readers may consider https://docs.oasis-open.org/security/saml-subject-id-attr/v1.0/cs01/saml-subject-id-attr-v1.0-cs01.html for more information.

This was kindly suggested in a comment at https://github.com/filesender/filesender/issues/1781#issuecomment-1991258902

### DIFF
--- a/ci/filesender-config.php
+++ b/ci/filesender-config.php
@@ -301,8 +301,7 @@ $config['auth_sp_type'] = 'saml';
 // // Get name attribute from authentication service
  $config['auth_sp_saml_name_attribute']         =       'cn';
 // 
-// // Get uid attribute from authentication service.  Usually eduPersonTargetedId or eduPersonPrincipalName
-$config['auth_sp_saml_uid_attribute']           =       'eduPersonPrincipalName';
+$config['auth_sp_saml_uid_attribute']           =       'pairwise-id';
 // 
 // // Get path  attribute from authentication service
 $config['auth_sp_saml_authentication_source']   =       'default-sp';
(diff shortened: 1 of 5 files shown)


WebSpider:
Given that this is a defaults change, maybe consider moving this to 3.0, instead of incremental in 2.x?

monkeyiq:
I was thinking since the default setting used by FileSender has been deprecated for so many years it might make sense to change it to a more modern one. Obviously I would have a big mention of that in the release notes with the historic value for sites that have not needed to change the setting. I could just make the change to the 3.x series.

Though it is probably an idea to have some procedure or precedent for changing the default values shipped by filesender. Assuming something might happen in the 3.x release life time where we might feel it a plus to change a default. One example from the 2.x series is `encryption_key_version_new_files` which is still at 1 when 3 is a much better default and has been the case for *years*.

WebSpider:
> I was thinking since the default setting used by FileSender has been deprecated for so many years it might make sense to change it to a more modern one. Obviously I would have a big mention of that in the release notes with the historic value for sites that have not needed to change the setting. I could just make the change to the 3.x series.

Yeah, while I agree, we should take into account that when people just upgrade FileSender inside `2.x`, it should not break if you're running on all defaults. That would be a bad experience.

> 
> Though it is probably an idea to have some procedure or precedent for changing the default values shipped by filesender. Assuming something might happen in the 3.x release life time where we might feel it a plus to change a default. One example from the 2.x series is `encryption_key_version_new_files` which is still at 1 when 3 is a much better default and has been the case for _years_.

I agree!
Since the change to `encryption_key_version_new_files` does not change anything significantly for existing transfers and users, I see no reason not to change that default, as long it is in the release notes.

monkeyiq:
I am going to merge the changes into development3 with https://github.com/filesender/filesender/pull/1836 and close this.

Issue #1619, opened 2023-11-02
UI3: Code details - Resume transfer


WebSpider:
(code34) upload: when you leave an active large upload to stop it (only way to do that) and then return FileSender offers to resume the old transfer. Click yes.

No hints as to what files you need to add again to resume the transfer. If you add a file then it may be rejected and finally a hint is shown as to  what you need to add again to resume the transfer.

WebSpider:
I encountered this during tests as well, and the current behaviour is counter-intuitive. There is no way to pause/resume a download using buttons, and the tab filesender is running in, becomes pretty unresponsive as it is busy with the uploads.

Suggestion to improve this: Add at least a pause/resume button below the list of files that is being uploaded.

WanderGomes:
I'll check this

meijer:
Wander, any news?

adminnr1:
Is there an update for this ticket? This is a blocking issue for us. So keen to hear about the progress.

monkeyiq:
The release notes of each release list both the PR title and link directly to it. If there are better ways to convey information to the community I am also interested for other improvements and new additions (eg, nodejs rest client).

monkeyiq:
With https://github.com/filesender/filesender/pull/1832 the resume dialog appears and adds the files back in the file selection section to hint that you need to use the "add files" button to set them up again for the upload.

PR #1832, opened 2024-03-19
Fix/resume transfer

WanderGomes:
Fixing resume transfer feature and updating the modal button classes


WebSpider:
Hopefully fixes #1619

Issue #1838, opened 2024-03-25
Running filesender in PHP8.2+ causes deprecation warnings


WebSpider:
When running filesender in PHP8.2 of 8.3, many deprecation warnings are issued, sometimes causing HTML to be clobbered.

These deprecation warnings are about the 'Creation of dynamic property XXX is deprecated'

Since PHP8.1 is out of active support, and only receives security updates for roughly 8 more months, this deprecation warning should probably be fixed.

I believe it is related to the first issue on the list here: https://www.php.net/manual/en/migration82.deprecated.php

I've seen the warnings so far relating to:
- Translation::$allow_replace
- Browser:$isEdge

but there probably are more around

monkeyiq:
I will merge things from development into development3 and then mod dev3 to not use strftime and have updated time formatting strings.

WebSpider:
Ran into another one: 
Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /opt/filesender/filesender-master3-filesender-3.0alpha6/classes/data/ShredFile.class.php on line 253

Do you want a seperate issue for this?

monkeyiq:
There are probably still a few around. adding them here is fine.

PR #1839, opened 2024-03-26
fix for some deprecations in modern php

monkeyiq:
For the FS3 series we should probably remove the use of strftime and change the string formatting for date formats to the non strftime format at the same time.

This relates to https://github.com/filesender/filesender/issues/1838


WebSpider:
This got merged into `development`, not `development3`, was that intentional?

PR #1841, opened 2024-03-26
feat: user theme selector; theme improvements;

WanderGomes:


monkeyiq:
It would be nice to have more of a description.

It seems
* There is a new theme selector on the profile page which allows selecting a theme and that is only saved in the current browser session. A future PR will allow this to be saved in the database (if that is indeed something that is planned).
* There are theme modifications including the introduction of gold and silver contributors. These classes of contributors are outlined at http://www.example.com/link-here.

In this case when I go to the profile page I can select the dark theme but it is lost if I navigate to any other page from there or if I click on the save button on the profile page.

WanderGomes:
The new theme selector feature works as follows:

We have the theme files, default.css (light) and dark.css, in these theme files, we have the CSS 3.0 colors variables used by the pages and components, these variables can be modified and there can be several different themes. The theme selector feature just load the CSS theme file selected by the user.

On the profile page, we have three theme options:

•⁠ ⁠Light
•⁠ ⁠Dark
•⁠ ⁠Device

The "Device" theme is the theme that is selected by default, this theme follows the user's browser/operating system theme, if the user's browser/OS theme is set to light, then FileSender will select and apply the light theme, if the user's browser/OS theme is set to dark, then FileSender will select and apply the dark theme.

The user can change the theme on the profile page, don't need to press the save button, it changes and saves the theme automatically, the selected theme is saved in the user's browser's local storage, and whenever the user accesses the FileSender, the theme.js file searches for the theme saved in local storage, and loads the previously chosen theme, if there is no theme saved, the user's browser/OS theme is applied (device theme).

I tried my best to explain the feature. If you require any further information, let me know.

monkeyiq:
Thank you for the detailed description.

PR #1853, opened 2024-04-03
V3 theme block config

madsi1m:


madsi1m:
last commit fixes hardcoded /filesender/ in theme css path

PR #1875, opened 2024-04-28
Update filesender.py

tijldeneut:
Small bugfix when running this script without the username argument:
E.g.:
`python3 filesender.py -h`

```
Traceback (most recent call last):
  File "/home/tijl/filesender.py", line 130, in <module>
    if not bool(re.match("([^@|\s]+@[^@]+\.[^@|\s]+)", username)):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/__init__.py", line 166, in match
    return _compile(pattern, flags).match(string)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'NoneType'
```
The variable "username" is addressed before it was instantiated, this might not be the best resolution, but it works.

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -126,7 +126,8 @@ else:
 requiredNamed.add_argument("-r", "--recipients", required=True)
 
 # if username is not a valid email address then ensure user supplies a valid email address
-if not bool(re.match("([^@|\s]+@[^@]+\.[^@|\s]+)", username)):
+args = parser.parse_args()
+if args.username and not bool(re.match("([^@|\s]+@[^@]+\.[^@|\s]+)", args.username)):
   requiredNamed.add_argument("-f", "--from_address", help="filesender email from address", required=True)
 
 args = parser.parse_args()


monkeyiq:
Isn't the `username` set on line 73? The problem seems to be that `re.match` may be passed the `None` value that is initially set for `username`. 

So we could do the `requiredNamed.add_argument("-f", "--from_address"...` code if `username` is still set to `None` or if the regex fails matches. ie the username was not able to be loaded from config?

PaulLocksley:
I agree, I think its better to check if its None, instead of parsing the arguments twice. 
ie: ```if username is None or not bool(re.match(r"([^@|\s]+@[^@]+\.[^@|\s]+)", username)):```
I believe/ did some quick testing and think this code will still error with no username and an invalid config file.

This does raise two questions for me:
1.  Should we also be doing the regex check for the user provided argument?
2.  On one of my computers the regex string prints out a warning, probably due to a python version difference.  Changing the pattern to a string literal seems to fix it: 
ie ```re.match(r"([^@|\s]+@[^@]+\.[^@|\s]+)", username)```

tijldeneut:
Yeah, I agree, this is probably a more proper solution:

`if username is None or not bool(re.match(r"([^@|\s]+@[^@]+\.[^@|\s]+)", username)):`

monkeyiq:
@tijldeneut would you like to update the PR and push the changes for a merge? Thank you for reporting the issue and also going the extra mile to make a PR for it!

monkeyiq:
I have merged the recommended update in https://github.com/filesender/filesender/pull/1906.

Thank you for reporting the issue. Thank you for also making a PR to suggest the update!

Issue #1887, opened 2024-05-02
Graph display is very slow and uses a lot of memory


makinog3:
When there are many transfer histories, the graph display on the upload screen is very slow and uses a lot of memory on the MySQL database.
In our site's case, it is normally over 20 seconds.

Unfortunately, our site went down because it ran out of memory when multiple users displayed the upload screen simultaneously.
So, we don't show the graph now.

* Number of rows in Transfers table;
about 40,000

* Number of rows in AuditLogs table;
about 800,000

* Number of rows in StatLogs table;
13,843,255 

* SQL
```
SELECT days.date, speed.speed, speed.enspeed FROM (SELECT (SELECT Date(NOW() - INTERVAL '30' DAY)) + INTERVAL a+b DAY date
FROM (SELECT 0 a UNION SELECT 1 a UNION SELECT 2 UNION SELECT 3 UNION SELECT 4
UNION SELECT 5 UNION SELECT 6 UNION SELECT 7 UNION SELECT 8 UNION SELECT 9 ) d,
(SELECT 0 b UNION SELECT 10 UNION SELECT 20 UNION SELECT 30 UNION SELECT 40) m
WHERE (SELECT Date(NOW() - INTERVAL '30' DAY)) + INTERVAL a+b DAY <= (select date(now()))
ORDER BY a + b) as days LEFT  JOIN (SELECT DATE(created) as date,
AVG(case WHEN time_taken > 0 AND additional_attributes LIKE '%\"encryption\":false%' THEN size/time_taken ELSE null END) as speed,
AVG(case WHEN time_taken > 0 AND additional_attributes LIKE '%\"encryption\":true%' THEN size/time_taken ELSE null END) as enspeed,
AVG(case WHEN additional_attributes LIKE '%\"encryption\":false%' THEN id ELSE null END) as count,
AVG(case WHEN additional_attributes LIKE '%\"encryption\":true%' THEN id ELSE null END) as encount
from StatLogs WHERE event='file_uploaded' AND created>NOW() - INTERVAL '31' DAY AND size > 1073741824
GROUP BY Date) as speed on days.date=speed.date  ORDER BY days.date;
```

monkeyiq:
Could you please post the output of an `EXPLAIN` on the query. I may be able to improve things without it but knowing what the database is thinking might also help to reason about an index which might help.

makinog3:
Thank you for your response.
We use MariaDB 5.5

```
[...]

PR #1892, opened 2024-05-05
option to precache values for upload graph

monkeyiq:
A new option that, when enabled, allows the upload graph to used cached data in the database rather than executing a more complex query each time the upload page is loaded.

Using this option requires you to run `scripts/task/update-upload-graph-table.php` in order to keep the cached data up to date. This can mean one execution every 15 minutes, hour, or whatnot instead of one execution for each load of the upload page.

This relates to https://github.com/filesender/filesender/issues/1887.

### DIFF
--- a/classes/autoload.php
+++ b/classes/autoload.php
@@ -85,6 +85,7 @@ class Autoloader
         'AVResult' => 'data/',
         'DownloadOneTimePassword' => 'data/',
         'RateLimitHistory' => 'data/',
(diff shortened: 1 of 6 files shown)


monkeyiq:
It may be that interesting to use different SQL that updates the cache table in a single query instead of relying on database transactions to protect the delete/insert pair.

monkeyiq:
The last push uses upserts to isolate things without needing transactions.

Issue #1881, opened 2024-05-01
UI3: feature request: logged out pop up?


madsi1m:
If we log into filesender and leave it (go to lunch) and then try and transfer a file, it fails because the session has expired.
Be nice to do a popup when the session dies (and we are not already uploading)

This has already happened to one of our testers

monkeyiq:
This sounds good. I have also seen somebody leave the partly setup upload page open in a suspended laptop and it asked for a login again and was a bit confusing when they tried to complete. I guess for that case people don't often stall at the almost doing an upload stage.

monkeyiq:
This has been merged and released. Thank you for the issue report.

PR #1905, opened 2024-05-17
ui: option to warn user if session has expired

monkeyiq:
For SAML sessions, when the new `auth_warn_session_expired` feature is enabled a cookie is used to allow the client to warn the user when their session has expired. This warning is not presented when an upload is in progress because an upload can outlive a session in order to complete.

This relates to https://github.com/filesender/filesender/issues/1881


monkeyiq:
It makes things a bit quicker testing this when modifying the saml config.php `'session.duration' => 20,` so things move along a bit quicker. Ungainly session for actual interaction, but handy to see when the time is up :)

PR #1911, opened 2024-05-23
SimpleSAMLphp2xSupport

so010:
Referring to https://github.com/filesender/filesender/pull/1783/commits/0bf3d1ab657828a87295ed7922c0c8056a748b30#diff-1cbe5813828ff8dbaba5d417dbb28d6ee62f0657e63f41303d4c3a596cc09efa

make SimpleSAMLphp 2.x logout support for filesender3

I don't speak PHP so review carefully

### DIFF
--- a/templates/user_page.php
+++ b/templates/user_page.php
@@ -173,18 +173,7 @@ $user = Auth::user();
                             $icon = '<i class="fa '.$faicon.'"></i> ';
 
                             $url = AuthSP::logoffURL();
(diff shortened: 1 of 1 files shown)


madsi1m:
I've effectivy done the same but I have not tested on older SSP versions:

## fix logout, SSP doesn't like POST
sed -i 's/"saml"/"saml-MD-broken-so-bypass!"/' /filesender/templates/user_page.php

PR #1936, opened 2024-06-17
Add parameter to Python client so it can be used without config file

ToonSpinTUe:
Hi,

I want to use the Filesender Python client from GitHub actions, and I am running into the fact that the Python client needs the `filesender.py.ini` to be in a certain location in the home directory. I would like to run the client without having to create the .ini file and placing it in a certain directory with a certain name.

The easiest way to fix this, in my opinion, is to add a parameter `--base_url`, so that every setting from the .ini file can be passed on the command line. I have tested these changes by making filesender requests over the CLI with no ini file with filesender.surf.nl, which is the Filesender instance I use at work.

I have also considered adding a parameter for the location of the ini file, but that makes for a much bigger change - this way the changes are much more readable for you.

Also, this PR fixes a few typos.

Everything should be fully backwards compatible with previous versions of the client.

Thanks in advance for your time,

Toon Spin


monkeyiq:
Thank you for the update. Sorry about the delay.

I am also open to a PR to use a custom config file location as an argument. Such a feature might be useful for other testing such as using different username/apikey during command line testing.

PR #1937, opened 2024-06-18
Create v3 documentation

WebSpider:
V3 documentation is initially created as a copy of V2 documentation, with some small changes:
- Search / Replace v2 to v3
- Updated POIeditor link as it is different for V3
- Updated support statement for v2 (only security updates after v3 is released)
- Changed upgrade notes to make clear that a direct upgrade from v1 to v3 is not supported


monkeyiq:
I had put myself forward to do this a week ago. I had a local copy with updates already though some activities of the week slowed me down pushing that tree. 

That is now merged at https://github.com/filesender/filesender/pull/1941

I relinquish ownership of the docs update now.  Feel free to make PRs against the new docs. I don't think they will conflict as badly now and we can both work in individual files. The changes in 1941 will be merged forward in the normal release process.

PR #1961, opened 2024-08-01
Fix range offset calculation

yent:
Fix download range offset math, most likely was a typo


monkeyiq:
I am guessing that this is intended to allow end to be set negative to get just that many bytes from the end of the file. It may also be handy to make sure that `$start >= 0` in there to double check out of bounds attempts.

PR #1963, opened 2024-08-04
Fix S3 guests losing transfer settings

WebSpider:
When upgrading an install from 2.40 to 2.48 recently, we ran across an issue where transfer settings for guests would be lost. 

This PR should fix this.


monkeyiq:
The FileSender board wishes for all PRs to be made against the `development` branch. Then updates to be patched in a single application against `master` at time of release.

WebSpider:
Changing target branch

PR #1971, opened 2024-08-15
Pull out images and make them attachments

madsi1m:
My attempt at getting images in emails to show on most mail clients. Long story short, images need to be attachments.

How it works:
in the email templates set images as:
```
<img src="cid:<name>;<filepath>">
```
Where name is a unique name for the attachment and filepath is the path either absolute or relative to the filesender/www directory
For example:
```
<img src="cid:footerimage;images/footer.png">
```

### DIFF
--- a/classes/utils/ApplicationMail.class.php
+++ b/classes/utils/ApplicationMail.class.php
@@ -84,7 +84,25 @@ class ApplicationMail extends Mail
             $this->writePlain($content->plain);
             
             if ($use_html) {
-                $this->writeHTML($content->html);
+                // Do we have images in the html that should be attachments? Eg: <img src="cid:<name>;<filepath>">
+                $html = explode("\n",$content->html);
+                $htmlLength = count($html);
+                for ($i = 0; $i < $htmlLength; $i++) {
+                    $pos = strpos($html[$i],'"cid:');
+                    if ($pos === false) continue;
+                    $topos = strpos($html[$i],'"',$pos+1);
+                    if ($topos === false) continue;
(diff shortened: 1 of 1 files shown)


monkeyiq:
I am assuming the example code worked ok against gmail :) I will take a look at this shortly.

madsi1m:
gmail + outlook/exchange for the win!

monkeyiq:
I was originally thinking about doing something fancy like this PR to allow automatic image attachments via the cid. One issue here is that at that stage we are processing the html after all lang and other substitutions have occurred. It may be possible at that time to have crafted some data on the page to include an "image" file that was not intended to be attached to the email. 

I have an update to this in  https://github.com/filesender/filesender/pull/1972 

The update requires that images be listed in the config in a new key `template_email_images`. This is an array of `cid -> path`. As this array is in your config one might assume it is a good actor and does not contain paths like `/etc/passwd`.  You can then use the `cid` in your template directly and the mail code checks for an explicit `img src="cid:` tag with only lower characters in the `cid` definition. If that cid is found in the config then the associated image from the config will be attached to the email. The intention is that even in the face of an explicitly nefarious user we will only allow images from paths in the config variable to ever be sent. The path is never in the template, only the value to select the `$cid`. Since the HTML is ok to send as is no update is needed to it, just some matches looking for fairly narrow regexes on an image src tag to find the cids that are of interest.

The config variable is slightly more work but I imagine that images will be fairly few and fairly static over time so it shouldn't be a huge thing to define them in a config variable.

madsi1m:
Closing mine as I prefer yours

PR #1972, opened 2024-08-15
allow image attachments from mail templates

monkeyiq:
This idea was originally proposed in https://github.com/filesender/filesender/pull/1971. 

This update requires the images that can be used in the cid to be listed in a config value and limits which cids can be selected to only using a range of a-z characters. If the cid is not in the config variable then no image will be attached.

### DIFF
--- a/classes/utils/ApplicationMail.class.php
+++ b/classes/utils/ApplicationMail.class.php
@@ -84,6 +84,21 @@ class ApplicationMail extends Mail
             $this->writePlain($content->plain);
             
             if ($use_html) {
+                
+                // Do we have images in the html that should be attachments? Eg: <img src="cid:<name>;<filepath>">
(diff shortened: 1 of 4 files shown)


monkeyiq:
I will perform more validation and checks on this in the next days.

madsi1m:
I prefer this

madsi1m:
But also merge in development3 :)

monkeyiq:
Will do. I imagine you will be wanting dev->dev3 code movements more frequently now :)  Certainly for stuff you have made a PR for :)) 

Thank you for the PR. I will mark some time on Sunday to retest this and merge it forward into dev3.

monkeyiq:
OK. A bit more expanded. A few cases where the config might generate no image attachment now also make log events. Also merged into development3.

madsi1m:
cheers

Issue #1974, opened 2024-08-19
CLIDownload.php currenlty provides a file that cannot easily be used with the host institution


PaulLocksley:
The Text replace currently in use results in the following if statement in the python client:
![image](https://github.com/user-attachments/assets/86fa398e-fb65-4b63-b91b-fe39dda9ca39) 
This makes  the user add an argument to use the expected filesender server.

I believe this is caused by the following [line](https://github.com/filesender/filesender/blob/e3afc10a734e70128c69f5d859e9b3073e6eba85/www/clidownload.php#L66C1-L66C57). From my understanding either removing this line or replacing the method with a string replace that can cap it at 2 replacements should resolve this issue.

PaulLocksley:
I have a proposed fix on https://github.com/PaulLocksley/filesender/blob/CLIDownload-fix-test/www/clidownload.php 

Replacing the string replace with a preg_replace that allows limiting to the first match only.
We already use preg_replace just below that so there shouldn't be any issues there. 

If that is an appropriate way to fix it let me know and I can open a pr.

monkeyiq:
I think I while ago the link for the script itself was changed to be just the original script off github.

https://github.com/filesender/filesender/blob/5fd763a9bc423fef13599332d642942998b3981f/templates/user_page.php#L413

This was likely done because when the user is going to make a configuration file anyway which will contain the base_url, username, apikey in one location then there isn't much need to update the script itself.

That said, if you would like to limit the base_url update to just the first match I am happy to include that update. It seems the original script is trying to limit the match to just `$one` but may not be doing that any more. The regex could obviously match and replace the whole line as well which would be more secure and less brittle.

PaulLocksley:
Ah thanks for that, it looks like we were injecting the clidownload.php link on that page in our language file...
We will remove that and shouldn't have any users with this issue any more :) 

I will still send through a PR for dev3 so any old links continue to work, if that is not desired feel free to close this issue.

monkeyiq:
I have merged https://github.com/filesender/filesender/pull/1975. Thank you for reporting and fixing this!

PR #1975, opened 2024-08-20
Clidownload.php fix for python client

PaulLocksley:
Hey Team, 

Minor change to address: https://github.com/filesender/filesender/issues/1974 

The CLIDownload will now only replace the default [base_url] string in the python file. 

I tested this on our dev server to confirm the new expected behavior, I kept the $one variable tracking how many changes. there have been but as far as I can tell that isn't used.


monkeyiq:
Thank you for the detailed issue and PR.

Issue #1976, opened 2024-08-22
UI3: some templates still have inline styles


madsi1m:
We should add these to use CSS sheet, like everything else:

```
/templates$ grep -nr "style="
new_invitation_page.php:142:                                    <label class="invalid" id="message_can_not_contain_urls" style="display:none;">{tr:message_can_not_contain_urls}</label>
download_page.php:365:                    <col style="width:25%">
download_page.php:366:                    <col style="width:75%">
```

Fixing this will allow us to remove unsafe-inline from CSP

monkeyiq:
Removing unsafe-inline is a good thing! This would be two steps back at the moment.

monkeyiq:
@madsi1m let me know if there are other CSP fixes that should be in 3.x.

monkeyiq:
I think I will close this issue. If other CSP updates come up please create new issues so I can make sure we can take advantage of the added security of removing unsafe-inline.

PR #1977, opened 2024-08-22
fix these to allow removing unsafe-inline from CSP

monkeyiq:
This is in response to https://github.com/filesender/filesender/issues/1976

Being able to remove unsafe-inline from CSP is very useful!


madsi1m:
nice naming :)

PR #1982, opened 2024-08-28
Update docs for SSP 2.x use

monkeyiq:
We should be using SSP 2.x these days. The hash extractor script is updated and many parts of the docs. A clean install into a VM would be useful to try to catch any outstanding issues.


monkeyiq:
I updated docs/v3.0/install/index.md first and made a patch to apply to docs/v2.0/install/index.md and committed that. This seems like some sort of way to handle the two docs trees issue.

PR #2025, opened 2024-10-01
Fix typo in translate string

WebSpider:
Minor typo causing missing translations

### DIFF
--- a/templates/user_page.php
+++ b/templates/user_page.php
@@ -228,7 +228,7 @@ $user = Auth::user();
                                     </li>
                                     <li>
                                         <span>
-                                            {tr:clear_stores_recipients_emails}
+                                            {tr:clear_stored_recipients_emails}
                                         </span>
                                     </li>
                                 </ul>


madsi1m:
@monkeyiq @WebSpider can we update the en_AU to also use `clear_stored_recipients_emails` rather than `clear_stores_recipients_emails` ? Might want to also do pt_BR

madsi1m:
```
#grep -nr clear_stores_recipients_emails
language/pt_BR/lang.php:118:$lang['clear_stores_recipients_emails'] = 'Limpe os e-mails de destinatários armazenados do uso anterior (usado para preencher automaticamente os campos de e-mail).';
language/en_AU/lang.php:119:$lang['clear_stores_recipients_emails'] = 'Clear stored recipients emails from past use (used to automatically complete email fields).';
```

madsi1m:
If i remember correctly the lang files are auto generated yet it still has the old name:
https://github.com/filesender/filesender/blob/4b7e47df6dcbe91a74b329790cd996ec3025c274/language/en_AU/lang.php#L119

monkeyiq:
The json file exported from poeditor contains a blank "" translation for items that have not been entered into poeditor for a language. We could then I guess lookup and import the english baseline term for such terms. The import script currently ignores terms that have no translation so it was skipping `clear_stored_recipients_emails`. I copied the english translation to the en_au on poeditor and now it is importing as expected.

See PR https://github.com/filesender/filesender/pull/2040

Thank you for bringing the issue to my attention.

WebSpider:
We can reconfigure poeditor to send the base language value in case something isnt translated as well, if desired

monkeyiq:
I updated there, a new PR is on the way. It will likely be much less surprising to translators / devops.

monkeyiq:
That becomes quite large... should be a smaller change set over time.

https://github.com/filesender/filesender/pull/2041

PR #2036, opened 2024-10-09
Added the option for upload complete popups.

PaulLocksley:
Hi Team, 

We've noticed more users raising support requests unsure if their upload was complete and wanting to double check with our team. 
To resolve this we wanted the option for a more aggressive confirmation for the users.

This PR adds a new transfer option called 'popup_on_complete'  following the examples of 'WEB_NOTIFICATION_WHEN_UPLOAD_IS_COMPLETE' and 'redirect_url_on_complete'. 

When enabled it provides a notification using the standard language as follows:
image

The admin simply needs to add the configuration option to their transfer options like they would any other.

### DIFF
--- a/classes/constants/TransferOptions.class.php
+++ b/classes/constants/TransferOptions.class.php
@@ -52,7 +52,8 @@ class TransferOptions extends Enum
     const HIDE_SENDER_EMAIL                         = 'hide_sender_email';
     
     const REDIRECT_URL_ON_COMPLETE                  = 'redirect_url_on_complete';
-
+    const POPUP_ON_COMPLETE                         = 'popup_on_complete';
+    
     const ENCRYPTION                                = 'encryption';
     const COLLECTION                                = 'collection';
     const MUST_BE_LOGGED_IN_TO_DOWNLOAD             = 'must_be_logged_in_to_download';
--- a/includes/ConfigDefaults.php
+++ b/includes/ConfigDefaults.php
@@ -280,7 +280,7 @@ $default = array(
                                                                 , 'email_daily_statistics', 'email_report_on_closing'
                                                                 , 'enable_recipient_email_download_complete'
                                                                 , 'add_me_to_recipients', 'redirect_url_on_complete'
-                                                                , 'hide_sender_email'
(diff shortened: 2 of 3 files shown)


madsi1m:
nice work!

monkeyiq:
I was in the process of taking a look when this got merged. AFAIK there are some other UI updates happening for the 3.x release. 

A few thoughts:

* I would recommend an entry in the transfer_options docs page with some information
https://github.com/filesender/filesender/blob/2b83eea69962c2a80af0f8bb465eb3422efbfad0/docs/v3.0/admin/configuration/index.md?plain=1#L1828

* If the option is shown you see `{popup_on_complete}` so either documenting not to show the option or adding a new translation.

* I haven't considered how this interacts with the guest upload options. If the purpose was to only add a dialog to the upload page and not allow it as a guest option then further investigation might not be needed.

monkeyiq (commented):
I might put this in here as the review part rather than just as a comment.

A few thoughts:

*  I would recommend an entry in the transfer_options docs page with some information at docs/v3.0/admin/configuration/index.md

* If the option is shown you see {popup_on_complete} so either documenting not to show the option or adding a new translation.

* I haven't considered how this interacts with the guest upload options. If the purpose was to only add a dialog to the upload page and not allow it as a guest option then further investigation might not be needed.

PaulLocksley:
Thanks for looking:

- I would recommend an entry in the transfer_options docs page with some information at docs/v3.0/admin/configuration/index.md
    - Agreed, I will add some documentation here. 
-   If the option is shown you see {popup_on_complete} so either documenting not to show the option or adding a new translation.
    - I didn't consider that an admin may configure it to be shown, I don't think it would make sense as a user option so I will mention it in the documentation above. 
-   I haven't considered how this interacts with the guest upload options. If the purpose was to only add a dialog to the upload page and not allow it as a guest option then further investigation might not be needed.
    - Guest uploads also get the same dialog when the option is enabled at the transfer level, I'm don't believe there are any guest options that would conflict.

Please let me know if I misunderstood anything here, I'm happy to make any other changes that you would like.

PR #2037, opened 2024-10-09
Documentation update

PaulLocksley:
Updating documentation to include popup_on_complete in response to discussion in [2036](https://github.com/filesender/filesender/pull/2036 )

You can preview the file [here](https://github.com/PaulLocksley/filesender/blob/upload-alert-documentation/docs/v3.0/admin/configuration/index.md)

Please let me know if you would like anything changed.


monkeyiq:
Thanks for the initial PR and update. If you think the same issue might be on the guests page I could take a look at expanding things to allow this option to be set for guest uploads too.

monkeyiq:
The guest upload path looks fairly similar to the upload path for a normal user. There is 'Transfer completed!' in the middle of the screen but it is also in a position to "blend into the screen". So maybe adding an option for "all done" dialog there too.

monkeyiq:
ok, so with the option in config.php it also makes the dialog for guest uploads already :)

PR #2038, opened 2024-10-09
Bring back language dropdown behaviour from v2.x

### LINKED ISSUE #1997: UI3: My Profile / Language lists langauges 2 or 3 times.
I see two entries for many languages and 3 for Italian. It is a bit confusing which of the entries is the right one to select as they are all the same for the language.

WebSpider:
- Filter duplicate languages
- Only show if there are multiple languages configured

Fixes #1997

### DIFF
--- a/templates/user_page.php
+++ b/templates/user_page.php
@@ -44,7 +44,7 @@ $user = Auth::user();
                     <h2>{tr:user_preferences}</h2>
 
                     <?php
-                    if (Config::get('lang_userpref_enabled')) {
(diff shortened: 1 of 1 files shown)


WebSpider:
Re-reading my own PR made me realise there is a bit missing. Marking as draft.

monkeyiq:
Which bit is missing. It would be handy to know for subsequent updates and/or other PRs that try to do the same thing.

monkeyiq:
I may close this for now. If we know what is missing and what we need to do then this could be reopened.

WebSpider:
I tested this PR and it was fine after all.

Issue #2035, opened 2024-10-08
Filesender 3.0rc3 - Admin page issue


kmihalj:
Well transfers are here, but click on any gives "transfer not found" .... no any info on any transfer except my own, I can get data from database, but preffered is from GUI like in versions 2.xx.

monkeyiq:
I can confirm that this shows 'transfer not found' in `development3` but works in `development`.

madsi1m:
Yup, it seems an admin can open his own transfers and vouchers from the admin page but all others are "not found"

monkeyiq:
From investigation this come in when I tighten up security around viewing transfers. I will create a PR to relax that again for admin users.

madsi1m:
thought i would give vouchers a go https://github.com/filesender/filesender/pull/2051

monkeyiq:
@madsi1m Thanks for the patch for vouchers! I made a slight tweak there and merged voucher support for admin now.

madsi1m:
saw that, seems to work in our test env 👍
Will close this

PR #2047, opened 2024-10-15
ui3: allow admin to view transfers from other users

monkeyiq:
This allows an admin to view transfers from any user. This is useful in the admin pages of FileSender when enabled.

This relates to https://github.com/filesender/filesender/issues/2035


madsi1m:
do you need to do the same on `‎templates/invitation_detail_page.php` ?

monkeyiq:
yep. Things got locked down just a touch to much.

PR #2050, opened 2024-10-17
Fix return path validation

madsi1m:


monkeyiq:
Thanks. I knew rolling all 3 into using the one method call was a bit risky

PR #2059, opened 2024-10-28
Filesender.py enhancements

PaulLocksley:
Hi Team, 

Just two small changes here to make the script more useful for some of our users. 
- Added the option to explicitly set transfer expiry days.
- Added additional logic to not retry when authentication fails.

The core logic for transfer timeout has been here for a while but the users never had an option to manually set it themselves. 
Setting the value to high will generate the relevant log message regardless of verbosity settings for the user. 

Please let me know if you have any questions.

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -360,7 +360,7 @@ def call(method, path, data, content=None, rawContent=None, options={}, tryCount
     if debug:
       print(_exc)
     if tryCount < worker_retries:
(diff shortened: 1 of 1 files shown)


madsi1m:
once merged will need to also go into master3 as that is where users download the client from

madsi1m:
tested, works as expected

madsi1m (approved):
nice

Issue #2081, opened 2024-11-28
Can we store IDP of user in the FileSender DB then filter configs on it?


madsi1m:
Hi, I have been exploring `$config['auth_config_regex_files']` and figure it would be much cleaner and easier if we could just filter on IDPs coming from SimpleSAMLPHP rather than having to learn each IDPs list of email domains.

Looking at `attributes()` function in `classes/auth/AuthSPSaml.class.php` we can query the IDP from SimpleSAMLPHP by doing:
```
$attributes['idp'] = parse_url(trim($ssp->getAuthData('saml:sp:IdP')), PHP_URL_HOST);
Logger::info('User IDP: '.$attributes['idp']);
```

Could we add this as a field in the `Authentications` table? (or similar) This will help us generate graphs per idp which much simpler SQLs.

And could we extend `$config['auth_config_regex_files']` to have an `idp` filter? For example:
```
  $config['auth_config_regex_files'] = [
      'uid' => [
          '@mydomain.com$' => 'mydomainfile',
//          '@myotherdomain.com$|@yetanotherdomain.com$' => 'myotherdomainfile',  //now using IDP filter below
      ],
      'idp' => [
          'idp.myotherdomain.com$' => 'myotherdomainfile',
          'idp.institute2.com$' => 'institute2',
    ],
  ];
```

madsi1m:
edited above comment to also put idp in `$attributes` which as far as i can see will make `$config['auth_config_regex_files']` work

WebSpider:
Isnt this what `auth_sp_additional_attributes` and `auth_sp_save_user_additional_attributes` is for?

madsi1m:
not exactly, as the IDP name/host is not actually an attribute that is returned with user information like "sn" or "department". The info actually comes from the `saml:SP` API

madsi1m:
>Can we store IDP of user in the FileSender DB then filter configs on it?
Seems Yes: https://github.com/filesender/filesender/pull/2082

monkeyiq:
I guess the fix merged label is to early. The IdP stuff is there but the config selection isn't yet merged ;)

PR #2082, opened 2024-12-01
Saml user identification idp

madsi1m:
resolves https://github.com/filesender/filesender/issues/2081
also fixes permissions on some php files


monkeyiq:
It seems in the past there was some desire to obtain the SAML attributes and stash them into a JSON field. Though the docs for auth_sp_save_user_additional_attributes are a little aged. 

https://github.com/filesender/filesender/blob/d31b3b3bc6c788a7a991cfd778ef1836910cbfd6/classes/auth/Auth.class.php#L230

The `additional_attributes` code also has the downside of storing the attributes untreated into a JSON string in the database. Knowing the IdP that the user has authenticated with is an interesting enough piece of information that it should be put into the system as a cleaned up string as in this PR. 

Given that this is cited from https://github.com/filesender/filesender/issues/2081 I should mention that if a config is loaded based on the `saml_user_identification_idp` we will want to have a pass regex on that value to avoid any path tricks to try to load other files and perhaps also put the idp specific config files into an "idp" subdirectory. You have likely done these already in your local code but it doesn't hurt to mention them. The more rechecks on that sort of thing the better.

classes/data/Authentication.class.php
```diff
@@ -60,6 +60,11 @@ class Authentication extends DBObject
             'size' => 200,
             'null' => true
         ),
+        'saml_user_identification_idp' => array(
+            'type' => 'string',
+            'size' => 200,
```
monkeyiq, line 65: The saml_user_identification_uid is only 170 chars because of some mysql index limitation IIRC. Though this is also quite old information and may not be relevant now.

madsi1m: happy for it to be 170 if you prefer

monkeyiq: I added it in https://github.com/filesender/filesender/pull/2096. The database script *should* truncate it down to 170 for you there.

monkeyiq: Seems like a better time to do this now rather than later when/if the issue is reported with one of the storage back ends on mysql/mdb if that still exists in 2025.

Issue #2083, opened 2024-12-10
Bug: Confirm email request upon download being asked while setting is disabled


syphernl:
We upgraded our FileSender to v2.51 but noticed a change. Whenever someone clicks the download link they get the download page and it asks them for email verification to download.
However, clicking "Send" to send the verification code isn't doing anything neither does the "Check" option. Nothing shows up in the console nor the network tab and downloading using the Download button just works.

To add to the weirdness [download_verification_code_enabled](https://docs.filesender.org/filesender/v2.0/admin/configuration/#download_verification_code_enabled) is supposed to be `false` by default. 
Even explicitly adding the following to the config is not working:

```php
$config['download_verification_code_enabled']= false;
```

PR #2084, opened 2024-12-10
fix: only show download verification section if enabled

syphernl:
Fixes https://github.com/filesender/filesender/issues/2083

--

Seems that this issue was more a cosmetic one, the confirmation request section should not be shown if this feature is disabled. 
This PR should fix that.

### DIFF
--- a/templates/download_page.php
+++ b/templates/download_page.php
@@ -36,7 +36,10 @@ function presentAVName( $v )
 }
 
 $rid = 0;
-if( Utilities::isTrue(Config::get('download_verification_code_enabled'))) {
+
+$verificationCodeEnabled =  Utilities::isTrue(Config::get('download_verification_code_enabled'));
+
+if($verificationCodeEnabled) {
     if(array_key_exists('token', $_REQUEST)) {
         $token = $_REQUEST['token'];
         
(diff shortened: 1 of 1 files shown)


monkeyiq:
I am happy to merge this. Though the CSS should make this also hidden by default. I think the release notes should mention that the default.css might have been updated and should be considered for local update.

In www/css/default.css there should be something like the following. 
```
.verify_email_to_download {
    width: 50%;
    display: none;
}
```

If I change the class of `verify_email_to_download` in templates/download_page.php I can see those elements. However if it remains `verify_email_to_download` then they do not show up for me. Using these hidden by default classes allows the html page to have fewer php if blocks and less chance of not matching them up.

WebSpider:
>I think the release notes should mention that the default.css might have been updated and should be considered for local update.
> 

Is `default.css` override-able somewhere, like translations and templates are?

monkeyiq:
I dont think there is a default local default.css file in 2.x

monkeyiq:
Please reopen this request if updating the css to the one that came with the release didn't fix things. If you are changing default.css I am happy to have a new feature request to allow a local.css to also be served if it exists on the server.

PR #2086, opened 2024-12-17
Fix: add empty file error to cli

PaulLocksley:
Hi Team, 

Just a small change to add the empty file protection that exists in the webui into the cli. 
Currently these transfers just slowly fail, early warning makes it a little nicer for users.

I tested this with both empty files, normal files and a mix passed as a directory and had the expected behavior.
Please let me know if you have any questions.


monkeyiq:
Looks good. Thank you for the contribution :)

Issue #2091, opened 2025-01-08
symfony/process might need updating (CVE-2024-51736)


madsi1m:
Hi, our security scan has detected that FS is on symfony/process v7.0.3 which has CVE-2024-51736.

Apparently fixed in these versions:
5.3.0, 6.1.0, 5.0.0, 5.2.0, 5.4.46, 6.2.0, 6.3.0, 4.0.0, 5.1.0, 6.4.0, 7.1.0, 7.1.7, 3.0.0, 5.4.0, 6.4.14

Perhaps we can update it to 7.1.7?

monkeyiq:
I think the code formatters etc really should be split into another composer for dev installs. I'll take a look at that more after the vacation.

https://github.com/filesender/filesender/pull/2092

monkeyiq:
I updated the install docs with
```
composer install --no-dev
```
so that only non dev deps are installed by default. I have updated the composer.json in d+d3 now. If this dep is still at large please let me know and I'll update/remove it.

WebSpider:
> so that only non dev deps are installed by default. 

Small question for the container builds that are derived from this: Are the 'non dev deps' designed to be unneeded for production workloads?

monkeyiq:
Yes. The  'non dev deps' shouldn't be needed for production. Please let me know if I have trimmed too much here and need to add something back.

madsi1m:
seems to be fixed, closing

PR #2092, opened 2025-01-11
move some deps to require-dev

monkeyiq:
It seems you need to use --no-dev to exclude these. Perhaps dev stuff might go into a different composer.json file instead...

```
composer  install --no-dev
```

This was referenced here https://github.com/filesender/filesender/issues/2091


monkeyiq:
This is superseded by https://github.com/filesender/filesender/pull/2097

PR #2097, opened 2025-02-03
Move development only packages to requre-dev & docs update

monkeyiq:
The php-cs-fixer is only needed for the rare reforatting of code. That was seen as something needed at some stage. At any rate that should only need to be installed on a development machine.

The sauce/sausage package is not used now. The selenium tests are run through codeception. codeception can handle doing remote code runs.

The sauce/sausage had become untenable as it forced older versions of phpunit which can not be used in the GHA environment in the modern day.

The docs are updated as composer will install all items by default but in an installation env we only want the --no-dev packages installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,18 +1,11 @@
 {
     "require": {
-        "sauce/sausage": ">=0.15.2",
-        "appium/php-client": "@dev",
+        "ramsey/uuid": "^4.7"
+    },
+    "require-dev": {
         "phpunit/phpunit-selenium": "^9.0.1",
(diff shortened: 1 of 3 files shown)


monkeyiq:
I will make an update to development3 soon.

monkeyiq:
Other than the very tedious explicit mirroring of changes to the v2.0 and v3.0 docs tree the composer.json update is brought in with the functional merge and conflict resolution. The docs are explicitly updated for v3.0 here https://github.com/filesender/filesender/pull/2098.

PR #2100, opened 2025-02-03
fix: Be nicer when deletions fail in cron

WebSpider:
There are moments when even forcibly deleting causes failures, and causes cron to abort.
This adds a small handler in these cases. You are not 100% sure the file is deleted, but you have a warning in your logs and cron finishes.

### DIFF
--- a/scripts/task/cron.php
+++ b/scripts/task/cron.php
@@ -114,7 +114,11 @@ if( $verbose ) echo "cron.php delete failed transfers...\n";
 foreach(Transfer::allFailed() as $transfer) {
     Logger::info($transfer.' failed, deleting it');
     if( $force ) {
-        $transfer->deleteForce = true;
+       try {
+            $transfer->deleteForce = true;
+       } catch (Exception $e) {
+               Logger::warn("Deleting transfer failed. error:" . $e->getMessage());
+       }
     }
     $transfer->delete();
 }


WebSpider:
@monkeyiq could you tell me what feedback you need?

monkeyiq:
> @monkeyiq could you tell me what feedback you need?

It seems my browser is saying I posted but when I view on other browser I do not see the comment... ? I don't know how the gh web interface got wedged like that... the original comment is/was:

We want the try/catch around the delete() call as that should be what throws the exception? 

When `deleteForce` is set then failures in `removeFile` should only result in a log in cron. 

https://github.com/filesender/filesender/blob/8f617cc46076565d80a5c76fa4c424b37ae411f6/classes/data/Transfer.class.php#L724

Looking at the beforeDelete if there are errors deleting auditlogs in the db or translatable emails then it might be still throwing. Maybe expanding the beforeDelete method to be more robust when deleteForce is set? What sort of cron errors are you seeing that cause halts?

madsi1m:
Any reason to have a try catch around setting a bool in `$transfer->deleteForce = true;`

I think what you might want is what is already in the Transfer class

```
            try {
                $this->removeFile($file);
            } catch (Exception $e) {
                if( $this->deleteForce ) {
                    Logger::warn("Transfer::delete() Failed to delete file error:" . $e->getMessage());
                } else {
                    throw $e;
                }
            }
```

WebSpider:
The use-case here is when a cron-run had troubles deleting the complete transfer while storage is on S3.

Scenario is as follows:

- `cron.php` runs and is working on deleting chunks of a file in a transfer
- `cron.php` for any reason is interrupted by the OS (memory, network, whatever)
- `cron.php` then is restarted and will cause an Exception trying to delete the transfer it was working on

This causes cron to effectively stop running, causing all kinds of housekeeping issues.

madsi1m:
I'm starting to get the feeling the cron might need a rethink, some improvements off the top of my head:
- split the job types into its own cron jobs, eg, sending reminder emails should be be dependant on transfers expiring with successful file deletes.
- delete files, before closing a transfer, this should stop having leftover old orphan files

ownCloud have an overly complicated way of running cron jobs, perhaps we can get some inspiration from them? They used to have a jobs table in the db and the cron job would update when each job last ran and if it was successful etc etc. From memory you could also trigger particular jobs from cron if you want, so in our case we would run a cron just for expiring transfers, and another to do all the other things.

monkeyiq:
There are also some frameworks that allow async jobs, can track them, can restart them if they failed etc. 

I am guessing we will move to reimplementing that sort of functionality from the ground up.

WebSpider:
Yes, there are async frameworks like this, but it this really necessary? 
I agree that cron needs a bit of tlc to become more robust.

Implementing some form of checkpointing within longer running tasks and allowing tasks to run at various intervals gets us a long way, where async seems to make more sense for longer running processes, which spawn tasks of their own.

monkeyiq:
I have made an issue to further discuss. If this PR is still desired please reopen.

WebSpider:
I'm fine with further discussion. For now, this PR is keeping at least the cronjobs of one install going :-)

monkeyiq:
I am happy to merge it if it is helping you.

PR #2099, opened 2025-02-03
feat: Updated Statistics page & added "Tenant Admin" user role

madsi1m:
# In this PR
- Added a role called tenant_admin which is a normal user but can view statistics of the IDP the user belongs to. Can be added via idp filter in `auth_config_regex_files` by setting `tenant_admin`
- Uplift Statistics page to follow new UI and be more useful to Admins and Tenant Admins for reporting.
- Tenant Admins can access the Statistics page through the web UI and view stats related to the IDP they are a tenant admin of.

Sample config for an IDP:
config.php:
```
<?php
...
$config['auth_config_regex_files'] = [
  'idp' => [
    'FileSender test IDP A' => 'FS-test-IDP-A',
  ]
```
config-FS-test-IDP-A.php:
```
<?php

//Some random config to make this idp have different rules to our defaults
$config['default_transfer_days_valid'] = 1;
$config['max_transfer_days_valid'] = 5;
$config['max_guest_days_valid'] = $config['max_transfer_days_valid'];

//The tenant admin(s) of this idp
$config['tenant_admin'] = 'tenantadmin@test.edu.au';
```

Global Statistics also have new rows, 

## New Graphs, looks back 30 days:
- Transfers and Guests/Vouchers per day
- Data transferred per day
- Transfer Speeds/s per day
- Encryption Split of end to end encryption used vs not.

## New Tables
- Transfers per user
- Transfers per user including expired
- File/Mime types of files being transfered
- Users who have created API keys

The tables use AJAX to change page and does not require full page reloads.

Note: Screenshots taken from a test instance of FileSender and are not as data rich as a production environment.

## As an admin
image

- Admins can see what a tenant admin would see by selecting an IDP from the IDP drop down box, or can select All to see everything.

## As a tenant admin
image

- Tenant admins can only see statistics of the IDP they belong to only!

This is a very functional first attempt of what our users/customers have been asking for. It would be good for the community to use and expand/improve on this. In the past we used to present this in a grafana dashboard and send screenshots to customers. With this, customers can simply view the information as required.

## Note on language changes
As the new statistics page introduces many new things the following needs to be added to the language files. I have reused as much as possible in the existing file but the following is required to be added. Here is what i have used :
```
$lang['user_count_estimate'] = 'Users';
$lang['recipient_count_estimate'] = 'Unique Recipients';
$lang['guest_count_estimate'] = 'Guest Vouchers';
$lang['user_aup_count_estimate'] = 'Users Accepted T&Cs';
[...]

PR #2105, opened 2025-02-10
Move Guest count to using views and joining into auth table.

monkeyiq:
I mentioned this on the PR
https://github.com/filesender/filesender/pull/2099 I thought it might be simpler to add these updates after the merge.

I likely have more familiarity with the database schema and view setup code that most people will want to have :)


madsi1m:
nice

PR #2108, opened 2025-02-10
Statistics speed (add transferssizeidpview)

madsi1m:
add/use transferssizeidpview to speed things up


madsi1m:
@monkeyiq the "Browser stats" table is super slow to load when statslog is huge, i have distant memory of discussions to make this faster somehow, have these been implemented (it is possible i am thinking of the speed graph instead)

PR #2118, opened 2025-02-13
IDP in path on disk

madsi1m:


monkeyiq:
At first look this seems ok. 

I dont recall off the top of my head how migration was handled for these updates and will have to play around a bit in an A/B to verify things anyway.

madsi1m:
From memory we manually moved files into the days folders when we enabled that.

Will this change break the cleanup jobs that cron runs? I guess we do not want to ever remove the idp subdirs but normal cleanup should still happen.

madsi1m:
Seems moving the var into the db like the day buckets allows us to make only new transfers go in the idp. I noticed the hashing option doesn't work like this, perhaps it also should?

madsi1m:
my testing seems to work well with latest commit

PR #2119, opened 2025-02-15
docs: warning about migration

monkeyiq:
If this new setting is changed then migration will have to happen manually if old files are to be discoverable.

This relates to https://github.com/filesender/filesender/pull/2118


monkeyiq:
Some of the other settings mangling paths are also not solid to changes. I am thinking storing the computed subpath in the database is the cleanest way to allow changes to these settings while existing files can live on.

PR #2120, opened 2025-02-16
Possible to save storage subpath in database for easier migration.

monkeyiq:
New option to allow the path from storage to be stored per file in the database. This is the first step towards allowing various settings like per day buckets and idp in the storage path to be changed and existing files continue to be downloadable.

This PR only optionally stores the path. A future PR will allow that information to be used and this allow storage settings to change and existing files (uploaded with this new
storage_filesystem_explicitly_store_subpath_per_file enabled) to continue to be downloaded in the face of settings changes.

This was inspired by https://github.com/filesender/filesender/pull/2119

### DIFF
--- a/classes/data/File.class.php
+++ b/classes/data/File.class.php
@@ -122,7 +122,17 @@ class File extends DBObject
         'have_avresults' => array(
             'type' => 'bool',
             'default' => false
-        )
+        ),
+
+        // used by storage::buildPath to cache the storage path used for this file
+        // this may allow migration of settings in the future with access
+        // to existing files.
+        'storage_path' => array(
+            'type' => 'string',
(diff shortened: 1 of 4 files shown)


madsi1m:
With this, it might just be better to have this forced, and backfill it, then we can remove these from Transfer:
```
        'storage_filesystem_per_day_buckets' => array(
            'type'    => 'bool',
            'null'    => false,
            'default' => false,
        ),
        'storage_filesystem_per_hour_buckets' => array(
            'type'    => 'bool',
            'null'    => false,
            'default' => false,
        ),
        'storage_filesystem_per_idp' => array(
            'type'    => 'bool',
            'null'    => false,
            'default' => false,
        ),
```

Issue #2137, opened 2025-02-21
Installation issue


Sacriyana:
Hello,

I'm following the installation guide of the filesender project inside a docker.
But I'm struggle to have a working simplesamlphp configuration with a local authentication (I don't want use any IdP in my deployement).

I used these links to create my config files:
- https://docs.filesender.org/filesender/v2.0/install/
- https://docs.filesender.org/filesender/v2.0/faq/#simplesamlphp-for-local-users-for-small-scale-setup-or-testing

My docker setup:
- Php: 8.4-apache (`6.8.0-52-generic #53-Ubuntu`)
- Filesender version: 2.52
- SimpleSamlPHP version: 2.36-full

I get the following errors when I tried to access to the simplesamlphp page :

```
Deprecated: Constant E_STRICT is deprecated in /opt/filesender/simplesamlphp-2.3.6/src/SimpleSAML/Error/ErrorHandler.php on line 49

Deprecated: Constant E_STRICT is deprecated in /opt/filesender/simplesamlphp-2.3.6/src/SimpleSAML/Error/ErrorHandler.php on line 49

[...]

PR #2139, opened 2025-02-22
docs: update metadata paths for ssp install

monkeyiq:


monkeyiq:
This was reported here https://github.com/filesender/filesender/issues/2137#issuecomment-2675842119

Issue #2138, opened 2025-02-22
[Docs] Local user installation guide


Sacriyana:
Hello,

My docker setup:

    Php: 8.4-apache (6.8.0-52-generic #53-Ubuntu)
    Filesender version: 2.52
    SimpleSamlPHP version: 2.36-full

I'm trying to setup local users in simplesamlphp using this guide (https://docs.filesender.org/filesender/v2.0/faq/#simplesamlphp-for-local-users-for-small-scale-setup-or-testing).
I have tried authcrypt module, the sqlauth with the documentation available at https://github.com/filesender/filesender/tree/development/scripts/simplesamlphp/passwordverify (I did not applied everything as it is for 1.x ssp version).
I tried to follow this guide : https://simplesamlphp.org/docs/stable/simplesamlphp-reference-idp-hosted without any success on filesender.
But when I tried it throught the test portal on the admin page of simplesamlphp, it's working. 

I have tried to update some fields on authsources.php for the default SP without any success too.

```php
$config = [
        // This is a authentication source which handles admin authentication.
    'admin' => [
        'core:AdminPassword',
        // 'admin:$2y$10$PnFsSEv.lda1Qlw4iMtmB.B.ab5y.aT56stBmo9hdCN.rUywQMChC',
    ],
    // An authentication source which can authenticate against SAML 2.0 IdPs.
    'default-sp' => [
        'saml:SP',

        // The entity ID of this SP.
        'entityID' => 'https://localhost/',
```

Can you details how to get filesender working with simplesamlphp for local IdP ?

There is the error on the https://localhost/simplesaml/module.php/saml/sp/login/example-userpass?ReturnTo=https%3A%2F%2Flocalhost%2Ffilesender%2F%3Fs%3DuploadAuthId=example-userpass&ReturnTo=https%3A%2F%2Flocalhost%2Ffilesender%2F%3Fs%3Dupload :
```
SimpleSAML\Error\Error: UNHANDLEDEXCEPTION
Backtrace:
2 src/SimpleSAML/Error/ExceptionHandler.php:36 (SimpleSAML\Error\ExceptionHandler::customExceptionHandler)
1 vendor/symfony/error-handler/ErrorHandler.php:538 (Symfony\Component\ErrorHandler\ErrorHandler::handleException)
0 [builtin] (N/A)
Caused by: SimpleSAML\Error\Exception: Authsource must be of type saml:SP.
Backtrace:
[...]

PR #2151, opened 2025-02-27
Since PasswordVerify.php is now upstream it is easier again

monkeyiq:
And sspsmall makes setting up the SimpleSAMLphp installation an easier process too.


monkeyiq:
An update prompted by https://github.com/filesender/filesender/issues/2138 

Since PasswordVerify is now upstream there is even less work to get a local FileSender. sspsmall will also help close down some of the gaps.

Issue #1876, opened 2024-04-29
Feature request: Allow upload_chunk_size to change in an existing installation


jmdeboer-surfsara:
**The situation as I understand it:**

In the current implementation of filesender, the upload chunk size is determined at the time of deploying the service by setting the 'upload_chunk_size' option in config.php. If this option is not specified, it defaults to 5 MB.

Once this choice has been made it is not possible to change this without making it impossible to already existing retrieve data from the storage, since the data will not reassemble correctly and decryption will fail.

**Why can this be seen this a problem?**

We are running a filesender instance, using our own S3 service as a storage backend. When the service was deployed, the default was not altered hence data is stored in 5 MB chunks (there is a suggestion in the documentation to use 40 MB for greatly increased performance but this was overlooked). 

Of course there are good reasons to use chunks. Also, one of the limits outlined in the documentation mentions PHP memory usage. But with memory being cheap, larger chunks (like 100MiB) are probably preferred.

However, not being able to change the chunk size once it has been chosen is sub-optimal imho. This leads to unnecessary load on our storage cluster resources, something we now cannot rectify.

**How might this be solved?**

I suggest making upload_chunk_size an attribute of the stored file, with the value for any new uploads being the config value as it is used now.

First of all an int field name upload_chunk_size could be added to the (file?) db table. The migration for this should also make the value of this field the same as the current upload_chunk_size for existing records.

Next, the code that retrieves the data should look at the value of this field for the file being retrieved, and use that for fetching/decrypting the chunks and reassembling the file.

Lastly, the code that does the chunking and encryption for uploads can remain largely the same; it just looks at the global upload_chunk_size, uses that, and stores this value with each file.

This will make it possible to change the chunk size on an existing installation during a maintenance window. Any data will be read with the chunk size that was in effect at the time of storage.

**A bonus:**

Using a larger chunk size (or being able to change it afterwards) comes with an additional benefit if you are not using your own S3 storage, but are using AWS; after all one of the metrics they bill you on is the number of requests. I am sure Jeff would be happy if we all use smaller chunks, that just means more requests Amazon can bill you for...

This are just some initial thoughts and I am probably overlooking things, but I am wondering what your thoughts on the matter are?

monkeyiq:
I think this is a good idea. I would probably target the Transfer table for storing the `upload_chunk_size` used as I think it can just apply to all the files for the transfer.

[...]

PR #2166, opened 2025-03-26
Ability to modify chunk size and retain existing uploads

monkeyiq:
With this you should be able to change the config.php upload_chunk_size (with upload_crypted_chunk_size and download_chunk_size to allow a valid configuration) and still be able to download existing uploads with new uploads performed using the current settings.

This PR has been tested manually and with the new
ConfigurationABCest.php acceptance test. The acceptance test targets both the normal filesystem and filesystemChunked storage_type and varies the chunk size in an A,B,A test with uploads in A, downloads in B then uploads in B and downloads in A for two configurations. These automatic tests check for unencrypted, encrypted, and encrypted to zip transfer scenarios. See testConfigurationAB for details.

Some testing has also been performed using a fake S3 backend and chunk size changes. As in the past I find myself with no credentials to a real S3 service so can only test with these environments. For this I have used a zenko/cloudserver server.

The Azure storage was updated inline with the S3 one but has not been extensively tested for migrations.

The database.php script creates new columns in the Transfer table and sets their values to the current upload_chunk_size and upload_crypted_chunk_size values which should allow existing uploads to work properly after the upgrade.

This relates to https://github.com/filesender/filesender/issues/1876


WebSpider:
323 commits?

monkeyiq:
something went wrong here. The patch is more like 50kb and around 20 files. I'll make a new one tomorrow.

monkeyiq:
Maybe something went off the track with the merge dev3-chunks into the PR. I am not too sure dev3 stuff can be merged back into the dev branch.

madsi1m:
apologies, not sure what happen exactly i merged monkeyiq:2025/march/chunk-size-vary to madsi1m:/dev3-chunks, looks like it went backwards?

madsi1m:
I wonder if the resolving conflicts step broke it...
either way, apologies

monkeyiq:
I was sort of wondering how it came about. Mostly to avoid accidentally doing a similar thing myself at some stage. 

My best guess is that the dev3 thing got merged into the branch in my fork. Though if that came about because git somehow thought that was related to some other push that happened to another fork and moved some objects over or updated this fork as well.

monkeyiq:
In other news I have a dev3 port started at https://github.com/monkeyiq/filesender/tree/2025/march/chunk-size-vary-d3

seems to work with crypto for upload and download + download zip. No rigorous testing done at all as yet on the dev3 branch.

PR #2167, opened 2025-03-26
updated chunk vary

monkeyiq:
With this you should be able to change the config.php upload_chunk_size (with upload_crypted_chunk_size and download_chunk_size to allow a valid configuration) and still be able to download existing uploads with new uploads performed using the current settings.

This PR has been tested manually and with the new
ConfigurationABCest.php acceptance test. The acceptance test targets both the normal filesystem and filesystemChunked storage_type and varies the chunk size in an A,B,A test with uploads in A, downloads in B then uploads in B and downloads in A for two configurations. These automatic tests check for unencrypted, encrypted, and encrypted to zip transfer scenarios. See testConfigurationAB for details.

Some testing has also been performed using a fake S3 backend and chunk size changes. As in the past I find myself with no credentials to a real S3 service so can only test with these environments. For this I have used a zenko/cloudserver server.

The Azure storage was updated inline with the S3 one but has not been extensively tested for migrations.

The database.php script creates new columns in the Transfer table and sets their values to the current upload_chunk_size and upload_crypted_chunk_size values which should allow existing uploads to work properly after the upgrade.

This relates to https://github.com/filesender/filesender/issues/1876

### DIFF
--- a/classes/data/File.class.php
+++ b/classes/data/File.class.php
@@ -267,6 +267,7 @@ class File extends DBObject
     private function calculateEncryptedFileSize()
     {
         $upload_chunk_size = Config::get('upload_chunk_size');
+
         
         $echunkdiff = Config::get('upload_crypted_chunk_size') - $upload_chunk_size;
         $chunksMinusOne = ceil($this->size / $upload_chunk_size)-1;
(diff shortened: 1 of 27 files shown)


monkeyiq:
I have a dev3 port started at https://github.com/monkeyiq/filesender/tree/2025/march/chunk-size-vary-d3

It seems to work with crypto for upload and download + download zip. No rigorous testing done at all as yet on the dev3 branch other than that first test. Subsequent testing is needed starting with varying the chunk size and validating that existing files can indeed still be downloaded.

monkeyiq:
The heavy lifting for the d3 port should be contained in this PR itself. There may be UI elements which are not treated as they should be which may not be passing the correct values to the front end for decryption of existing uploads in the face of chunk size alterations.

PR #2173, opened 2025-04-01
Fix race condition in non-terasender transfers where make $transfer->makeAvailable() is called before all files have completed transfer

pengufrost:


www/js/transfer.js
```diff
@@ -1078,12 +1079,14 @@ window.filesender.transfer = function() {
...
      * @param object file
-     * @param callable complete is file done
+     * @param bool complete is file done
      */
     this.reportProgress = function(file, complete) {
         var now = (new Date()).getTime();
         
         file.progress_reported = now;
+
+        if(!complete) complete = false;
```
Razorfang, line 1089: Can you explain this line? It's basically saying _if complete is not true, set complete to false_ . In what cases would this be not true but also not false?

pengufrost: Was there as my thinking was that it's not always passed through so I wanted to explicitly set it to false, though removing this line has no negative effect so I'll remove it

monkeyiq:
It seems that the core issue would seem to be that  `was_last_file` is based on the calculation  `transfer.file_index >= transfer.files.length`.

It should be based on counting the number of files to complete and only trigger when `completed_count >= transfer.files.length`

It may be simpler to resolve that issue first and refactor things as a secondary PR?

PR #2175, opened 2025-04-01
Add visual cue for users to fill in recipients

WebSpider:
Feedback we got from users is that it wasn't clear to everyone they needed to add a recipient address.
This adds a visual cue that the field is in fact mandatory.

The term has been added to FS3.0 translations, so should be importable.

Possibly this is not the way to do this in the new UI. I'm not a frontend guy, so if this needs to be done differently, by all means :grin:


monkeyiq:
In other sites I have seen things like red borders if a field has not been set but must be set.

On the one hand I imagine this is currently as RNP wishes it to be. 

On the other hand things like the encryption password are already red by default to indicate that they require attention.

monkeyiq:
It is difficult that most UI related conversions have not taken place.

madsi1m:
I to prefer the red borders, the javascript already blocks the send/next button with this field not filled in.

WebSpider:
> I to prefer the red borders, the javascript already blocks the send/next button with this field not filled in.

Unfortunately, my frontend skills are not that good. If you want to make it a red border, i guess thats more in line with the overall style of the new UI

Issue #2178, opened 2025-04-02
v3: Button inconsistency in transfer details


WebSpider:
In v3.0rc7:

When attempting to add recipients in a non-English language, the modal popping up has 2 inconsistencies:

- The buttons are not translated
- The cancel button is not colored red

![Image](https://github.com/user-attachments/assets/09f2b302-84bf-4b22-863c-9a2d3f0ac0a1)

monkeyiq:
On the transfer details page itself there is bare text

> Your transfer has been sent to the following email addresses

and

> Add recipient

monkeyiq:
I moved my above finding to a new issue https://github.com/filesender/filesender/issues/2187

This issue itself should be fixed with https://github.com/filesender/filesender/pull/2186

PR #2186, opened 2025-04-09
Improve transfer details page for add recipient

monkeyiq:
This makes the buttons the right colour, translates them, and makes an email address required in order to positively process the dialog.


monkeyiq:
This handles the dialog part of https://github.com/filesender/filesender/issues/2178

PR #2190, opened 2025-04-09
UI3: "Send transfer to" use red border rather than * mandatory

madsi1m:
I don't like https://github.com/filesender/filesender/pull/2175 so made it use red border

Also fixed a typeO on the "from" input line that was labeled as "to"


WebSpider (approved):
Lgtm 👍

PR #2191, opened 2025-04-10
UI3: Add option cli_client_from_github

madsi1m:
Add option cli_client_from_github + allow config to also use local python script


monkeyiq:
Nice to keep things on the same domain. Good for it to be an option. Thank you for the PR!

PR #2193, opened 2025-04-13
i18n: ui3: auto import from poedtor on 2025-04-14-1744586131

monkeyiq:


madsi1m:
you can prob remove $lang['mandatory'] now

WebSpider:
I've removed Mandatory from the list of terms in POEditor, it should be cleaned up automatically upon next import.

PR #2196, opened 2025-04-17
PGP encryption option for transmitting passphrase used to encrypt transfer

monkeyiq:
This is the first level of support for PGP encryption. It supports one simple work flow: allowing guests to send files to the person who invited them to the system without needing to work out how to transmit the passphrase.

This works as follows

* A regular user of the system visits the `My Profile` page and uploads their PGP public key
* Optional: On that page they can click a button to send a test message to themselves encrypted with that key. In their email program they can then see that decryption is working and are happy with that part.
* The user then invites a guest to the system
* The guest uploads a file and does not need to enter a pass phrase (it is generated) and the pass phrase is encrypted with PGP and will be sent as the transfer "message". All of this happens in the browser, the server only ever sees the encrypted PGP message.
*  The user sees an email that their guest  has completed an upload including an encrypted block with the pass phrase.
* The user then visits the download link and uses PGP in their email program to decrypt that message to get the pass phrase. They enter that in the FileSender UI and download the file.

Easy simple uploads for the guest for the price of uploading your PGP public key to the system.

TODO

* add more docs
* revalidate usage when the pgp option is set to off.
* revalidate usage paths where no key is uploaded.


monkeyiq:
I have done some initial A/B testing with this disabled, enabled without a key, and enabled with a key. Still more testing to do.

PR #2201, opened 2025-04-22
Add download functionality to filesender.py

PaulLocksley:
Hey team, we had some users request the ability to easily download files in filesender.py. 

It doesn't require a configuration to be downloaded so it's a bit easier to use.
Currently this supports serial downloading of normal transfers and encrypted transfers if the server is using aes-gcm. 

### Changes:
- Enable anonymous access to the extended file transfer endpoint to match https://github.com/filesender/filesender/pull/2165. to enabled sub directories correctly and encrypted transfer downloads.
- Change in required argument parsing in the script to enable either the previous upload parameters or a new -d download parameter. 
- Adds a -d argument to filesender.py that takes in a download link for any file sender server and saves the contents of the transfer locally.

### Usage:
`python filesender.py -d 'https://filesenderserver/?s=download&token=4127c216-1d48-4d7b-9f41-8e7f19c0674f' -pv`
It will save the transfer in a directory under the token name by default, however you can also specify a folder with the -o argument. 
if the transfer was encrypted you can also pass in the -e argument with the password.

### Testing:
I've tested uploading and downloading large encrypted files, regular files and a large number of files. 
All worked without issue and all file hashes matched.

There is some room for performance improvement as there is no parallel workers like in the upload logic but overall it works fairly well in my testing.

Please let me know if you have any issues with this work.

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -541,7 +541,7 @@ def downloadFile(token,file_info:dict,download_key:bytes|None, attempt:int=0):
 def _downloadFile(token,file_info:dict, download_url:str, local_file_path:str):
   """save file to disk at local path"""
 
-  with requests.get(download_url,params={"token":token,"files_ids":file_info['id']},stream=True) as download:
+  with requests.get(download_url,params={"token":token,"files_ids":file_info['id']},stream=True,verify=(not insecure)) as download:
         download.raise_for_status()
         with open(local_file_path, mode='wb') as f:
             for chunk in download.iter_content(chunk_size=20_000):
@@ -559,7 +559,7 @@ def _downloadEncryptedFile(token,file_info:dict, download_url:str, local_file_pa
       end_offset = min(1 * (chunk_no * upload_chunk_size + (1*upload_crypted_chunk_size)-1),
                        (1*file_info['size']) + 32 - 1)
       download = requests.get(download_url,params={"token":token,"files_ids":file_info['id']},
-                              headers={ "Range": f"bytes={i}-{end_offset}"})
+                              headers={ "Range": f"bytes={i}-{end_offset}"},verify=(not insecure))
       download.raise_for_status()
       f.write(decrypt_chunk(download.content,chunk_no,file_info,download_key))
       chunk_no += 1
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -515,21 +515,28 @@ def getFilesInTransfer(transfer_token) -> list[dict]:
     {},
   )
 
-def downloadFile(token,file_info:dict,download_key:bytes|None):
+def downloadFile(token,file_info:dict,download_key:bytes|None, attempt:int=0):
   """Download a given file to disk."""
-  download_url = base_url.replace("rest.php","download.php")
-  path = file_info['name'].split("/")
-  download_file_name = path[-1]
-  prefix = token
-  if download_folder is not None:
-    prefix = download_folder
-  path = os.path.join(prefix,*path[:-1])
-  Path(path).mkdir(parents=True, exist_ok=True)
-  local_file_path = os.path.join(path, download_file_name)
+  try:
+    if attempt > 10:
+      print("Unable to donwload file.")
+      sys.exit(1)
+    download_url = base_url.replace("rest.php","download.php")
+    path = file_info['name'].split("/")
+    download_file_name = path[-1]
+    prefix = token
+    if download_folder is not None:
+      prefix = download_folder
+    path = os.path.join(prefix,*path[:-1])
+    Path(path).mkdir(parents=True, exist_ok=True)
+    local_file_path = os.path.join(path, download_file_name)
 
-  if file_info["encrypted"]:
-    return _downloadEncryptedFile(token,file_info,download_url,local_file_path,download_key)
-  return _downloadFile(token,file_info,download_url,local_file_path)
+    if file_info["encrypted"]:
+      return _downloadEncryptedFile(token,file_info,download_url,local_file_path,download_key)
+    return _downloadFile(token,file_info,download_url,local_file_path)
+  except Exception as e:
+    print(f"Retrying on file {file_info['name']}")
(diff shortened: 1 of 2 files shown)


scripts/client/filesender.py
```diff
@@ -167,65 +174,85 @@
...
+  if not base_url:
+    missing_fields.append("-b, --base_url")
+  if not username:
+    missing_fields.append("-u, --username")
+  if not recipients:
+    missing_fields.append("-r, --recipients")
+  if not arg_files:
+    missing_fields.append("[FILE]")
+
+  print(f"Missing required paramter, please provide the following: \n {','.join(missing_fields)}\n or -d --download with a valid download link")
```
WebSpider, line 190: There is a typo here

PaulLocksley: Fixed :)

monkeyiq:
Thank you for this contribution! 

I now have a working tree that allows the `Auth::isAuthenticated()` to stand and will function. I have not tested it with an encrypted transfer as yet and might need a slight tweak. I will clean up my tree and start sharing back and/or push an update back to the base branch.

I tend to avoid things that loosen the auth or consider if there are ways around it. One example is that my upload has the option "User must login to FileSender to download file(s)" enabled. For me this means that you shouldn't be able to grab this metadata from the REST API without login. If I visit the download page link with the token in a browser it kindly asks me to login to proceed. 

If the REST client can form a signature with a user secret then it is somewhat equivalent to being "logged in".

PaulLocksley:
I didn't think of that setting. Will it still enable complete guest users with just a link to download? 
That was my main thought behind removing it. 

If not, maybe conditionally adding that requirement if you have "User must login" enabled would be a good option?

monkeyiq:
I have always tended to think of the REST API as something that is targeted at registered users. Getting a "my profile / api secret" is only possible if you can login. 

It makes sense that guests might like to use the REST clients too but I haven't really had that in my frame of mind. It is always better to err on the side of caution so I think for that API call I just assumed you should be able to login. (even without the "must login" transfer setting being considered).

On the other hand actions like following a link with token to a download page on the web and using the filesender.py to download the same transfer should be possible if you are a guest. Though that scenario isn't something I have tested on.

monkeyiq:
These changes allow it to run as a logged in user with the auth check in place.

https://github.com/filesender/filesender/compare/de8e8f12...monkeyiq:filesender:pr-2201

The rstrip is needed as the `signature` is over the URL string and if you get a double `/` in there it will make the signature check fail on the server side.

We may loosen the restriction on this call to allow the client to work as a guest. Assuming the "must be logged in" setting is `False`. I'll have to do a little more testing on that case. Sorry about the delay.

monkeyiq:
If you remove the ini file and run with `-v` then you get the following. 

```
    print('username          : '+username)
          ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
TypeError: can only concatenate str (not "NoneType") to str
```

Without with config and without `-v` and without the auth check for the getfileinfo then it runs ok.

monkeyiq:
Interesting, if we want to selectively deny access to metadata if "must be logged in" is set and you are not logged in then you have to have signed the request. I guess here the filesender.py could sign the `fileidsextended` if it has an api secret and then it would pass even if the must be logged in is set.

monkeyiq:
I have issues downloading encrypted files. This might be because of no requiring a configuration to be downloaded so the settings do not match up?

```
./filesender.py --insecure  -v -d 'https://fqdn/filesender/?s=download&token=8...6c' -e password  -v 
```

monkeyiq:
It might be just my env but it seems I have to cast the upload_chunk_size to an int for encrypted download to work...?

```
diff --git a/scripts/client/filesender.py b/scripts/client/filesender.py
index 8ef414ab..fd32bd81 100755
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -221,7 +221,7 @@ if base_url != '[base_url]':
       info_response = requests.get(base_url+'/info', verify=False)
       config_response = requests.get(base_url[1:-9]+'/filesender-config.js.php',verify=False)
 
-  upload_chunk_size = info_response.json()['upload_chunk_size']
+  upload_chunk_size = int(info_response.json()['upload_chunk_size'])
 
   try:
       regex_match = re.search(r"terasender_worker_count\D*(\d+)",config_response.text)
@@ -681,7 +681,7 @@ def download_transfer(download_link):
     encryption_details['salt'] = file_list[0]['key-salt'].encode('ascii')
     encryption_details['password_hash_iterations'] = file_list[0]['password-hash-iterations']
     info_response = requests.get(base_url+"/info")
-    globals()['upload_chunk_size'] = info_response.json()['upload_chunk_size']
+    globals()['upload_chunk_size'] = int(info_response.json()['upload_chunk_size'])
     globals()['upload_crypted_chunk_size'] = upload_chunk_size + 32
     globals()["encryption_details"] = encryption_details
     download_key = generate_key()
```

madsi1m:
wouldn't hurt to cast all the ints to ints from the json response

PaulLocksley:
Weird, I wonder why the data types are inconsistent. I have noticed that on some rest endpoints it renders as a string instead of an int, haven't seen it on the same endpoint based on config before but useful to know.

I see you have --insecure that should also cause issues. I'll got a long weekend but I'll aim to address all these on Tuesday.

monkeyiq:
No worries. I thought I'd dig into using it with encryption to see if things worked ok with the "need to login" check I was playing with. I guess that check should be cross ported to the updated made in pr 2165.

monkeyiq:
I have `--insecure` set because I just rolled some self signed certs and things complain there. I also used to have the env variable set to tell python where to find the CA stuff to not complain.

madsi1m:
maybe the config.php is different on both??

PaulLocksley:
Morning, I've edited the script to fix the issues mentioned above and removed the guest download ability for now.
I did some light testing and everything appears good still, can you let me know if that fixed your issue with parsing configuration?

monkeyiq:
I can download fine with these changes. Thank you for the PR and the subsequent updates! :)

PR #2205, opened 2025-04-25
Update the docu for SimpleSAMLphp example-sp

baumrasen:
I had a lot of trouble get the current version of SimpleSAMLphp working with a local hosted SP. So I updated the documentation for that, so that others will have faster a working solution.


monkeyiq:
Thank you for the update!

PR #2214, opened 2025-04-29
allow guests to click "send" to upload

madsi1m:


madsi1m:
Not sure how, but guest/voucher uploads stopped with recent changes, this fixes it, not sure if it is the best way to do it

PR #2215, opened 2025-04-29
update gha ubuntu version

monkeyiq:
IIRC mariadb was locked back at 10.2 because there were some issues with that older version and views. But that was a long time ago so we shouldn't be locked there any more.


monkeyiq:
This works. Next time I will re-enable the postgresql CI tests to get everything up and running.

monkeyiq:
wild ride. have to clean things up now that I have all green again

PR #2216, opened 2025-04-29
Fix: theme v3.0

WanderGomes:
Fixes and improvements to the v3.0 theme, including correcting the authentication theme color and saving the user's selected theme when it's changed in the settings screen.

### DIFF
--- a/www/css/new-ui/_general.css
+++ b/www/css/new-ui/_general.css
@@ -19,9 +19,13 @@ main {
 }
 
 main .core {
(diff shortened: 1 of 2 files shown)


madsi1m:
Tested on our deployment, seems ok (note we have disabled dark theme on our deployment)

monkeyiq:
Saving the dark theme doesn't hold for me. Perhaps this will be the same for others.

PR #2234, opened 2025-05-13
Ability to store IdP metadata in a new table and link to transfer

monkeyiq:
This should help to dice up data for stats and work out which IdP the transfers and users are coming from.

This is a rough first draft. It needs a combing over and other testing before merge but I am sharing it early as others are interested in it ;)

### DIFF
--- a/classes/auth/AuthSPSaml.class.php
+++ b/classes/auth/AuthSPSaml.class.php
@@ -242,12 +242,15 @@ Logger::error("AAABBB ensureLocalIdPMetadata 6 " . print_r($sspmd,true));
         $cfg = $sspmd; 
         $lang = Config::get('default_language');
         $md   = Config::get('auth_sp_idp_metadata_to_capture');
-
+        $lang2 = strtok($lang, "_");
 Logger::error("AAABBB ensureLocalIdPMetadata 6");
         foreach( $md as $k ) {
             if( $cfg[$k]) {
                 $n = $cfg[$k];
                 $data = $n[$lang];
+                if(!$data) {
+                    $data = $n[$lang2];
+                }
                 $idp->{$k} = $data;
             }
         }
--- a/classes/auth/AuthSPSaml.class.php
+++ b/classes/auth/AuthSPSaml.class.php
@@ -218,14 +218,20 @@ class AuthSPSaml
     }
 
 
-    public static function ensureLocalIdPMetadata( $entityId, $idp )
+    public static function ensureLocalIdPMetadata( $entityId, $idp, $force = false )
     {
-        $secs_ago = Config::get('auth_sp_idp_metadata_to_capture_frequency');
-        if( $secs_ago ) {
-            if( $idp->updated > (time() - $secs_ago )) {
+        if( !$force ) {
+            $secs_ago = Config::get('auth_sp_idp_metadata_to_capture_frequency');
+            if( !$secs_ago ) {
                 return;
             }
+            if( $secs_ago ) {
+                if( $idp->updated > (time() - $secs_ago )) {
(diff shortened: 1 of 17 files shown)


classes/auth/AuthSPSaml.class.php
```diff
@@ -216,6 +216,49 @@ public static function attributes()
...
+Logger::error("AAABBB ensureLocalIdPMetadata 2");
+
+   
+        $sspmd = $mdh->getList('saml20-idp-remote', true );
+Logger::error("AAABBB ensureLocalIdPMetadata 3");
+Logger::error("AAABBB ensureLocalIdPMetadata 3 " . print_r($sspmd,true));
+        $sspmd = $sspmd[$entityId];
+Logger::error("AAABBB ensureLocalIdPMetadata 4 " . print_r($sspmd,true));
+Logger::error("AAABBB ensureLocalIdPMetadata 5 entityId $entityId ");
+Logger::error("AAABBB ensureLocalIdPMetadata 6 " . print_r($sspmd,true));
```
WebSpider, line 240: Probably the logging needs to be removed before ending up in a release?

monkeyiq: I would say these logging definitely needs to be removed before ending up in a release.

From the PR description "This is a rough first draft. It needs a combing over and other testing before merge"

These copious debug messages are here because I am debugging the use of this with the help of an admin at an NREN.

madsi1m:
We can probably also remove the config option “auth_sp_idp_filters” which was an attempt to get nicer names from the entityid

monkeyiq:
This will also want Authentications::saml_user_identification_idp updated to reference the new IdP table.

monkeyiq:
Also now that there is an updated column in the IdP table `ensureLocalIdPMetadata()` can limit how often the IdP table might be refreshed for a given IdP. Avoiding churn is probably a good thing.

monkeyiq:
I will push the IdP metadata refresh to a `cron-update-idp-metadata.php` script. Give the option to disable it in the main code path.

madsi1m:
The idpselect on templates/statistics_page.php could probably just pull from the idp table rather than DISTINCT authidp view and use the id as value and a nice name for text

monkeyiq:
I added the extra column in the IdP table to be able to capture more metadata from SSP if we want. It is not used at the moment but since I'll have to do some schema backporting to 2.x for this I thought I'd add it here now.

madsi1m:
Looks ok, i haven't fully tested it yet and I also have to work out a nice way to prefill the idp table and link it back to users. It would be nice to have a oneshot php script that gets all the entiryIDs from SSP and builds the idp table from it

madsi1m:
am i right in thinking if i prepopulate with just an id and entityid and trigger the cron it will fill the idp table?

monkeyiq:
yeah. I might make another script to prepopulate the idp table from the ssp md. This is likely what you are planning to do by other means.

monkeyiq:
import-idp-metadata should prepopulate the `idps` table completely. I might change the inline timeout to not do anything now that the cron scripts are here. If you want to refresh the idps table for an entityid then the `cron-update-idp-metadata.php` will do that. The import will prepopulate. So populate and update are both handled.

madsi1m:
import-idp-metadata.php is perfect, I can run that then use an existing ssp table i have from a plugin i wrote a while back to update the auth table to link the entityid/ids to a user

madsi1m:
happy for this to be merged

monkeyiq:
I will update the docs a little to give an overview of the new scripts. It is likely we might do a little touch up on the views post merge.

PR #2246, opened 2025-05-27
Stats to use select2 lib for dropdown box so we can search idps

madsi1m:
Please double check that I have done the scripts/dev/www-lib-update/* files correctly.


monkeyiq:
www scripts update looks ok. I'll double check next time I do a version bump in there.

monkeyiq (approved):

PR #2251, opened 2025-05-27
fix: run testsuite on ubuntu-latest

WebSpider:
This should use the new `ubuntu-latest` runner instead of the removed `Ubuntu-20.04`


monkeyiq:
This took a bit of effort to get working before https://github.com/filesender/filesender/pull/2215

These changes will be merged into dev3 when the bulk merge next happens.

PR #2255, opened 2025-05-30
attempt to remove saml_user_identification_idp and use idpid instead

madsi1m:
I'm not 100% sure if this will break anything


monkeyiq (approved):
LGTM I might make a follow up PR with a new `fromId` type method, say `viewTupleFromId()` since I can see the `pick`ing of a column from the matching view being a reasonably common activity.

Issue #2260, opened 2025-06-02
v3: transfersauditlogsview might need some left joins


madsi1m:
Saw this while running upgrade/database.sh with mariadb

transfersauditlogsview might need some left joins? 

```
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e] \"SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (utf8mb4_general_ci,IMPLICIT) and (utf8mb4_uca1400_ai_ci,IMPLICIT) for operation '='\"","time":1748826084}
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e] Array","time":1748826084}
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e] (","time":1748826084}
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e]     [name] => exec","time":1748826084}
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e]     [args] => Array","time":1748826084}
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e]         (","time":1748826084}
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e]             [0] => CREATE OR REPLACE VIEW transfersauditlogsview as select t.*,0 as fileid,a.created as acreated,a.author_type,a.author_id,a.target_type,a.target_id,a.event,a.id as aid  from Transfers t, AuditLogs a  where  a.target_id=cast(t.id as char) and target_type = 'Transfer'   UNION select t.*,0 as fileid,a.created as acreated,a.author_type,a.author_id,a.target_type,a.target_id,a.event,a.id as aid  from Transfers t, AuditLogs a, Files f  where  f.transfer_id=t.id     and a.target_id=cast(f.id as char)   and target_type = 'File'  ","time":1748826084}
{"app":"FileSender","process":"upgrade","level":"error","message":"[DBIUsageException details uid:683cf7e47e07e]         )","time":1748826084}

```

madsi1m:
maybe something like this (but i do get less rows on my not quite right dev environment)?

```
[...]

PR #2265, opened 2025-06-02
db: remove some historic views and update scripts over a year old

monkeyiq:
Older explicit upgrade scripts might not work as they jump from one
version to the next.

The most recent one of these only updates the download count for a
transfer from historic values in the auditlogs. If you are jumping
over that to the current version then the download counts might not be
accurate right away unless you migrate via the 2.49 version.


monkeyiq:
This relates to https://github.com/filesender/filesender/issues/2260

PR #2266, opened 2025-06-02
Accept outlook formatting of emails

rick-blok:
I use outlook as my email client and Outlook is the source of email addresses when I need to send files via filesender. When a email address is copied from outlook it is in the format "name <name@domain.com>". Multiple addresses are separated by a semicolon. 

After pasting in filesender the name and brackets have to manually removed in the tiny input field, which is cumbersome especially when multiple addresses are pasted. 

Hence a PR to accept email addresses in outlook format.

Outlook can use a comma to separate first and last name, so that cannot be used as an email separator. 

I updated the language files for the email separator change using deepl translator and google translate, so I'm pretty sure those are correct. I only have some doubts for the Urdu (ur-PK) correction.


madsi1m:
Please fix code spacing and remove unnecessary console.log(). Both will help when helping users debug things.

rick-blok:
done

rick-blok:
Outlook does have mixed mode email addresses. Only when it knows the name of the recipient (from address book, or specified in a mail you've once replied to) it will use the brackets, otherwise it will just be a plain email address. 

As far as I know Outlook only uses a semicolon as a mail separator, and puts a comma between surname and first name when in that order (as opposed to first name first). But a user may enter a comma (or space) to separate the next email address, which may or may not be in the outlook format. Validating a collection of email addresses with different formatting is going to be really challenging. 

My code allows for different formatting after a outlook email address, but ignores anything before it if no semicolon is used to separate. 

One way to circumvent this is to process the email addresses in the input field as soon as a email separator is entered. For example:
`filesender.ui.nodes.recipients.input.on('keydown', function(e) {  `
`    	if(e.keyCode != 13 || e.keyCode !=32 || e.keyCode !=186 || e.keyCode !=188 ) return;`
`        // enter, space, comma or semicolon is pressed`
additionally it could check if the last character in the input is a separator to catch pasting as well.

monkeyiq:
I would imagine that a very common (if not the most common) mode of entering email addresses from outlook would be copy and paste. 

Perhaps we should make a list of test subjects (scrubbed to example.com or whatnot) so we can explicitly narrow down what is working with 2288 above and what we still want to add?

monkeyiq:
I am looking to be able to accept outlook output but not create issues for folks copying from gmail or other locations finding new outcomes in an updated FileSender.

rick-blok:
A colleague mentioned that outlook sometimes also uses parenthesis to indicate the email address, complicating the matter a little more. 
For me, Gmail and mail.live both use angle brackets to indicate the email addresses, similar to outlook. 


I searched on StackOverflow for an answer, and found [one](https://stackoverflow.com/questions/42407785/regex-extract-email-from-strings/54340560). That approach is different by searching for email addresses instead of separators.

The first solution from the accepted answer finds most address in my check list (below), and after adding a + it finds all:
`/[a-z0-9+._-]+@[a-z0-9.-]+\.[a-z0-9-]+/gi`
But it matches 8 out of 18 valid (exotic) mail addresses on wikipedia

The second solution slightly modified finds all except the IPv6 domains, but I'm not sure how useful it is to be able to capture all those valid email addresses
`/(?:[a-z0-9+!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[ \x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)*[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])/gi`

I've tested the regex at https://regex101.com/r/lMYTwg/3

I didn't check yet how to implement this in filesender


List of email addresses with different formats, based on the stackoverflow question
```
ekoprasetyo.crb@outlook.com
db.maulana@gmail.com.
 dee.wien@yahoo.com. .
"deninainggolan@yahoo.co.id" 
Fajar.rohita@hotmail.com	klaas knot (firmansya1404@gmail.aero); 
dydmm, sss <fransisca+jw@gmail.com>, donny_tri_wardono@yahoo.co.id thank a
"Helpdesk | Energiem**ine" <helpdesk@ene---ine.nl>
invalid: John..Doe@example.com 
```

Valid addresses from https://en.wikipedia.org/wiki/Email_address
```
simple@example.com
very.common@example.com
FirstName.LastName@EasierReading.org
x@example.com
long.email-address-with-hyphens@and.subdomains.example.com
user.name+tag+sorting@example.com
name/surname@example.com
admin@example
example@s.example
" "@example.org
"john..doe"@example.org
mailhost!username@example.org
"very.(),:;<>[]\".VERY.\"very@\\ \"very\".unusual"@strange.example.com
user%example.com@example.org
user-@example.org
postmaster@[123.123.123.123]
postmaster@[IPv6:2001:0db8:85a3:0000:0000:8a2e:0370:7334]
_test@[IPv6:2001:0db8:85a3:0000:0000:8a2e:0370:7334]
```

PR #2267, opened 2025-06-02
chore: search/replace SURFnet with SURF

WebSpider:
On 1 januari 2021 the SURFnet name was changed to SURF.

This PR reflects the new name in the FileSender project.
I am unsure if I should target `development` or `development3` at this point. If this needs to change, just let me know.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2009-2017, AARNet, Belnet, HEAnet, SURFnet, UNINETT
+Copyright (c) 2009-2017, AARNet, Belnet, HEAnet, SURF, UNINETT
 All rights reserved.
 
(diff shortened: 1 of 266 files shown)


monkeyiq:
I am not sure that this is needed. The copyright is remains owned by a company after they "rebrand" or are taken over etc.

OTOH I am not a lawyer.

madsi1m:
do we also update the year in the copyright? (also not a lawyer) might be an opportunity to update all the headers

monkeyiq:
I guess I could make a script for this... seems like a better use of time than manually making these things.

WebSpider:
> I am not sure that this is needed. The copyright is remains owned by a company after they "rebrand" or are taken over etc.
> 
> OTOH I am not a lawyer.
> 

I was requested by SURF to address this

PR #2272, opened 2025-06-05
D3UI3 disable user theme until it is updated.

monkeyiq:
Some issues with themes have been reported. Some sites disable them by defualt. Perhaps we should make that the default until the themese have been updated a little bit.

It is best not to present a UI being "unable to log in and unable to upload files". Also a catch22, if you can not login then you can not change your theme.


monkeyiq:
This was also mentioned here https://github.com/filesender/filesender/issues/2273

Issue #2290, opened 2025-06-15
Unable to upgrade from 2.51 to 2.56


miroslavIUCC:
When running the database upgrade, I get this:
Starting upgrade
Upgrading from 0.0 to 2.56
Running 1 pre tasks
Backuping table files
PHP Warning:  Undefined array key 0 in /opt/filesender/classes/utils/DBI.class.php on line 367
PHP Warning:  foreach() argument must be of type array|object, null given in /opt/filesender/classes/utils/DBI.class.php on line 367
database_access_failure
#0 /opt/filesender/scripts/upgrade/versions/upgrade_1.5_2.0a.php(44): DBIStatement->__call()
#1 /opt/filesender/classes/utils/Upgrader.class.php(135): Upgrader::{closure}()
#2 /opt/filesender/scripts/upgrade/upgrade.php(93): Upgrader::run()

Tried to change the user in config.php to root. Same issue.

miroslavIUCC:
Thanks!
It's really confusing.
Works now.

PR #2296, opened 2025-06-22
Remove old script which might be confusing for some users.

monkeyiq:
The upgrade.php has not changed in around 10 years and has a versions that references a 1.5 release. Without re-validating that the upgrade.php is working as expected on the current 2.x series release it is a bit of a liability being here. I have also added a README.md in there to direct folks as to what is interesting in this directory and mention the "view drop" permission issue for MariaDB and direct folks to the right part of the documentation for that issue.


monkeyiq:
This has been raised from time to time and most recently on https://github.com/filesender/filesender/issues/2290.

I have added some docs and removed this old script to save folks time having to choose between one script and the other.

Issue #2291, opened 2025-06-18
Cloud storage backends don't implement canStore($transfer) function


juhassi:
Hi,

One of our users noticed that our filesender testing instance (version 2.54) was throwing a "Not enough space left in storage" error when trying to upload a large file. This should not happen, as our testing instance is currently using CloudS3 backend for storage.

This is caused by StorageCloudS3.class.php not implementing a `canStore($transfer)` function - it fallbacks to the StorageFilesystem.class.php's version of the function, which checks if the transfer would fit on server's local disk.

My ugly hack was to add a dummy canStore() to classes/storage/StorageCloudS3.class.php:

```
    public static function canStore(Transfer $transfer)
    {
        return true;
    }
```

I imagine same issue is also with other StorageCloud\* classes.

monkeyiq:
The dummy function is better than what is there now. I'll take a look at adding that. I will probably add a //FIXME tag to it so that folks who have a limited size window for cloud backing and an interest in enforcing it can look to expand this method. 

Thank you for the detailed report!

WebSpider:
I've had a look at the S3 case, and in the S3 protocol there is no pre-flight checks to see if an object could be stored. Reasons for this are documented, but mostly comes down to S3 policies that you put in place to allow/deny an object based on for instance size that would make this check meaningless.

That being said, I know the current implementation stores objects/chunks in tmp-space before sending them to S3. We could check if there is sufficient space in tmp-space to store the object instead.

monkeyiq:
It is probably a really good idea to document the tmp space requirement. If chunks are stored there until sending is confirmed a sysadmin is likely to want to know it and adjust the local scratch space to ensure it isn't an issue. Disk is reasonably cheap, service interruptions are reasonably inconvenient.

WebSpider:
This has been fixed and released in 2.56/3.0rc9

PR #2297, opened 2025-06-24
Document usage of php temp-dir for s3 cloud backend better

WebSpider:
This documents the behaviour mentioned in #2291 and sheds a bit of light into why filesender will use PHP's temp space (which defaults to `$TMPDIR` in many cases) when you enable a CloudS3 storage backend.


monkeyiq:
Thanks mate.

PR #2326, opened 2025-07-13
fix: set 2 to CURLOPT_SSL_VERIFYHOST

makinog3:
Because 1 (true) caused an error in curl 7.28.1 and later


monkeyiq:
Thank you for the update. 

I have verified `interface.c` (line 1692) in php 8.4.6 which implicitly converts a value `1` to `2` before passing it onwards to the C API `curl_easy_setopt`. I see in curl 7.29.0 that it rejects a value `1` for this option. I see in curl 8.14.1 that they have started treating `1` and `2` as the same. 

The short take away is that using `2` in FileSender should always work and not depend on which libcurl is backing your php. So this PR is very useful.

Issue #2334, opened 2025-07-17
AuthSPShibboleth does not have a method "ensureLocalIdPMetadata" (development3)


makinog3:
We are testing FileSender 3.0, the latest of `development3`, with Shibboleth authentication.

`AuthSPShibboleth does not have a method "ensureLocalIdPMetadata"` error happened when creating a transfer.

```
[17-Jul-2025 13:12:23 Asia/Tokyo] PHP Fatal error:  Uncaught TypeError: call_user_func(): Argument #1 ($callback) must be a valid callback, class AuthSPShibboleth does not have a method "ensureLocalIdPMetadata" in /opt/filesender/global-filesender/classes/auth/AuthSP.class.php:105
Stack trace:
#0 /opt/filesender/global-filesender/classes/data/Transfer.class.php(646): AuthSP::ensureLocalIdPMetadata()
#1 /opt/filesender/global-filesender/classes/rest/endpoints/RestEndpointTransfer.class.php(752): Transfer::create()
#2 /opt/filesender/global-filesender/classes/rest/RestServer.class.php(302): RestEndpointTransfer->post()
#3 /opt/filesender/global-filesender/www/rest.php(37): RestServer::process()
#4 {main}
thrown in /opt/filesender/global-filesender/classes/auth/AuthSP.class.php on line 105
```

makinog3:
Thank you very much! I confirmed it to be fixed.

PR #2335, opened 2025-07-19
add basic ensureLocalIdPMetadata for Shibboleth

monkeyiq:
For this to be fully functional in the new stats update I need to play around more with some IdPs for Shibboleth and check that per IdP data is properly presented. Many of the fields in the idps table for Shibboleth will be left as null for now.

For SSP the code was reading the remote idp listing from the library. For the code to match in Shibboleth it should also be looking at the metadata for the remote IdPs that the Shibboleth SP knows about and adding that information into the local idps table in FileSender.

### DIFF
--- a/classes/auth/AuthSPShibboleth.class.php
+++ b/classes/auth/AuthSPShibboleth.class.php
@@ -170,6 +170,24 @@ class AuthSPShibboleth
         
         return self::$attributes;
     }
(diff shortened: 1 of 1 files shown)


monkeyiq:
This was raised in https://github.com/filesender/filesender/issues/2334

PR #2342, opened 2025-07-31
Fix filesender.py so that it doesn't get messed up by clidownload.php

### LINKED ISSUE #2341: Clidownload.php messes up filesender.py
Clidownload.php replaces [base_url] with current FileSender URL, this is fine as this provides a default value.

Problem is, it replaces one too many as arround L128 of filesender.py this :

`if base_url == "[base_url]":
  requiredNamed.add_argument("-b", "--base_url", required=True)
else:
  parser.add_argument("-b", "--base_url")`

Becomes this :

`if base_url == "https://my.file.sender/rest.php":
  requiredNamed.add_argument("-b", "--base_url", required=True)
else:
  parser.add_argument("-b", "--base_url")`

Making it so using the default value leads to the --base_url argument mandatory ...

I fixed this by changing `if base_url == "[base_url]"` to `if base_url == "[" + "base_url" + "]"` so that the test is not messed up with.

I will make a pull request asap.

yent:
Fixes #2341

Clidownload.php was replacing one too many [base_url] instances in filesender.py, thus messing up --base_url argument mandatory status, this uses concatenation to "hide" the specific instance of [base_url] from the str_replace in clidownload.php

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -125,7 +125,8 @@ else:
   
 requiredNamed.add_argument("-r", "--recipients", required=True)
 
-if base_url == "[base_url]":
+# Do not change this seemingly out-of-place concat as it avoid getting this test messed up by clidownload.php
+if base_url == "[" + "base_url" + "]":
   requiredNamed.add_argument("-b", "--base_url", required=True)
 else:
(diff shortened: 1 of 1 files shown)


yent:
Also included fix for escapement warning when using filesender.py

PR #2345, opened 2025-08-05
Global FileSender

makinog3:
Global FileSender project from NII, the file forwarding feature, is a function that asynchronously and quickly forwards uploaded files to another FileSender site. The user can instruct forward to another server using the Transfer setting(`forward_to_another_server` and `forward_server_name`). The receiver will receive a download notification mail from the server at the destination. The file transfer tool used is MMCFTP, developed by NII, but other methods (SCP, REST API) can also be used. Currently, the `storage_type` supported by these transfer tools is only `filesystem`.


monkeyiq:
Once I do a merge I will import the translations again from poeditor. This will stomp over the lang updates in this PR but we can work on updating those through poeditor after.

monkeyiq:
Thank you @makinog3 and NII for the PR!

PR #2349, opened 2025-08-08
Load additional config from $auth_attrs['additional'][$attr]

madsi1m:
This patch allows admins to load different configs from `$config['auth_sp_additional_attributes']` using `$config['auth_config_regex_files']`.

The additional attributes show up as auth_attrs['additional'][$attr] as an array of one item.


madsi1m:
https://github.com/monkeyiq/filesender/tree/2025/aug/auth_config_additional_regex_files seems to do what we need, closing this one.

PR #2350, opened 2025-08-11
Config: SAML additional attriubtes to config

monkeyiq:
A new auth_config_additional_regex_files config key that matches against the whole array (one at a time) for the picked attributes.

Perhaps we should also add at some stage a "sort" type key to allow the provided array to be sorted so that the order that multiple matching config files is applied is determined explicitly. For example, student and employee we might like the student to sort first so that the employee config can be more permissive or have a larger quota.

### DIFF
--- a/classes/utils/Config.class.php
+++ b/classes/utils/Config.class.php
@@ -218,8 +218,8 @@ class Config
 
         // Load config regex overrides if used and present
         // if not authenticated then do not throw, just do not load these files
-        self::handleConfigRegexFiles( 'auth_config_regex_files', false );
         self::handleConfigRegexFiles( 'auth_config_additional_regex_files', true );
+        self::handleConfigRegexFiles( 'auth_config_regex_files', false );
         
         // ensure mandatory config settings file exists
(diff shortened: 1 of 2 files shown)


monkeyiq:
Since this is using a new `auth_config_additional_regex_files` it makes sense to log an error if that array contains a key that is not also in `auth_sp_additional_attributes`. We could add it implicitly but I think for such complicated configs it is better for the config.php to be the golden source and not rely on implicit adds.

monkeyiq:
This also clears the way for a `auth_config_additional_regex_sort` that can sort an array of values for an additional attribute to allow a user who is a student and employee to have the employee.config applied after the student.config so they get the later privs and the ordering is explicit and not depending on the ordering that the auth system returns the array in.

monkeyiq:
This relates to https://github.com/filesender/filesender/pull/2349

madsi1m:
This seems to do what I need

monkeyiq:
never an easier time to switch the order :)

madsi1m:
OOPs, can we merge to dev3 as well?

monkeyiq:
I think we both did the merge now :)

PR #2360, opened 2025-08-19
Forward-p(parallel)REST

madsi1m:
Do forward-REST.php in parallel.

Created new forward-pREST.php as it is a good idea to have a basic forward-REST.php as a fall back.

pREST allows you to set `workers` in the `method_options`.

As you might imagine, this is a lot faster than REST as things can happen in parallel.

Have also added fseek capabilities to the Stream storage classes.


madsi1m:
@makinog3 might be of interest

monkeyiq (approved):

PR #2363, opened 2025-08-20
Forward: forward capabilities in config

madsi1m:
Each site shows what forwarding options is possible in /info api call using the new option `$config['forward_capabilities']`

No longer need to set a method for each remote site in `forward_server_list`

Added some smarts that allows FS to decide which method to use based on site forward_capabilities vs remote site forward_capabilities (findBestForwardMethod)


madsi1m:
Modified code slightly so we can have a setup where the same config can be used across multiple sites around the country and the code ignores itself as a forwarding site. For example if i have two sites each site would show only the other site as a forwarding location.

monkeyiq (approved):

PR #2368, opened 2025-08-24
tiny forward UX fix

madsi1m:
When forward is enabled in advanced transfer settings and user selects "a transfer link" the forward heading is shown which it shouldn't.


madsi1m:
not 100% sure if you want to name divs a certain way, feel free to change

monkeyiq (approved):

PR #2410, opened 2025-10-04
perf: it can be expensive to process nulls

monkeyiq:
If you have millions of files this can be a good performance boost.

### DIFF
--- a/classes/utils/Template.class.php
+++ b/classes/utils/Template.class.php
@@ -255,6 +255,9 @@ class Template
      */
     public static function Q($data)
     {
+        if( !$data ) {
+            return "";
+        }
         return self::sanitizeOutput($data);
     }
     


madsi1m:
Can we merge this and some of the last few days work into dev3 please?

monkeyiq:
Merged. 

I am planning to move some of the uid fields from varchar to using the uuid type in mariadb and postgresql. Not using that type is just leaving performance on the table. I also have some other db updates flagged (precache dl count per file instead of interrogating the auditlog table, maybe add transfer_id to auditlog to simplify some queries). I have seen load times for my transfers go from many minutes to 10s with 100,000 files on the my transfers page in 2.x. It is interesting because if even one of the queries falls to ineffective execution on that page then you loose. I also plan to allow millions of files so it has to be 10x+ better again. So getting them all going as fast as possible and knowingly moving some out of band until they are really needed will help.

Some of those performance things flow into 3.x UI in the transfer details page if you have larger transfers.

PR #2425, opened 2025-10-10
Feat OIDC

### LINKED ISSUE #2424: Feature Request: Add native OpenID Connect authentication support
## Motivation

FileSender currently supports SAML / Shibboleth via a separate SimpleSamlPhp installation.  
In practice the SimpleSamlPhp setup is often **hard to maintain** and requires extra manual steps (installation, configuration, updates). This creates friction for administrators who want to integrate FileSender with modern Identity‑Provider (IdP) stacks.

OpenID Connect (OIDC) has become the de‑facto standard for web‑based authentication and is widely adopted. Providing native OIDC support would:

- **Reduce the operational overhead** – no external SimpleSamlPhp installation is required.
- **Broaden compatibility** – administrators can connect FileSender directly to any OIDC‑compliant IdP (Keycloak, Auth0, Azure AD, Google, etc.).
- **Improve user experience** – a single, consistent configuration inside the main application keeps the setup simple and transparent.
- **Future‑proof the authentication layer** – OIDC is actively maintained and offers modern security features (PKCE, etc.).

## Proposed Implementation

- Integrate the **OpenID‑Connect‑PHP** library ([https://github.com/jumbojett/OpenID-Connect-PHP](https://github.com/jumbojett/OpenID-Connect-PHP)).
- Add a new configuration section (e.g. `auth_sp_oidc_...`) to `config/config.php` and `config/config_private.php` (client ID, client secret, discovery URL, claims, etc.).
- Implement the authentication flow (authorization request → token exchange → userinfo retrieval) within the core authentication module, following best‑practice OIDC patterns.
- Provide clear documentation on how to enable OIDC, configure the required parameters,.
- **Tested successfully with Keycloak** (including login, logout, and attribute mapping).

flowdnb:
Closes #2424

### DIFF
--- a/docs/v2.0/admin/configuration/index.md
+++ b/docs/v2.0/admin/configuration/index.md
@@ -270,6 +270,8 @@ A note about colours;
   * [auth_sp_oidc_uid_attribute](#auth_sp_oidc_uid_attribute)
 	* [auth_sp_oidc_email_attribute](#auth_sp_oidc_email_attribute)
 	* [auth_sp_oidc_name_attribute](#auth_sp_oidc_name_attribute)
+  * [auth_sp_oidc_groups_claim](#auth_sp_oidc_groups_claim)
+  * [auth_sp_oidc_required_groups](#auth_sp_oidc_required_groups)
 * __Shibboleth__
 	* [auth_sp_shibboleth_uid_attribute](#auth_sp_shibboleth_uid_attribute)
 	* [auth_sp_shibboleth_email_attribute](#auth_sp_shibboleth_email_attribute)


flowdnb:
An alternative worth taking a quick look at could be [`facile-it/php-openid-client`](https://github.com/facile-it/php-openid-client). I could run a quick proof‑of‑concept.

A small compatibility wrapper could be created around the OIDC client (e.g. `class OIDCClientWrapper`) so that the specific library can be replaced later without changing the rest of the code base.

www/oidc.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+require_once('../includes/init.php');
+require_once __DIR__.'/../optional-dependencies/oidc/vendor/autoload.php';
```
monkeyiq: It might make sense here to check if `auth_sp_type != oidc` and exit early. Probably before the oidc vendor autoload.

www/oidc.php
```diff
@@ -0,0 +1,117 @@
...
+
+    if (!$target) {
+        $target = Config::get('site_logouturl') ?: '/';
+    }
+    
+    header('Location: ' . $target);
+    exit;
+}
+
+$target = isset($_GET['target']) ? filter_var(urldecode($_GET['target']), FILTER_SANITIZE_URL) : '';
```
monkeyiq: Since `target` is coming in from a CGI parameter it might make sense to do more sanitizing here.

I don't think there are that many places that make this `$target` in the code. For example, logon_page.php uses `Utilities::http_build_query()` to make the URL passing the active page as the `s` parameter. For that case we might expect that `$target` here matches the same prefix `Utilities::http_build_query` would have used. It might also be a full value to the `?` part.  Then we might also expect that perhaps there is only a single `s` parameter. That `s` could also be obtained and passed to `GUIPages::isValidValue` if we wanted to be really pedantic.

I could also add these checks though you might find the matching strings using `Logger::error()` and also be able to see if the system there is calling odic.php in other ways than the one I mention above and expand the cloud of valid values accordingly.

I don't have any real clean attacks for this at the moment. Passing an `action=logout` with a redirect to another site perhaps.

monkeyiq (approved):

www/oidc.php
flowdnb: According to my analysis, the only use of `www/oidc.php` with a `target` is for authenticated downloads. (e.g., `https://filesender.example.org/oidc.php?action=login&target=https%3A%2F%2Ffilesender.example.org%2F%3Ftoken%3D464fbf63-2803-4f69-b4ad-c60d43b108a3%26s%3Ddownload`). So sanitisation could be that stringent.

Issue #2427, opened 2025-10-13
The bundled asmCrypto.js is old and archived


bozicm:
Hello!

The [bundled project](https://github.com/filesender/filesender/tree/development/www/js/asmcrypto) was [archived in 2023](https://github.com/asmcrypto/asmcrypto.js/) and encourages to use native WASM. I guess this is what it refers to: https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API Mozilla used it in their [Send application](https://github.com/mozilla/send/blob/master/docs/encryption.md).

monkeyiq:
That is to say we might be ok to remove support for non Chromium Edge and older browsers in 2.x now.

madsi1m:
"Edge Legacy" the official name it seems was discontinued in 2020. I see no reason to keep support for it.

PR #2431, opened 2025-10-15
crypto: remove webasm fallback for non chromium Edge

monkeyiq:
This was raised in https://github.com/filesender/filesender/issues/2427

The only thing the webasm stuff was being used on was pbkdf2 for AES-CBC for Edge that is not Chromium backed. That is a 5+ year old browser at this stage and should have been auto updated to something newer over that time frame.


WebSpider:
This isnt a deprecation, it's a removal of the feature.

monkeyiq:
It is probably better to just leave this in the 2.x series at this stage.

Issue #2434, opened 2025-10-16
UI3: Guests page can't move valid_only_one_time to advanced area


madsi1m:
Hi, Setting the following does not move it into advanced settings

```
$config['guest_options'] = [
  'valid_only_one_time' => [
    'available' => true,
    'advanced' => true,
    'default' => false
  ],
```

<img width="902" height="1036" alt="Image" src="https://github.com/user-attachments/assets/b6eb3531-38c6-44b3-a8d7-f90e8b0afef5" />

madsi1m:
while `guest_options_to_force_to_top_array` works, should those things be hard coded to begin with?

monkeyiq:
Things shouldn't be hard coded. I am trying to make things possible without interfering with the vision of the UI team.

PR #2435, opened 2025-10-17
ui3d3 allow the advanced option for guest_options to work on all

monkeyiq:
The new option allows a sysadmin to stop items being lifted to the top of the list and put them into the advanced section instead.


monkeyiq:
This was raised in https://github.com/filesender/filesender/issues/2434

Issue #2428, opened 2025-10-14
ui3: allow one time upload guests to see their uploads


monkeyiq:
A follow up to https://github.com/filesender/filesender/pull/2397 to allow one time upload guests to see their previous uploads.

PR #2437, opened 2025-10-20
ui3: allow one upload and recently expired guest to see past uploads

monkeyiq:
The past uploads are only the time and file names. The guest can not download the files so they can not use this is a distribution center.

This PR adds a new option
guest_transfers_page_number_of_days_expired_guest_can_return which can allow guests to view files in a grace time of this many days. A guest who can only upload once will expire right away. Without this provision they can not see that their upload has succeeded later on.

I am still testing this PR.

This was requested by an NREN and I recorded that in https://github.com/filesender/filesender/issues/2428

### DIFF
--- a/classes/auth/AuthGuest.class.php
+++ b/classes/auth/AuthGuest.class.php
@@ -87,9 +87,26 @@ class AuthGuest
                 }
                 
                 $guest = Guest::fromToken($vid);
-                    
+
+                $allowExpiredGuestToSeeLimitedPages = false;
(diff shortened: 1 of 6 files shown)


monkeyiq:
I am happy with that mod. I sort of grew the menubar out a bit to allow a guest to see the names of past uploads or make another if they are allowed. I'll import the new terms onto poeditor after the merge so the text can be updated there.

monkeyiq:
New terms created ... https://github.com/filesender/filesender/pull/2444

Issue #2420, opened 2025-10-08
Encrypted downloads fail in restricted environments due to blocked file editing browser permission


flowdnb:
We have received reports that encrypted downloads fail in environments with restricted browser settings, specifically when the "File editing" permission is blocked in Chromium-based browsers (e.g., Chrome, Edge), which is sometimes applied at the organisational level.

This restriction prevents the `showSaveFilePicker` window from opening, which is required for encrypted downloads to complete.

**Error Message:** `Failed to execute 'showSaveFilePicker' on 'Window': The request is not allowed by the user agent or the platform in the current context.`

**Question:** Is this behavior by design, or is there a potential solution to allow encrypted downloads in environments with these restrictions? We are seeking clarification on whether a workaround or alternative approach is possible.

## Steps to Reproduce

1.  Configure a Chromium-based browser with the "File editing" permission blocked.  
    *   **For Edge:** This can be done through settings at `edge://settings/content/fileEditing`.
2.  Attempt to download an encrypted file.
3.  Observe that the download fails, and no "Save As" dialog appears.

## Expected Behavior

The download should either:

*   Succeed by utilizing an alternative method for saving the encrypted file that does not require the blocked permission, if such a method is available.
*   Provide a clear error message to the user explaining that the download is blocked due to restricted permissions.

## Additional Information

This issue is likely to affect users in clinical or other sensitive environments where browser permissions are centrally managed for compliance/regulatory reasons.

monkeyiq:
Thank you for the detailed error report.

The first update should likely be to improve the error handling around this. I will probably experiment with trying to fall back to another method of writing data or all the way back to `blobSinkLegacy` if the browser policy is disabling the more modern API use. 

Falling "all the way back" will be bad as it has RAM implications where the whole download has to be stored twice in RAM while downloading IIRC. But for modern machines and relatively small files (200mb etc) it should at least get the job done which is much better than no download.

Falling back from filesystemwritablefilestream_sink to streamsaver_sink will be a better option assuming that later works in that environment.  Something for me to test & update for locally using your detailed instructions.

PR #2438, opened 2025-10-22
Added more robust Error handling and fallback to other Download options if FileSystemwritableFilestream does not work

flowdnb:
Proof-of-concept implementation for #2420

### DIFF
--- a/www/js/crypter/crypto_app.js
+++ b/www/js/crypter/crypto_app.js
@@ -872,7 +872,7 @@ window.filesender.crypto_app = function () {
             oReq.addEventListener("progress", function(evt){
                 window.filesender.log("downloadAndDecryptChunk(progress) chunkid " + chunkid
                                       + " loaded " + evt.loaded + " of total " + evt.total );
-                if (evt.lengthComputable) {
+                if (evt.lengthComputable || evt.loaded) {
                     var percentComplete = Math.round(evt.loaded / (1*encryption_details.crypted_chunk_size) *10000) / 100;
                     var percentOfFileComplete = 100*((chunkid * encryption_details.chunk_size + evt.loaded) / encryption_details.filesize );
                     


monkeyiq:
It may be that an inline fallback is needed here to work nicely.

It may not be easy or robust to be able to check if you have the "file editing" permission from javascript.  If I force file editing off in Chrome (Settings, Privacy, Permissions, File Editing) and then change this function in `filesender-config.js.php` to just return `false` then the fallbacks are triggered and I can download a file.

So another option would be to allow an admin to force this to off if their browser policy is to disallow it. That is assuming that the permission can not be reliably tested for in JS.

```
window.filesender.config.isFileSystemWritableFileStreamAvailableForDownload = function() {
return false;
//    return 'showSaveFilePicker' in window;
};
```

monkeyiq:
Thank you for making the pull request!

monkeyiq:
As there are a number of changes already in `development` I think I will delay merging this until after the next release. I might make the next release 2.60 which will have the database updates and performance improvements  in it. After that and a period of time for possible bugfix updates then I will turn attention to merging this and retesting the download paths with it.

flowdnb:
Thank you for the thorough review and for considering this PR in the next release plan.

www/js/crypter/crypto_app.js
```diff
@@ -873,7 +864,7 @@ window.filesender.crypto_app = function () {
             oReq.addEventListener("progress", function(evt){
                 window.filesender.log("downloadAndDecryptChunk(progress) chunkid " + chunkid
                                       + " loaded " + evt.loaded + " of total " + evt.total );
-                if (evt.lengthComputable || evt.loaded) {
+                if (evt.lengthComputable) {
```
monkeyiq: I don't think we want to revert this change https://github.com/filesender/filesender/pull/2324

monkeyiq:
I see a regression on Safari. Downloading an encrypted file goes from showing the `download %` on the download page to downloading everything in the background with no progress and only showing on completed. I haven't dug into the diagnostics to see why this is the case yet. Just an A/B of this PR vs `development` with a macos client + safari.

monkeyiq:
This is in 2.62. Thanks again for investigating things and making this PR!

PR #2451, opened 2025-11-19
add encryption_mandatory condition handling to upload_page.js

### LINKED ISSUE #2450: UI3: Encryption switch can be disabled even if encryption_mandatory config is set (resulting in silently failed upload attempt)
The file `upload_page.js` contains a condition for the configuration item `encryption_mandatory_with_generated_password`, but lacks a condition for `encryption_mandatory`.

Currently, it hides the encryption switch/slider and locks the display of the encryption form elements. This is already a workaround, as strangely enough, displaying the switch but disabling it via `filesender.ui.nodes.encryption.toggle.prop("disabled", true);` in`upload_page.js` does not work, even though this is possible via the JavaScript console.

flowdnb:
Closes #2450

### DIFF
--- a/www/filesender-config.js.php
+++ b/www/filesender-config.js.php
@@ -88,6 +88,7 @@ window.filesender.config = {
     chunk_upload_security: '<?php echo Config::get('chunk_upload_security') ?>',
     
     encryption_enabled: '<?php echo Config::get('encryption_enabled') ?>',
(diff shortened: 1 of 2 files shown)


monkeyiq:
I am guessing github doesn't like the `\u2011` character in the branch name.

PR #2457, opened 2025-11-24
chore(csp): provide sane default for 'frame-ancestors' content policy

WebSpider:
This provides a sane default for the `frame-ancestors` content policy.

This policy will prevent FileSender from being embedded in a frame by default, to prevent clickjacking and is based on a recommendation from a periodic automated security scan.


monkeyiq:
@madsi1m IIRC at some stage in the past there was some use case with frames and FileSender aarnet had. It seems from a quick look at the deployment that there are no frames in use now for you.

madsi1m:
Yes, we used to need to do this. No longer the case. Happy with this PR

PR #2465, opened 2025-12-04
chore(docs): Update 3.x to be release, move 2.x to security-fixes-only

WebSpider:
Also update documentation references from 2.x to 3.x as needed:

- Security policy is updated to reflect this
- Updated some links pointing to v2 docs now go to v3
- Updated the `roadmap.md` a bit, but this still seems out-of-date
- Updated contributing guidelines (maybe we should create a `CONTRIBUTING.md` ?)
- Removed some passages that mention a minimum required version 2.x, that now is outdated since we are on 3.x

Feel free to suggest changes :)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,13 +4,13 @@
 
 We provide security updates for the following versions:
 
-| Version | Supported          |
-| ------- | ------------------ |
-| 3.x     | ⚠️ Not supported (yet) |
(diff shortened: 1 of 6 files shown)


WebSpider:
Ok, i wasnt aware of that, probably because im not in the board. 

Lets keep this PR open until after that release then?  I'll fix any merge conflicts that arise.

Issue #2512, opened 2026-01-12
Array to string conversion in GUI.class.php getFileNamesForDisplay()


victoritis:
The function `getFileNamesForDisplay()` in `classes/utils/GUI.class.php` uses `$transfer->downloads` (an array) where an integer value is expected.

Problem: When viewing the transfers table on FileSender 3.3, PHP throws:

PHP Warning: Array to string conversion in classes/utils/GUI.class.php on line 484

The file name column displays "Array" before the actual filename.
Comparison:
- Line 479 (Incorrect): `$dlcount = $transfer->downloads;` (downloads is an array)
- Should be: `$dlcount = $transfer->download_count;` (download_count is an integer)
Steps to reproduce:
1. Upload a file using FileSender 3.3
2. Navigate to "My Transfers"
3. Observe the file names column in the transfers table
4. Check PHP error logs
Suggested Fix: Change `$transfer->downloads` to `$transfer->download_count` on line 479.

PR #2514, opened 2026-01-13
This seems to have been left out with the efficiency updates.

monkeyiq:


monkeyiq:
This was mentioned here https://github.com/filesender/filesender/issues/2512

PR #2518, opened 2026-01-20
feat: Dynamic expiry days selector based on max_transfer_days_valid

victoritis:
## Description
This PR improves the UX of the expiry days selector in `upload_page`. Instead of using a static list or requiring manual configuration of `selectable_transfer_days_valid`, this change dynamically generates a progressive list of options based on the `max_transfer_days_valid` setting.

### The Problem
Previously (in master/v3), the selector had hardcoded values (7, 15, 30, 40). In development3, it relied on `selectable_transfer_days_valid`, but this requires admins to manually sync two configuration arrays. If `max_transfer_days_valid` was set to a low value (e.g., 5 days), users might see no options or only the default.

### The Solution
The `selectable_*_days_valid` arrays now default to empty. When empty, the code generates options dynamically:
- **Days 1-7:** Shown individually (for fine-grained control on short transfers).
- **Days 8-30:** Shown in weekly intervals (14, 21, 28).
- **Days 30+:** Shown in monthly intervals (30, 60, 90...).
- **Max limit:** The configured `max_transfer_days_valid` is always included.
- **Default:** The `default_transfer_days_valid` is always included.

If an admin explicitly configures the array with custom values, those are used instead (backward compatible).

This logic is applied to:
- Standard Uploads (`templates/upload_page.php`)
- Guest Invitations (`templates/new_invitation_page.php`), respecting `min_guest_days_valid`

## Examples
- `max_days=5` -> Options: `1, 2, 3, 4, 5`
- `max_days=14` -> Options: `1...7, 10 (default), 14`
- `max_days=90` -> Options: `1...7, 14, 21, 28, 30, 60, 90`

## Type of change
- [ ] Bug fix
- [x] New feature (non-breaking change which adds functionality)
- [x] Enhancement (UX improvement)


victoritis:
Thanks for the feedback! I've updated the implementation:

Instead of adding new boolean config flags, I changed the default value of selectable_transfer_days_valid and selectable_guest_days_valid to an empty array.

Empty array (default): Uses progressive day generation automatically.
Populated array: Uses the admin's custom values (fully backward compatible).
This way, existing installs that have explicitly configured these arrays will continue working as before, while new installs get the improved. In addition, this is done for instalations where a low 'max_transfer_days_valid' is setted.

PR #2520, opened 2026-01-21
Docker deployment for local testing

sunetfreitag:
This PR changes the scope of the provided docker example, reducing it to a simple local deployment running on http. It uses the Filesender 3.3 release, is based on a recent php-fpm image (php:8.5-fpm) and updates SimpleSAMLphp to 2.4.4. It also adds a few minor patches to the local deployment. Instructions on how to run the image have been updated in the readme file.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -54,7 +54,7 @@ stop-docker:
 .PHONY: certificate
 certificate:
 	-rm assets/config/simplesamlphp/cert/localhost.*
-	rm assets/config/simplesamlphp/metadata/saml20-idp-remote.php.dist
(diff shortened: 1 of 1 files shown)


monkeyiq:
It looks like the third issue for `token` is also fixed in developement3 now.

https://github.com/filesender/filesender/blob/3f1f8a704afc3a3666bc86696e324d2bfac6f6fa/templates/transfer_detail_page.php#L87

monkeyiq:
Given that 3.4 will be needed for those 3 patches to be dropped I am thinking about merging this now and we can continue with updates (particularly for the patches) as newer releases arrive.

Issue #2569, opened 2026-02-19
# Does FileSender support native OIDC authentication without SimpleSAMLphp?


victoritis:
Hi,

I've been using FileSender 3.5 with SimpleSAMLphp + the `authoauth2` module for OpenID Connect authentication against a Keycloak IdP. However, I've been running into dependency issues with `firebase/php-jwt` (security advisories PKSA-y2cr-5h3j-g3ys and PKSA-2kqm-ps5x-s4f5), which Composer 2.9+ now blocks by default.

While investigating alternatives, I noticed that the FileSender 3.5 source code seems to include built-in OIDC support:

- [www/oidc.php](cci:7://file:///tmp/filesender-3.5/www/oidc.php:0:0-0:0) — OIDC login/logout entry point
- [classes/auth/AuthSPOidc.class.php](cci:7://file:///tmp/filesender-3.5/classes/auth/AuthSPOidc.class.php:0:0-0:0) — OIDC authentication delegation class
- [optional-dependencies/oidc/](cci:7://file:///tmp/filesender-3.5/optional-dependencies/oidc:0:0-0:0) — `jumbojett/openid-connect-php` library
- [classes/utils/ConfigPrivate.class.php](cci:7://file:///tmp/filesender-3.5/classes/utils/ConfigPrivate.class.php:0:0-0:0) — reads `client_id`/`client_secret` from [config_private.php](cci:7://file:///root/Proyectos/filesender/%7B%7Bcookiecutter.project_slug%7D%7D/containers/app/filesender/config_private.php:0:0-0:0)
- `AuthSP.class.php` dynamically loads `AuthSPOidc` when `auth_sp_type = 'oidc'`

**My questions:**

1. Is this native OIDC client officially supported and production-ready?
2. Is there any documentation on how to configure it? The [installation guide](https://docs.filesender.org/filesender/v3.0/install/) only covers SimpleSAMLphp.
3. What is the recommended setup? Specifically:
   - Which config parameters go in [config.php](cci:7://file:///root/Proyectos/filesender/%7B%7Bcookiecutter.project_slug%7D%7D/containers/app/filesender/config.php:0:0-0:0) vs [config_private.php](cci:7://file:///root/Proyectos/filesender/%7B%7Bcookiecutter.project_slug%7D%7D/containers/app/filesender/config_private.php:0:0-0:0)?
   - What redirect URI should be registered in the OIDC provider? (I'm assuming `https://<domain>/oidc.php`)
[...]

PR #2600, opened 2026-03-05
fix(oidc): add missing idp attribute

flowdnb:
Fix missing 'idp' attribute in OIDC authentication 

#2569


monkeyiq:
This should also be merged into `development3` when I do the periodic merge between d and d3.

flowdnb:
This PR covers v2, but per PR #2616 and PR #2335's precedent, v3 still needs OIDC-specific `ensureLocalIdPMetadata`. PR #2335 demonstrated that explicit IdP metadata handling is required for complete SSO provider coverage - this pattern hasn't been applied to OIDC in v3 yet. Suggest implementing #2335 (or at least the changes to `classes/auth/AuthSPOidc.class.php`) to align OIDC with the established SAML/Shibboleth metadata persistence approach.

PR #2603, opened 2026-03-06
don't allow guests to choose the transfer forward to another server

makinog3:
At "advanced settings" on the new invitation screen, 
the `forward_to_another_server` and `forward_server_name` are displayed as a toggle switch.
But the sender who sends an invitation can't decide them before.


monkeyiq:
This makes sense. A guest who is looking to upload and forward a transfer is a fairly advanced guest. They should probably be a regular user at that point if they know that much about the system.

I am happy to limit confusion on the use of this system and help promote the wider adoption of it. We can always work out if a user wishes to forward some of their guests transfers having a way to make that happen later.

PR #2619, opened 2026-03-13
Add from_address as option, fixes case where username matches regex but is not the email address of the user

WesHarrell:
The client assumes that if the username resembles an email address then use it as the email address.  This fails in some cases, eg. username in form `SurnameF@mydomain.com`  but email address is `First.Surname@mydomain.com` 
This patch adds from_address as optional, keeping the requirement should username not match regex.

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -134,6 +134,8 @@ else:
 # if username is not a valid email address then ensure user supplies a valid email address
 if username is None or not bool(re.match(r'([^@|\s]+@[^@]+\.[^@|\s]+)', username)):
   requiredNamed.add_argument("-f", "--from_address", help="filesender email from address", required=True)
+else:
+  parser.add_argument("-f", "--from_address", help="filesender email from address")
 
 args = parser.parse_args()
 debug = args.verbose


monkeyiq:
There has been some discussion around moving the `-f` option from the client to stop users selecting arbitrary from addresses.

vladimir-mencl-eresearch:
Thanks @monkeyiq 

To solve this properly, is there a way we could make the client use the email address received in the federated login and stored in the filesender account, instead of assuming username is also an email address?

Looking at the code (from a distance), an ideal solution appears to stop passing `from` address in the `/transfer` API call and instead letting FileSender pick it up from the internal user account.

If that's not possible, we could perhaps at least include the email address in the generated `filesender.py.ini` - and read it from the config by default.

Would either of these sound reasonable to you?

monkeyiq:
I think a reasonable solution is storing the email address from the IdP into the database. It is harder to pass around because the filesender.py is using it's own auth system with that `apikey`. 

There is a small privacy hit for storing the user's email address, which many login systems warn you about (sharing email, name etc with login system). It would then be up to the installation to notify their users of this update somehow, that their email address will be stored for the duration of the account.

I think I'll add some code to enable storing this soon and then we can work on squeezing out the ability for a user to nominate anything else.

PR #2640, opened 2026-03-24
Efficiency: use stream_copy_to_stream to allow lower CPU/RAM use

monkeyiq:
This PR makes the main code path that is used to write data to disk in the server code more efficient in terms of CPU and memory use.

This update will delay reading bytes in a very specific case which is uploading file data. By not reading the data until the destination we wish to write those bytes we can allow the opportunity to have the Linux kernel copy the bytes inside the kernel by letting PHP use `copy_file_range(2)` to avoid memory allocation and extraneous data copying. This significantly reduces CPU use and increases upload speed even on a local installation.

For local testing I was using CentOS-6.10-x86_64-bin-DVD1.iso which is about 3.8 gb. For a non encrypted upload I saw wall times going from 27 to 20 seconds. For encrypted upload from 34 seconds down to 28 seconds. As one would expect with end-to-end encryption, the encryption does not effect the server load and an even 7-8 seconds for encryption added to both wall times.

I think this is significant as the time being spent between this two code paths is being used to allocate buffers and copy data from kernel to user space and into php memory only to be copied back to the kernel again to write to disk.

In the existing code, during an upload, the chunk data is sent to a PUT REST endpoint which then passes that to the selected "storage" back end in FileSender. The data is made available to php on stdin.

The existing FileSender code reads all the data for an HTTP request into memory at `RestServer::process()` time with a call to `Request::body()`. For a default 5mb chunk size this cases memory allocation of 5mb and data to be copied into that buffer.

For a concrete example the data may be eventually passed to `StorageFilesystemChunked::writeChunk()` which takes the data to save as a method parameter and uses the size of that data as how much data to write to disk.

This PR delays reading input data to the last moment possible. The update in this PR takes the expected size of the chunk from the client and does not read the data into PHP memory at all. The data is indirectly passed to `stream_copy_to_stream` from "php://input" along with the file descriptor to write that data into. This should give `stream_copy_to_stream` the chance to directly use the `copy_file_range(2)` call to copy that data directly in-kernel without needing to allocate or copy data to the user address space at all.

See for example:
https://github.com/php/php-src/blob/00c0a9ba308c899e08a6e959b669092b9b976b1b/main/streams/streams.c#L1680

Apart from code restructuring to make this possible there are some other effects to this change. We are relying on the client to report how large the data to write is going to be. We can check that that reported size is below the expected chunk size but can not know exactly how much data has been supplied until after the PHP `stream_copy_to_stream` call has returned. That function tells the caller how much data was actually written. So we must do the size validation a few times, the first time using the client supplied "size" information and the second time using the actual size of the data provided.

A specific issue with the StorageFilesystemChunked implementation is that `verifyChecksum` can not work as it stands. The current implementation of `verifyChecksum` expects to have the entire chunk passed as an in memory data buffer to verify the written data. This has resulted in an update to optionally have the client calculate the sha256 if we want to have both storage_filesystem_hash_check and performance_allow_direct_copy_from_put_to_disk set. That will result in a performance hit for the clients using this combination but the checksums can still be verified on the server on a per chunk basis.

### DIFF
--- a/classes/data/File.class.php
+++ b/classes/data/File.class.php
@@ -524,6 +524,37 @@ class File extends DBObject
     }
     
     /**
+     * Store a chunk from PUT input to an offset in the file
+     *
+     * @param int chunkSize the chunk data size
+     * @param int $offset the chunk offset in the file, if null appends at end of file
+     */
+    public function writeChunkDelayed($chunkSize, $offset = null)
+    {
+        if (!$this->upload_start) {
+            $this->upload_start = time();
+            $this->save();
+        }
+        
+        $res = Storage::writeChunkDelayed($this, $chunkSize, $offset);
+
+        // Update transfer's last chunk time so the cleanup cron can detect abandoned
+        // uploads vs active ones. Throttled to once per minute to avoid an extra
+        // UPDATE on every single chunk (default 5 MB chunks → thousands per large file).
+        // Since cleanup is measured in days, minute-level precision is more than enough.
+        if (!$this->transfer->last_chunk_time || (time() - $this->transfer->last_chunk_time) >= 60) {
+            $this->transfer->last_chunk_time = time();
+            $this->transfer->save();
+        }
+        
+        Logger::info($this.' chunk['.((int)$offset).'..'.((int)$offset + $chunkSize).'] written'.(Auth::isGuest() ? ' by '.AuthGuest::getGuest() : ''));
+        
+        return $res;
+    }
+
+    
+    
+    /**
      * End file upload
      */
     public function complete()
--- a/classes/rest/RestServer.class.php
+++ b/classes/rest/RestServer.class.php
@@ -108,7 +108,22 @@ class RestServer
             
             // Request data accessor
             $request = new RestRequest();
-            
+
+            if(Config::isTrue('performance_allow_direct_copy_from_put_to_disk')) {
+                // do not try to read the input for a PUT to file chunks
+                // it will handle that itself.
+                if( $endpoint == "file"  ) {
+                    if( $method == "put" ) {
+                        if (array_key_exists('PATH_INFO', $_SERVER)) {
+                            if( strstr($_SERVER['PATH_INFO'], '/chunk/')) {
+                                Request::$delay_all_reading = true;
+                                Logger::debug("RestServer delaying reading any data for this specific request to later...");
+                            }
+                        }
+                    }
+                }
+            }
+
             // Because php://input can only be read once for PUT requests we rely on a shared getter
             $input = Request::body();
             
--- a/classes/rest/endpoints/RestEndpointFile.class.php
+++ b/classes/rest/endpoints/RestEndpointFile.class.php
@@ -328,6 +328,113 @@ class RestEndpointFile extends RestEndpoint
         }
     }
 
+    /**
+     *
+     * Helper method called by put() method only.
+     * 
+     * put() $mode == chunk might like to verify the chunk size twice.
+     * Firstly to make sure the client has not sent bogus data from
+     * the size they are claiming. Secondly, with this mildly verified
+     * client data size (not too large etc) the data is written to
+     * disk using that size claim. Then the written size is passed in
+     * with $szIsFromWrittenDataBytes true and a second verification
+     * is performed using the number of bytes stored on disk for the
+     * PUT.
+     *
+     * $sz is either the client claimed bytes to write or the number of bytes written
+     * $szIsFromWrittenDataBytes is true if bytes have been written and $sz is thus accurate.
+     * 
+     */
+    public function putSizeChecks( $id, $file, $client, $sz, $mode, $offset, $szIsFromWrittenDataBytes )
+    {
+        $data_length = $sz;
+
+        if ($file->transfer->options['encryption']) {
+
+            // Calculate the correct length
+            $chunkLength = $sz;
+
+            switch( $file->transfer->key_version ) {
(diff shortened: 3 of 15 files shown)


WebSpider:
Excited to see what this brings on a S3 backend! 👍

monkeyiq:
It is only effective on the following Storage types at the moment...
```
$config['storage_type'] = 'filesystemChunked';
$config['storage_type'] = 'filesystem';
```

S3 is harder for me to test as I've never actually run FileSender against S3.

madsi1m:
Interesting, aren't you still reading the $data here? https://github.com/monkeyiq/filesender/blob/1fa77c108eba741ce748aeabb0b5ee18bbe473ff/classes/rest/endpoints/RestEndpointFile.class.php#L518

yent:
Very interesting, this is one major drawback we experienced for going production using S3, this is especially bad when dealing with 40m chunks ... How do you plan on dealing with signed requests ? Is it possible to rewind php://input ? This is the same problem as chunk validation I guess.

monkeyiq:
@madsi1m The RestServer explicitly shorts reading the contents if `performance_allow_direct_copy_from_put_to_disk` is set in the config and it is a PUT to the `File` endpoint. It is safe to do this as the REST code will return `''` if that is the case. 

If that setting is in effect the code might see a blank string trying to read it. But the code needs to know what to do here in order to leave that data on stdin to the last minute. Only the two storage backends I mention above currently know how to handle that case.

monkeyiq:
>Very interesting, this is one major drawback we experienced for going production using S3, this is especially bad when dealing with 40m chunks .

@yent I had found this optimization a little while ago and mentioned it to the FileSender board. For sites using larger chunks it will be very painful at the moment. 40mb chunk, 6 chunks in flight with TeraSender would possibly result in 200mb RAM use per client and maybe twice that if the chunk is being sent to the S3 library.

> .. How do you plan on dealing with signed requests ? 

I think the other metadata should be available as normal. If the sig can not be read then I am happy to add that to the http headers as the optimization sort of makes a little uncomfortable details ok.

I have tested this a little already with TS enabled/disabled and with the two storage back ends, and with encryption on and off from the browser.

> Is it possible to rewind php://input ? This is the same problem as chunk validation I guess.

I imagine you can only read forward from `php://input` as one might consider it sort of like stdin.

For the cloud back ends this optimization should result in less RAM congestion once it is deployed there too. 

As for chunk validation, the claimed chunk from the client size is tested and needs to be valid for all internal chunks (directly the exact size) and the last (exactly the remaining bytes). If a client wants to send less or more than they claim it will cause an error. Sending more is ok as the copy is limited to the chunk size so extraneous bytes will not be accepted.

monkeyiq:
I added an explicit turn off for other `storage_type`

yent:
@monkeyiq true, right now S3+40m chunks along with terasender is a RAM nightmare :)

I asked about signed requests (remote authentication) as the request body is part of the signature, so the body is needed twice, once for verifying the signature, once for actually storing the chunk ... I don't have a v3 test instance but I get the feeling that the current state of the PR breaks remote authentication, thus breaking the provided python client ...

I cannot find conclusive evidence that you can only read php://input once, I know that this was the case in the past but it was mentioned in PHP's documentation, it is not anymore so I'm wondering, have to do some testing.

yent:
@monkeyiq Ok, testing was fast :)

You can actually rewind php://input as many time as you want, at least with php8.5, most likely with any php8+.

This is a really good thing as this would allow to read once to validate remote auth, a second time to validate chunk integrity and a last time to send it to S3.

And passing a file handle around would allow to do sub-chunk reading (or even stream copy as mentioned before), thus keeping the memory profile to a minimum.

Last thing would be to be able to pass the file handle to the S3 lib instead of full chunk data.

I have a working (quick and dirty) implementation using curl directly instead of the bloated aws lib.

Libcurl supports reading from a file handle out of the box when the aws lib takes full chunk and even copies it in tmp before sending, resulting in a 40m chunk using up to 120m of RAM per terasender client ...

Quick math, we have up to 500 simultaneous uploads, 4 terasender per, using S3 would require at least 240G of RAM just for upload, so we are kind of eager to see improvements, especially with the current state of the RAM market :O

monkeyiq:
I have disabled the optimization for the python client for now.  This is one of the upsides to the config switch which allows easy turn on/off (https://github.com/filesender/filesender/pull/2640/commits/18ed2e7ac0878032ce64f8bf56ea4f6388e6a6c3) Best to be able the catch all the web browser requests to knock out a bunch of useless RAM/CPU use for a starter.

We can perhaps then look at a subsequent S3 PR to allow turning on and off `writeChunkDelayed` for the S3 backend. Maybe move `performance_allow_direct_copy_from_put_to_disk` to using an array of `storage_type` to apply to for more elegance. I might add that bool to array change here before docs and merging.
 
It might be interesting to move to using the `writeChunkDelayed` with the existing S3 lib first and then maybe allow direct client optimization using curl subsequently. At least when we add delayed chunks for S3 there should be some RAM relief already.

monkeyiq:
I would also like to investigate using SAML+ECB as an option for the python client to not have to brute force sign all the data as a way to authenticate.

monkeyiq:
I have documented performance_allow_direct_copy_from_put_to_disk. I might leave that as a boolean as a site is likely to only have one storage backend in use and if the feature doesn't work for any reason they are probably happy to revert to pre `3.8` setups for all requests.

monkeyiq:
I get a bucket already exists error with minio but the rest of the delayed write code seems to work for s3 now. The filesender.py doesn't use them because authentication = whole envelope on data for some reason.

PR #2645, opened 2026-03-27
Add the hour +59m59s of same day to getMaxExpire()

madsi1m:
Since adding the hour to expire date, when picking the maximum day from the datepicker causes a BadExpireException because getMaxExpire() returns the unix time for the day only.

This change fixes this, but please double check the logic.

### DIFF
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -702,7 +702,7 @@ class Transfer extends DBObject
             $days = Config::get('default_daysvalid');
         } // @deprecated legacy
         
-        return strtotime('+'.$days.' day');
+        return strtotime('+'.date('G').' hour', strtotime('+'.$days.' day'))+3599;
     }
     
     /**
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -703,7 +703,7 @@ class Transfer extends DBObject
         } // @deprecated legacy
 
         if (!empty($_COOKIE['x-filesender-timezone'])) {
-            return strtotime('+'.$days.' day') + (new Utilities)->getTimezoneOffset($_COOKIE['x-filesender-timezone']);
+            return strtotime('+'.$days.' day') + Utilities::getTimezoneOffset($_COOKIE['x-filesender-timezone']);
         }
         return strtotime('+'.$days.' day');
     }
--- a/classes/data/Transfer.class.php
+++ b/classes/data/Transfer.class.php
@@ -701,8 +701,11 @@ class Transfer extends DBObject
         if (!$days) {
             $days = Config::get('default_daysvalid');
         } // @deprecated legacy
-        
-        return strtotime('+'.date('G').' hour', strtotime('+'.$days.' day'))+3599;
+
+        if (!empty($_COOKIE['x-filesender-timezone'])) {
+            return strtotime('+'.$days.' day') + (new Utilities)->getTimezoneOffset($_COOKIE['x-filesender-timezone']);
+        }
+        return strtotime('+'.$days.' day');
     }
     
     /**
--- a/classes/utils/Utilities.class.php
+++ b/classes/utils/Utilities.class.php
@@ -208,7 +208,7 @@ class Utilities
      *
      * @return int offset in seconds
      */
-    function getTimezoneOffset($custom_timezone_string) {
+    public static function getTimezoneOffset($custom_timezone_string) {
         $server_tz = new DateTimeZone(Config::get('default_timezone'));
         $custom_tz = new DateTimeZone($custom_timezone_string);
         $datetime = new DateTime("now", $server_tz);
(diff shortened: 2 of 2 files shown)


victoritis:
When happens this to you? I have tried it and works correctly for me. I tried with a new tranfer and a new invitation.

victoritis:
Hi @madsi1m, thanks for looking into this.

I tested this scenario (selecting the maximum allowed day in the datepicker for both transfers and invitations) and it works correctly for me without any `BadExpireException`. I have PR #2624 and https://github.com/filesender/filesender/pull/2630 applied on my prod.

I also want to flag a potential issue with the proposed fix. `strtotime('+N day')` in PHP uses `time()` as its base, so it already returns the **current time + N days**, including the current hour — it does **not** return midnight. You can verify this easily with:

```php
echo date('Y-m-d H:i:s', strtotime('+7 day'));
// At 14:00 → outputs "2026-04-03 14:00:00", not "2026-04-03 00:00:00"
```

Because of this, the proposed fix:

```php
return strtotime('+'.date('G').' hour', strtotime('+'.$days.' day')) + 3599;
```

adds `date('G')` hours **on top of** a timestamp that already includes the current hour. For example, at 14:00 this would return `now + N days + 14 hours + ~1 hour`, which is roughly **15 extra hours** beyond what's intended.

That said, I think the underlying issue you may be experiencing could be related to a **client-server clock skew** — if your browser's clock is even slightly ahead of the server's clock, and you select the maximum allowed day, the JS timestamp could exceed `getMaxExpire()` by a few seconds, triggering the exception. This wouldn't happen if the clocks are well synchronized, which may explain why I can't reproduce it.

Could you check if your client machine's clock is slightly ahead of the server's? That would help confirm whether this is the root cause.

madsi1m:
Interesting, we run servers in UTC and local time is about +10 hours, I wonder if that has any relation.
Any ideas @monkeyiq? I remember you did some offset math for timezones in the past

madsi1m:
I agree @victoritis above logic was wrong. I assume you are running yours with `client_send_current_timezone_to_server = false` which might explain why it works for you.

Have added the offset of users timezone which seems to work for me.

@victoritis and @monkeyiq please double check I haven't missed anything.

monkeyiq:
I am happy with this. One thought, perhaps some users might not allow or send the cookie. Maybe we should have things be a bit more lenient by default and/or have the ability for the sys admin to set a default time offset if the cookie is not present. 

Disk is relatively cheap so if something hands around for 1-23 hours longer it probably doesn't effect most installations so we could also have the ability to configure things to go back to how it was before.

victoritis:
Hey @madsi1m @monkeyiq  — I'm on vacation for the next 8 days and can't look at the code right now, so I'm going a bit off memory here and I'll admit I'm getting a bit confused trying to follow everything.

That said, I think I understand what was happening. Let me try to lay it out to make sure I'm not missing something:

**My case (Spain, ~UTC+1/+2, presumably `client_send_current_timezone_to_server = false`):**
It's 18:00 in Spain. I create a transfer and pick the maximum expiry (say 7 days). My JS calculates the Unix timestamp correctly: 18:00 local = 16:00 UTC, so it sends the timestamp for 16:00 UTC on day X+7. The server (running in UTC) calculates `getMaxExpire()` as now + 7 days = 16:00 UTC on day X+7. Comparison: 16:00 ≤ 16:00 → all good. Both client and server are speaking pure UTC, no timezone adjustments involved.

**@madsi1m's case (UTC+10, `client_send_current_timezone_to_server = true`):**
It's 02:00. On the server it's 16:00 UTC (previous day). Because the timezone cookie is active, some part of the code internally adjusts the expiry timestamp using the user's timezone offset. So the expiry timestamp that reaches validation ends up shifted by +10 hours. But `getMaxExpire()` was NOT doing that same adjustment — it was still calculating the max as plain server time + N days. So the adjusted expiry exceeded the non-adjusted max → BadExpireException.

The fix (commit 2) makes `getMaxExpire()` also add the timezone offset when the cookie is present, so both sides of the comparison are consistent.

Is that understanding correct? If so, the fix makes sense to me..

I'll take a closer look at the code when I'm back from vacation. But i will be active for response here.

victoritis:
@madsi1m might be worth adding a quick debug log in the validation to see exactly what's being compared — i.e. the expiry timestamp coming from the client vs the value returned by `getMaxExpire()`. That way we can confirm whether the mismatch is indeed UTC+offset vs plain UTC.

victoritis:
Hey @madsi1m @monkeyiq — I've reviewed the fix and I can confirm that my analysis was correct.

The root cause is indeed a timezone asymmetry: when `client_send_current_timezone_to_server = true`, the client calculates the expiry timestamp based on the user's local day, but `getMaxExpire()` was calculating the maximum using only the server's UTC time. For users in timezones significantly ahead of the server (like UTC+10), the client could offer a date in the datepicker that the server would then reject as exceeding the maximum — hence the `BadExpireException`.

The fix correctly compensates for this by adding the user's timezone offset to `getMaxExpire()` when the cookie is present.

However, I noticed that `Guest::getMaxExpire()` in `classes/data/Guest.class.php` (line 348) was **not included in this fix** and still uses the plain `strtotime('+N day')` without considering the timezone offset. This means the same `BadExpireException` can occur when creating or editing guest invitations under the same conditions.

PR #2653, opened 2026-04-03
auditlog when recipients are added

makinog3:
When recipients are added to a transfer, it is not recorded in the audit log.

I'm not sure what FileSender's policy is regarding what should be recorded in the audit log, but our user has requested that it be recorded when a recipient is added to a transfer.
And I'm not sure how to add new text to POEditor, too.

### DIFF
--- a/classes/constants/LogEventTypes.class.php
+++ b/classes/constants/LogEventTypes.class.php
@@ -70,13 +70,14 @@ class LogEventTypes extends Enum
 
    
    /* TRANSFER */
-    const TRANSFER_STARTED         = 'transfer_started';         // Transfer started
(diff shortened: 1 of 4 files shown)


monkeyiq (approved):

monkeyiq:
Adding new terms is a bit manual at the moment. It is scripted but isn't really available to PRs to automatically feed back into poeditor. For the moment I am very happy if you mention that new term(s) are added and I will make sure they get into poeditor for the pull request.

Thank you for the update.

monkeyiq:
terms added, en_au and jp translations are in poeditor now so will be found in the next import poeditor -> github.

Issue #2601, opened 2026-03-05
CID Validation Mismatch Issue


flowdnb:
## Description

The FileSender backend validation is failing for CIDs generated by the frontend, causing "Validate FAILING. REST data filtering failing for transfer.files.cid" errors in logs.

## Environment

- FileSender version: 3.6
- Browser: 145.0.7632.159 arm64
- macOS: 26.3 (25D125)

## Steps to Reproduce

1. Upload a file through the web interface
2. Observe the logs for validation errors

## Expected Behavior

The CID format generated by the frontend should match the backend validation regex pattern.

## Actual Behavior

The validation fails with:
```
[rest:info] Validate FAILING. REST data filtering failing for transfer.files.cid
```

## Request Payload Details

From browser developer toolbar, we see:
```
"cid": "file_1772723908179_41_73525_578957"
```

The frontend generates CIDs with format: `file_{timestamp}_{filename_length}_{filesize}_`
The backend expects: `file_{number}_{number}_{number}_`  

The timestamp (1772723908179) is too large and breaks the expected 3-segment pattern.

## Proposed Solution

Update the PHP regex validation in `classes/rest/endpoints/RestEndpointTransfer.class.php` line 565 to accommodate the 4-component CID format, or modify the frontend CID generation to match the expected backend format.

monkeyiq:
Thank you for the report. This should be fixed with https://github.com/filesender/filesender/pull/2602.

I have done a local test on that with the input data supplied. The regex was not allowing for the nonce that is tagged onto the cid after creation.

PR #2660, opened 2026-04-08
statistics page: change tables order by column position, rather than name

madsi1m:
It seems psql processes the ORDER BY before the AS and cant sort by labels so we use column position instead

### DIFF
--- a/www/lib/tables/statistics_page.php
+++ b/www/lib/tables/statistics_page.php
@@ -53,9 +53,9 @@ function is_encrypted_to_html( $v ) {
 
 switch ($topic) {
     case 'top_users':
-        if (!in_array($sort,['User','Transfers','Size','Downloads']))
-            $sort='User';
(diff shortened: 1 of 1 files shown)


madsi1m:
Be nice to have this tested in a psql as I can only test in mariadb

WebSpider:
I can set up a psql18 today. What would be the best way to test this?

madsi1m:
@WebSpider as an admin click the statistics tab and on the bottom of the page are 4 tables, click its headers to change how it sorts. You may want to populate the db with some data before testing in order to see things in the tables

monkeyiq:
An alternate that doesn't need the colnum fudge is in https://github.com/filesender/filesender/pull/2661

WebSpider:
Considering @monkeyiq already tested it in #2601 my tests seem redundant

madsi1m:
yep i like https://github.com/filesender/filesender/pull/2661 more!

PR #2661, opened 2026-04-08
quoting of select as for many db types

monkeyiq:
We can also select ... `as foo` instead of `as "foo"` and things seem to work on both pg and mariadb.

This is an alternate to https://github.com/filesender/filesender/pull/2660.


monkeyiq:
The case folding isn't really needed for a minimal patch but I think perhaps since tables all have caps to start that might help with the mariadb world.

monkeyiq:
I have switched mariadb <-> postgresql a few times and this seems ok in both.

PR #2665, opened 2026-04-10
Fix Turkish mail templates subject header key

emresde:
This PR updates Turkish mail templates to use `subject:` instead of `konu:`.

Without this change, subject lines in Turkish mail templates may not be parsed correctly by the mail template system.

Changes were applied to the Turkish template files under `language/tr_TR/`.

### DIFF
--- a/language/tr_TR/bounce_report.mail.php
+++ b/language/tr_TR/bounce_report.mail.php
@@ -7,7 +7,7 @@
 // 
 // 
 ?>
-konu: İleti teslim hatası
+subject: İleti teslim hatası
 
 {alternative:plain}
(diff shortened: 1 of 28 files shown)


monkeyiq:
Thank you for the PR.

The files in `language` are recreated when an import is done from poeditor. I am happy to move these changes to there so they will survive future language updates as well.

monkeyiq:
I have updated poeditor. This PR will bring in the same updates from there.
https://github.com/filesender/filesender/pull/2668/changes

Thank you for your update. Though the PR itself is not merged the contents of it have been made permanent.

PR #2669, opened 2026-04-13
Faster downloads

madsi1m:
I have a user trying to download 100gb from the other side of the world and wget seems to keep failing "connection".

While debugging this I noticed that a single wget call would trigger `$read_range();` twice. Closer inspection seems that wget does a HEAD request before the GET.

While this is probably fine for most files, for large files on a poor connection wget fails the HEAD request due to timeouts.

This change skips single file download HEAD requests from reading chunks twice.

### DIFF
--- a/www/download.php
+++ b/www/download.php
@@ -441,6 +441,10 @@ function downloadSingleFile($transfer, $recipient, $file_id, $recently_downloade
 
         header('Accept-Ranges: bytes');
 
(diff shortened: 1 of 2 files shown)


madsi1m:
@monkeyiq not sure if i put the HEAD check in the right place and/or if we need it for other downloads like ranges and archives, but this is a good start that helps my user now.

@monkeyiq also fixed through a merge conflict a table sort in stats page that was missed.

PR #2670, opened 2026-04-13
Fix Partial downloads

madsi1m:
When using wget -c on an interrupted download wget fails.

This PR fixes that by fixing the HTTP Headers and upgrading the StorageFilesystemChunkedStream to be more flexible. It also means we no longer need to force download_chunk_size=upload_chunk_size.

It seems while this fixes partial downloads, it breaks encryption downloads which needs investigating. @monkeyiq any ideas? For now I have made it use the old method so encryptions work. I bet I am missing some chunksize encryption logic somewhere, maybe in the StorageFilesystemChunked class

### DIFF
--- a/www/download.php
+++ b/www/download.php
@@ -332,6 +332,7 @@ function downloadSingleFile($transfer, $recipient, $file_id, $recently_downloade
         if($transfer->options['encryption'] == 1){
             $end = $file->encrypted_size;
             $chunk_size = $file->crypted_chunk_size;
+            stream_set_chunk_size($stream, $chunk_size);
         }else{
             $end = $file->size;
         }
@@ -344,12 +345,8 @@ function downloadSingleFile($transfer, $recipient, $file_id, $recently_downloade
             
             Logger::debug('Send chunk at offset ' . $offset . ' with length ' . $length);
             
-            // TODO Encryption seems to not like the streams at the moment, should fix this but have a workaround
-            if ($transfer->options['encryption'] == 1) {
-                echo $file->readChunk($offset, $length);
-            } else {
-                echo stream_get_contents($stream, $length, $offset);
-            }
+            //echo $file->readChunk($offset, $length);
+            echo stream_get_contents($stream, $length, $offset);
             
             // TODO Log download progress ?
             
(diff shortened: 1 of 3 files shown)


monkeyiq:
IIRC the `download_chunk_size` is a relic from the past already. This happened around the time the chunk size was stored in the database when a transfer was being made.

Transfer sets these in the constructor which is what should then be used by the code in preference.
```
        $this->chunk_size = Config::get('upload_chunk_size');
        $this->crypted_chunk_size = Config::get('upload_crypted_chunk_size');
```

Which is done for example in `StorageFilesystemChunked::readChunk`
```
        $chunk_size = $file->transfer->chunk_size;
```

That existing call in `Chunked::getStream` that references Config::get('upload_chunk_size') should be using `$file->chunk_size;` instead. It is a bug.

monkeyiq:
I will make a PR to clean up these references to Config::get('upload_chunk_size') from the Chunked storage.

monkeyiq:
There are likely to be also some cases where the code is assuming you are wanting to read an entire chunk and read from the start of a chunk.

madsi1m:
fixed encryption with streams.

madsi1m:
I thought download_chunk_size was to make downloads faster by tuning how much to read as read/write speeds can be different on storages. A quick search looks like it isn't used anywhere really, so removed it from my change.

Maybe we should bring it back in the future?

www/download.php
```diff
@@ -253,6 +253,12 @@ function downloadSingleFile($transfer, $recipient, $file_id, $recently_downloade
     if(!$file->transfer->is($transfer))
         throw new FileNotFoundException(array('transfer_id : ' . $transfer->id, 'file_id : ' . $file_id));
 
+    $stream = Storage::getStream($file);
```
monkeyiq: This is likely a duplicate from when things were moved around. The stream is opened again below. Also in both cases something other than `ForwardException` would be better to throw.

monkeyiq:
I guess if there is some sort of priming for data size the Storage engine could do that. Like reading a whole chunk to assume that the kernel will keep it in kernel cache if the user only wants the first bit of it. I think there are likely to be other performance opimizations along the lines of the delayed PUT read which might have more impact.

www/download.php
madsi1m: I think i copied the ForwardException from somewhere else in the code that uses streams. Happy for you to change it, just make it consistent across the lot.

Cleaned up the old one that I shouldn't have pushed

PR #2675, opened 2026-04-14
Fix integer conversion for upload_chunk_size and sizes

jvhaarst:
The API might return strings, which then makes the CLI crash. These int casts fix that


madsi1m:
I recommend using the client from development3 branch as the one in master is not up to date.

https://github.com/filesender/filesender/blob/development3/scripts/client/filesender.py

monkeyiq:
Thank you for the PR. I have taken the changes here and put them into https://github.com/filesender/filesender/pull/2682/ made against `development` which is the 2.x development branch.

Because of how to project wishes to maintain the master/development (and master3/development3) branches one can not simply retarget this PR here to development.

PR #2686, opened 2026-04-29
Add named file(s) download specified with regex

tomgw:
[ UPDATE: This only works for files sent by your self or by your guest users, not for files sent to you by other filesender users ! ]

I added named file(s) download specified with regex to facilitate scripted downloads of known file names and filename patterns.

When downloading files from filesender, the transfer URL may be unknown in contrast to the file name or the pattern(s) of the filename(s) that should be downloaded. Alternatively, it may be necessary to download only a known subset of the files in a known transfer URL. Two new switches enable this behavior.
-l   --list_files    : list all file names across al available transfers
-n --named_file  : only download the file name(s) that comply with the specified regular expression.

examples:
```
filesender.py -n 'data-01.zip'  -d <URL>  :  download the file 'data-01.zip' in the specified transfer <URL>
filesender.py -n 'data-.*\.zip'   :  download all data-*.zip files across all available transfers
filesender.py -n '*\.zip|.*\.pdf  :  download all .zip files and all .pdf files from all available transfers
```

### DIFF
--- a/scripts/client/filesender.py
+++ b/scripts/client/filesender.py
@@ -130,10 +130,12 @@ class FileSenderCLI:
         parser.add_argument("-i", "--insecure", action="store_true")
         parser.add_argument("-p", "--progress", action="store_true")
         parser.add_argument("-s", "--subject")
+        parser.add_argument("-l", "--list_files", action="store_true")
         parser.add_argument("-m", "--message")
         parser.add_argument("-g", "--guest", action="store_true")
         parser.add_argument("-e", "--encrypted")
         parser.add_argument("-d", "--download" )
+        parser.add_argument("-n", "--named_file")
         parser.add_argument("-o","--output_dir" )
         parser.add_argument("--days", type=int)
         parser.add_argument("--threads")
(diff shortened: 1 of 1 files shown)


madsi1m:
I like the regex idea.
Can you please remove the .settings/.project from the PR?

madsi1m:
Does /transfer/@me show transfers I've uploaded or transfers sent to me?

monkeyiq:
The `@me` should be things the user has uploaded.

tomgw:
> I like the regex idea. Can you please remove the .settings/.project from the PR?

yes, i'll remove, it was in there by mistake

tomgw:
> Does /transfer/@me show transfers I've uploaded or transfers sent to me?

Wauw , that is silly of me. The API says 'List of user available transfers' so I assumed it was both. I' ve been sending files to myself (because I'm going to transfer files between my own machines). Apparently @me only shows what I've sent, not what I received from someone else. That narrows down the use case considerably. It will show transfers sent by guest users.

PR #2688, opened 2026-04-30
Update known-installs.md

Olga-Popcova:
Annual check of all (listed) FileSender installations.


WebSpider (approved):
Nice update, thanks! 

Is there a moment when unreachable installs are deemed historic?

Olga-Popcova:
> Nice update, thanks!
> 
> Is there a moment when unreachable installs are deemed historic?

I am trying to ask an unreachable installation to confirm that they do not plan to restore the FileSender service, before moving them to "decommissioned"... Sometimes they do reply, sometimes - do not... Sometimes it's clear just that an organisation (as it was previously) does not exist anymore... So, it's not easy... :)

PR #2690, opened 2026-05-05
remove support for lteq php 5.4

monkeyiq:


madsi1m:
did you also just notice this from the last pr?

monkeyiq:
maybe ;)

monkeyiq:
This old PHP support was needed before the UI test execution was moved to codeception https://github.com/filesender/filesender/pull/2090 

That PR itself uses more "framework" stuff but is only needed during testing.

PR #2692, opened 2026-05-05
feat: site announcement on logon_page

makinog3:
Our site sometimes wants to display an announcement to users, for example,  a scheduled maintenance notice.
In version 2, the CSS classes `info` and `message` were used to display an announcement on the logon_page.

  - Add .site-announcement CSS classes (blue/yellow/red variants)
  - Add templates/announce.php to display language-aware announcements
  - Include announce.php in logon_page.php
  - Add language/*/announce.html.php sample (empty by default)

PR #2693, opened 2026-05-05
fix: Lang translation fallback

makinog3:
Lang::translate() not falling back to English when translation is missing

### DIFF
--- a/classes/utils/Lang.class.php
+++ b/classes/utils/Lang.class.php
@@ -610,22 +610,14 @@ class Lang
             $stack = self::getCodeStack();
             Logger::warn('No translation found for '.$id.' in '.$stack['main'].' language');
 
-            $fallbackid = 'unknown';
-            // fallback is an array of lang codes, so loop through those
-            foreach (self::$translations['fallback'] as $fallback_lang) {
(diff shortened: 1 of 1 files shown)


makinog3:
During testing for #2692, I noticed this issue because, although the default en_AU announcement should be displayed if no language-specific announcement exists, it wasn't.

However, I'm not confident that my fix is ​​correct...

monkeyiq:
Thank you for the PR. I am always happy to also dig into things a bit to also verify if you are not confident in the offered solution.

I think some things are complicated with poeditor because it offers the English fallback for untranslated terms on the site so those will can imported into the `language/en_AU` tree during imports.

The fallbacks here is indeed a term -> string direct mapping:

https://github.com/filesender/filesender/blob/e314dbdc9412d73f0bf9fbe6781840cc6d6893e4/classes/utils/Lang.class.php#L567

As the language codes are built as a stack in places such as `self::getCodeStack` I think there might have been an expectation that the fallbacks were still using an array of 
```
[lang-name][term] = show-string-value-for-term
```

I have tested your update with various settings and works for me

* term in jp_JP lang
* term not in jp_JP but in en_au
* config in jp_JP directly
* config not in jp_JP but in en_au only

PR #2704, opened 2026-05-12
feat: add Ukranian translation

WebSpider:
This adds the reference to the Ukranian translations from the FileSender 3.0 POeditor project.


monkeyiq:
I assume that the only real change to `scripts/language/import-langs-directory.sh` was the uk line not the change to all the rest.

Melkin1968:
Hi! Could you please check scripts/language/import-langs-directory.sh? I think there's a typo: the Ukrainian locale is listed as uk_UK instead of uk_UA. Thanks!"

monkeyiq:
This should be fixed now with https://github.com/filesender/filesender/pull/2708

Please let me know if something else was missed.

PR #2712, opened 2026-05-26
Header-based authenticator compatible with SURF Research Cloud

sunetfreitag:
I have developed a custom authenticator that uses header-based authentication which is compatible with SURF Research Cloud when provisioned using the Ansible playbook here: https://github.com/sunetfreitag/src-playbooks/tree/main/filesender

It uses the username in the header Remoteuser to log on the user automatically. Configuration files for the docker deployment have been updated accordingly.


WebSpider:
CI failure is due to GitHub stability issues.

WebSpider:
One thing to note: With the current setup, anyone being able to access the FileSender instance would be able to inject this header, enabling the use of this header as a source for bruteforce attacks.

There is nothing in your nginx config preventing or limiting this, as far as I can see.

docker/filesender-config.js.php
```diff
@@ -162,23 +157,6 @@ function value_to_TF( $v )
     crypto_pbkdf2_delay_to_show_dialog: '<?php echo Config::get('crypto_pbkdf2_delay_to_show_dialog') ?>',
     crypto_pbkdf2_dialog_custom_webasm_delay:'<?php echo Config::get('crypto_pbkdf2_dialog_custom_webasm_delay') ?>',
 
-	language: {
```
monkeyiq, line 165: I am not sure we need to remove these in the docker file? If they are not hurting anything wouldn't it be a good choice to try to make this file as close to `www/filesender-config.js.php` as is reasonable? Being able to see the diffs that matter at a glance.

docker/assets/opt/filesender/filesender/config/config.php
```diff
@@ -110,8 +111,16 @@
 // If you want to overide the SAML simplephp configuration defaults parameter,
 // uncoment and edit the following lines
 //
-// // Authentification type ('saml' or 'shibboleth')
-$config['auth_sp_type'] = 'saml';
+// // Authentification type ('saml' or 'shibboleth' or 'header')
+$config['auth_sp_type'] = 'header';
```
monkeyiq, line 115: If we are moving away from a main sp_type here we should clearly mention that in the docker/README file so folks using this know it is using a specialized auth rather than the same auth that might be used in a normal deployment.

monkeyiq (changes requested):

sunetfreitag:
Surf Research Cloud has their own nginx configuration, which should prevent header injection. The authentication configuration for a SRC-based service can be found in this playbook here: https://gitlab.com/rsc-surf-nl/plugins/plugin-nginx/-/blame/main/plugin-nginx.yml?ref_type=heads#L97

That is why I did not add an explicit nginx configuration that would prevent header-injection. One option could be to rename the class to ensure that it is to be used in a Research Cloud context. 

I have a demo instance, feel free to test the header-injection scenario here: 
https://fsdemo01.researchcloudtu.ws.cloud.sunet.se

docker/filesender-config.js.php
sunetfreitag: I remember I had some issues with this in a previous version, where I had exceptions in the local instance. I will add this part again and also review the other options. Ideally, it should work with the ootb-version of the file.
While testing this, I found a very minor typo for the session_expired_warning: "Please logon to contiunue."

docker/assets/opt/filesender/filesender/config/config.php
sunetfreitag: I can add this to the README of the docker file.

Issue #2710, opened 2026-05-22
transfer details are shown twice on download page


Meindert-van-den-berg:
Filesender 3.9
php 8.5

When I open a download link than at the upper left the transfer details are shown.
But at the end of that page the information is shown again.

Is there a directive to suppress that last information?

Thanks.

<img width="1552" height="1470" alt="Image" src="https://github.com/user-attachments/assets/0a855ab3-68f4-445d-88b6-38a802c94513" />

monkeyiq:
This is likely a bug in the UI refresh. I will remove the lower items that are already shown at the top of the page.

monkeyiq:
Thank you for this report I will close it here as I am fairly sure that 2720 above fixes this. Please reopen / rereport if it pops up again.

PR #2720, opened 2026-05-28
ui3 do not display duplicates

monkeyiq:


monkeyiq:
This was raised here https://github.com/filesender/filesender/issues/2710

PR #2738, opened 2026-06-08
fix: store files_ids in session for POST-Redirect-GET to avoid URL length limit

WebSpider:
When downloading an archive with many files, the redirect to add `transaction_id` put all file IDs (URL-encoded) into a GET URL, causing the webserver to reject the request since the URL-encoding blows up the GET request.

This PR proposes to fix this by storing the file selection in the PHP session under the `transaction_id` key before redirecting.  The redirect URL now only carries `token`, `archive_format` and `transaction_id`.

Since this PR is one of the first times I am actually touching this code, there may be cases I have missed.

A workaround would be to increase the webserver's request limits

### DIFF
--- a/www/download.php
+++ b/www/download.php
@@ -38,7 +38,18 @@ require_once('../includes/init.php');
 
 
 try {
-    
+
+    // Read transaction_id early so we can restore the file selection from the session
+    $transaction_id = '';
+    if(array_key_exists('transaction_id', $_REQUEST))
+        $transaction_id = $_REQUEST['transaction_id'];
+
+    if($transaction_id && Utilities::isValidUID($transaction_id)
+        && !array_key_exists('files_ids', $_REQUEST)
+        && isset($_SESSION['download_fids_' . $transaction_id])) {
+        $_REQUEST['files_ids'] = $_SESSION['download_fids_' . $transaction_id]['fids'];
+    }
+
     // List of files to be downloaded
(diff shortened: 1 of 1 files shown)


monkeyiq:
Is this for encrypted transfers, unencrypted, or both?

Downloading a zip for a non encrypted transfer I see a POST and the file ids being sent as a payload.

For encrypted transfers each file is downloaded in succession for me. It does use a GET for that but each request only has one fileid in it so shouldn't overrun buffers.

It is possible that `transaction_id` is handled differently in our configs or setups.

This is on `development3` HEAD.

WebSpider:
> Is this for encrypted transfers, unencrypted, or both?

The case I have seen was for unencrypted
 
> Downloading a zip for a non encrypted transfer I see a POST and the file ids being sent as a payload.

Yeah, that is not the problem. The problem is when collecting information for zip creation, this then turns into a GET with urlencoded fileid's in my setup.
 
> For encrypted transfers each file is downloaded in succession for me. It does use a GET for that but each request only has one fileid in it so shouldn't overrun buffers.

I have not yet tried to reproduce this for encrypted.

monkeyiq:
I will limit my initial testing to unencrypted. Which version are you testing on?

WebSpider:
> I will limit my initial testing to unencrypted. Which version are you testing on?

Apologies for the delay. We have been testing this on 3.6, 3.8 and 3.9.

PR #2764, opened 2026-06-30
fix: add multiple email addresses as destinations on transfers screen

makinog3:
Co-Authored-By: Kataoka

### DIFF
--- a/www/js/transfers_table.js
+++ b/www/js/transfers_table.js
@@ -125,7 +125,7 @@ $(function() {
             recipients.push($(this).attr('data-email'));
         });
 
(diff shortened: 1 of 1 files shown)


monkeyiq (approved):
This looks good to me. Later in this method the `email_separator_msg` is appended to the dialog which for en.au reads `Multiple email addresses separated by , or ;`  which gives the expectation that many can be added. For this the bootbox inputType: 'email' used by `promptEmail()` is likely not the validation we are looking for.

The old code could lead to users being rightly confused. Thank you for the PR!

Issue #1341, opened 2023-01-16
unable to unzip a zip archive of  encrypted files on Archive Utility of macOS


makinog3:
Archive Utility.app of macOS is unable to unzip a zip archive of multiple encrypted files created on frontend side of FileSender.
Old macOS had not supported the zip64 archive, but now it might have supported it.
Because unencrypted zip64 files created on the backend side of FileSender can be unzipped.

I tested the FileSender 2.30.
attach files:
    * [transfer_209880_files_494a91a9.zip](https://drive.google.com/file/d/1rsqInqsFK93tPwCe8LoNBeWnPaldpJFo) : created on the backend side (unencrypted) -> OK
    * [FileSenderDownload_16-1-2023__16-30-23..zip](https://drive.google.com/file/d/1ZwimUZNTw9ym22tFZdbm2CmJAQrk1B_i) : created on the frontend side (encrypt) -> NG

makinog3:
By the way, renaming the ZIP file will cause a different error.

* FileSenderDownload_16-1-2023__16-30-23..zip -> Unable to expand "FileSenderDownload_16-1-2023__16-30-23..zip" into "Documents"
* rename to FileSenderDownload..zip -> Unable to expand "FileSenderDownload.zip" into "Documents" (Error 79 - Inappropriate file type or format.)

makinog3:
> By the way, renaming the ZIP file will cause a different error.

It was my mistake. The error message is not all shown if the file name is long.

WebSpider:
Does this work with The Unarchiver ?

makinog3:
> Does this work with The Unarchiver ?
Yes, The Unarchiver can unzip that.
However, many of our site users use the macOS standard Archive Utility, so we would like to support it if possible.

monkeyiq:
I currently do not have access to mac osx so can not test updates. 
When I implemented the feature I tested that it worked in the unarchiver.

WebSpider:
Turning this into an enhancement request for consideration.

PR #2765, opened 2026-06-30
fix: add ZIP64 extra field to local file header for macOS compatibility

makinog3:
This was raised in #1341

Local file headers with size fields set to 0xFFFFFFFF must include a ZIP64 extended information extra field per the ZIP64 spec ([APPNOTE](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) §4.5.3). Without it, strict parsers such as macOS Archive Utility see the 0xFFFFFFFF marker, look for a ZIP64 extra field, find none, and fail with an error. Lenient tools (The Unarchiver, 7zip, Windows) tolerate the omission and fall back to the data descriptor directly, which is why the bug was platform-specific.

The fix adds a 20-byte ZIP64 extra field (header ID 0x0001, uncompressed and compressed size both 0 as placeholders) to every local file header. Actual sizes remain in the data descriptor as before (general purpose bit 3 is set). This allows strict parsers to recognise the ZIP64 format and correctly read 64-bit sizes from the data descriptor.

### DIFF
--- a/www/js/zip64handler.js
+++ b/www/js/zip64handler.js
@@ -255,12 +255,20 @@ window.filesender.zip64handler = function() {
             this.writeu16( 0          );  // no compression
             this.writeu32( dts );
             this.writeu32( 0          );  // no crc-32 yet
-            this.writeu32( 0xFFFFFFFF );  // no size yet
-            this.writeu32( 0xFFFFFFFF );  // ...
+            this.writeu32( 0xFFFFFFFF );  // no size yet (ZIP64: value in extra field)
+            this.writeu32( 0xFFFFFFFF );  // ...         (ZIP64: value in extra field)
             this.writeu16( filename_array.length );
-            this.writeu16( 0          );  // no extra data
+            this.writeu16( 20         );  // extra field length: ZIP64 (4 header + 16 data)
 
             $this.write( filename_array );
+
+            // ZIP64 extended information extra field (required when sizes are 0xFFFFFFFF)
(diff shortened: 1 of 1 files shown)


makinog3:
I pull-requested this patch to the `development3` branch. However, we are applying and testing it on both FileSender 2.51 and 3.10.

monkeyiq:
Thank you for the update. I have done initial testing on Linux unzip and the Archive Utility 10.15 (176) on MacOS.

monkeyiq:
I am happy to back port this to the 2.x series as well.

monkeyiq (approved):

monkeyiq:
This is merged into development3 and will be in release 3.11 and beyond.

I have also cherry picked this back into development here https://github.com/filesender/filesender/commit/ed122a15cb6663424ef679c0cb254c21d42f1f68

That will make it also in release 2.69 and beyond. 

Thank you.

monkeyiq:
I should also mention that there are two zip64 creation paths. The one in this PR is used to create the archive when encryption is in use. In that case the zip64 is created in streaming mode in the browser using decrypted data chunks as they arrive.

For non encrypted files the ZipArchive class is used to directly create a zip64 taking the bytes directly off the disk on the server. As no decryption is needed,  the server can directly produce the archive.

lib/vendor/barracudanetworks/archivestream-php/src/ZipArchive.php

monkeyiq:
It seems Archive Utility is happy with the server produced zip64 for non encrypted.

PR #2772, opened 2026-07-08
ability to set a theme override for the site

monkeyiq:
The 'soothing' is a dark like theme which is brown based. The intention is to be less stark than straight "dark" mode and nowhere near as intense on the eyeballs for high output displays.

At the moment it is only selectable in config.php using a custom config key. IIRC the user preference theme selection box still has issues.

### DIFF
--- a/includes/ConfigDefaults.php
+++ b/includes/ConfigDefaults.php
@@ -308,6 +308,7 @@ $default = array(
     'theme' => '',
     'theme_userpref_enabled' => false,
     'openpgp_enabled' => false,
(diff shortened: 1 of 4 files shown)


monkeyiq:
This is not really intended for end user use at this stage. It is more to allow devs to have an easier on the eye experience. Parts of the colours are not so well aligned as one would expect from a proper theme etc.

monkeyiq:
I make no claims to be a colour palette maestro.

madsi1m:
could get inspiration from github colours

PR #2800, opened 2026-08-03
Fix file selection for downloads from transfer details

### LINKED ISSUE #2598: # Bug : Cannot download file from a guest unless I use the URL containing the download token
### Describe the bug
---

I cannot download a file sent by a guest (using a guest voucher) directly from the transfer details page. I need to click on the transfer details page, and then I have to copy the download link, and finally I'm able to download the file.

### To reproduce
---

1. Open the details page of an existing transfer created from a guest upload
2. Open the web browser's dev console
3. Select the file uploaded by the guest
4. Click on the "download" button
5. Observe the error on the dev console

### Expected behavior
---

The file should be downloaded without any issue.

### Screenshots

Image

Image

### Environment
---
FileSender Version : 3.6
Browser: Microsoft Edge 144.0.3719.151

YohanRodriguezMDM:
## Summary

Fix file selection in the download code used by transfer and guest-invitation detail pages.

When a transfer ID and a file ID have the same numeric value, the generic `[data-id]` selector may select the transfer container instead of the file row. This causes encrypted downloads to fail before any request is made to `download.php`.

Fixes #2598

## Root cause

The transfer detail page can contain both:

```html
<div class="fs-transfer-detail" data-id="8">
```

and:

```html
<tr class="file" data-id="8" data-fileaead="...">
```

The download code used selectors such as:

```javascript
"[data-id='" + fileId + "']"
```

When both IDs are equal, the selected element may not contain the file encryption attributes. Reading `data-fileaead` then returns `undefined`, causing the download flow to stop with:

```text
Cannot read properties of undefined (reading 'length')
```

## Change

Scope the download-related selectors to file rows:

```diff
- "[data-id='" + fileId + "']"
+ ".file[data-id='" + fileId + "']"
```

This ensures that the encryption metadata is read from the intended file element.

## Validation

The correction has been validated with mandatory browser-side encryption enabled.

Tested successfully:

- individual encrypted download from transfer details;
- archive download from transfer details;
- download of a file submitted through a guest invitation;
- download using the public transfer link;
- transfer and file sharing the same numeric ID.

## Disclosure

The initial diagnosis and this pull request description were drafted with assistance from ChatGPT.

The proposed selector change has been deployed and validated in our production FileSender 3.11 instance.


YohanRodriguezMDM:
Closing this pull request because it was opened against the wrong base branch. A replacement pull request has been opened against development3.

PR #2812, opened 2026-08-15
Expose SAML IdP for per-IdP configuration profiles

tvilla-kvvcr:
### Summary

Expose the identity provider that authenticated a SAML user as the standard `idp` authentication attribute.

FileSender already supports selecting configuration overrides through `auth_config_regex_files`, including selection based on the `idp` attribute. The SimpleSAMLphp authentication adapter currently does not expose the authenticating IdP through this attribute.

This change retrieves the IdP entity ID from SimpleSAMLphp's `saml:sp:IdP` authentication data and exposes it as `attributes['idp']`.

This allows existing configuration such as:

```php
$config['auth_config_regex_files'] = [
    'idp' => [
        '^https://idp.example.org/entity$' => 'example-profile',
    ],
];
```

to work for SAML-authenticated users as well.

No deployment-specific behaviour is introduced: the mapping from IdP identifiers to configuration profiles remains entirely in the operator's local FileSender configuration.

### Tests

Added unit tests covering:

* exposure of the SimpleSAMLphp IdP entity ID as the `idp` attribute;
* selection of a matching configuration profile through `auth_config_regex_files`;
* retention of the default configuration for a different or missing IdP.

Standalone test result:

`OK (4 tests, 4 assertions)`

### DIFF
--- a/classes/auth/AuthSPSaml.class.php
+++ b/classes/auth/AuthSPSaml.class.php
@@ -98,6 +98,8 @@ class AuthSPSaml
             
             $attributes = array();
 
+            // Expose the authenticating IdP as a standard authentication attribute.
+            $attributes['idp'] = $ssp->getAuthData('saml:sp:IdP');
             
             // Wanted attributes
             foreach (array('uid', 'name', 'email') as $attr) {
--- a/unittests/config_tests_filesender.xml
+++ b/unittests/config_tests_filesender.xml
@@ -44,6 +44,7 @@
         <testsuite name="core" >
             <file>tests/dbs/DBsTest.php</file>              <!-- Success -->
             <file>tests/utilities/UtilitiesTest.php</file>  <!-- Success -->
+            <file>tests/auth/AuthSPSamlTest.php</file>
 <!--         <file>tests/auditlog/AuditlogTest.php</file>  -->  <!-- Failed -->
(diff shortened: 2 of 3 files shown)


monkeyiq:
Thank you for the PR! And a nice new test case file as well :)

This was included in the 3.x series here though it didn't get backported into the 2.0. I think that update in 3.x was part of the tenant stats patch so the specific part making the `idp` available was in a larger 3.x only PR. Sorry about the duplication of effort on this :/ My workflow remains to try to add items in `development` and merge those updates to `development3` from time to time. IIRC there was discussion of sun setting the 2.x some time around the end of year. For some features like GPG I added only to 3.x because they are heavily reliant on UI interaction. 

https://github.com/filesender/filesender/blob/68b1ebf3b77e13dd9c28ca13b3afeeddf1914c0b/classes/auth/AuthSPSaml.class.php#L102

Anyway, happy with the PR :) I might do an update for the config-saml-idp-test.php file to avoid writing it to the config/ directory and deleting it on each run. Maybe storing it in `./ci/config` and setting things up to allow the test suite to use that path as a fallback.

tvilla-kvvcr:
thank you very much. Let me know if i can further improve this. e.g. i noticed a `trim` in the 3.x code, that i didn't do.

happy to help. keep up the good work.

PR #2814, opened 2026-08-18
slight update to the test suite for the idp expose

monkeyiq:
This was offered and merged in
https://github.com/filesender/filesender/pull/2812

It seemed simple to just go ahead and make these touchups rather than comment on the PR. 

Changes

* Header block trimmed and CommonUnitTestCase is super
* Support for finding config-saml-idp-test.php from ci/configs if run in test env setting
* Moved setStaticProperty et al into super.


monkeyiq:
The ^M issue in CommonUnitTestCase.php complicates the PR a bit.

PR #2834, opened 2026-09-15
fix: guest voucher creation mode incorrectly requires file argument

WebSpider:
When trying to create a guest voucher, the startup arguments checker forces you to specify a file.
Since this is not needed for voucher creation, as the voucher holder should send the file, this logic is flawed and should be changed.


monkeyiq:
Thanks!

PR #2836, opened 2026-09-16
Refuse a download of an expired or closed transfer

### LINKED ISSUE #2835: Download of an expired transfer fails with a storage error instead of the expired page
When a transfer expires, `Transfer::close()` deletes the storage objects but keeps the `Transfers`, `Files` and `Recipients` rows for as long as `auditlog_lifetime`. The recipient token therefore still resolves, and `www/download.php` checks nothing after `Recipient::fromToken()` (lines 59 to 71), so the download reaches `readChunk()` and throws `StorageFilesystemCannotReadException`. The recipient gets a failed download instead of the transfer expired page.

We see this on S3 storage: of 16 file uids that failed to read in the last month, 10 were links opened after the transfer had expired or after the sender deleted it, in one case a month after the data was gone.

A status check where the transfer is resolved would cover it:

```php
if ($transfer->isExpired() || $transfer->status == TransferStatuses::CLOSED)
    throw new TransferPresumedExpiredException();
```

Smaller, related: `StorageCloudS3::readChunk()` logs its own message and drops the S3 exception (`classes/storage/StorageCloudS3.class.php:157`), so a deleted object and a genuine S3 error are indistinguishable in the log. Adding `$e->getMessage()` there would separate the two.

Line numbers are against `master` as of 16 September 2026.

syphernl:
Fixes #2835.

A download link used after the transfer expired, or after the sender deleted it, got a storage error instead of the expired transfer page. Closing a transfer deletes the file data but keeps the transfer and its recipient tokens for as long as `auditlog_lifetime`, so the token still resolves and the download runs into `StorageFilesystemCannotReadException`.

The check sits where both the token path and the owner path have resolved the transfer, so one guard covers both.

The second commit logs the S3 exception that `StorageCloudS3::readChunk()` was dropping, so a deleted object and a genuine S3 failure can be told apart in the log.

Seen on a production S3 deployment: of 16 file uids that failed to read in a month, 10 were links opened after expiry. Both files pass `php -l`.


monkeyiq:
hmm, I saw the bug report before checking the PRs. I added your suggestions in other PRs.

monkeyiq:
Bug is here https://github.com/filesender/filesender/issues/2835 and links to the PRs that basically contain the same and a little more to this PR :(

PR #2839, opened 2026-09-16
mail: stop before sending to an empty address

syphernl:
On a link transfer the sender never enters addresses, so `$recipient->email` is empty, and the download page still offers the "notify me when the download completes" checkbox to that downloader. `mail()` then writes an empty `To:`, sendmail finds no recipient and the message lands in `dead.letter`, which grows unnoticed.

Nothing downstream can recover from that, so this refuses the send and logs it instead. The transfer owner still gets the `files_downloaded` mail.

Running these changes in production since August.


monkeyiq (approved):

www/download.php
```diff
@@ -541,7 +541,12 @@ function manageOptions($ret, $transfer, $recipient, $recently_downloaded = false
         }
     }
 
-    if ($transfer->getOption(TransferOptions::ENABLE_RECIPIENT_EMAIL_DOWNLOAD_COMPLETE)) {
+    // $recipient->email is empty for a link transfer, where the sender never
+    // entered addresses. The download page still offers the "notify me when
+    // the download completes" checkbox to such a downloader, so this branch is
+    // reached with nowhere to send to. The owner is covered separately by the
+    // files_downloaded mail below, so skipping here loses no notification.
+    if ($transfer->getOption(TransferOptions::ENABLE_RECIPIENT_EMAIL_DOWNLOAD_COMPLETE) && $recipient->email) {
```
monkeyiq, line 549: I would probably move the && $recipeint->email into another nested if block. There is some chance of it being unseen at the end there. I'll shift it in a moment.

PR #2840, opened 2026-09-16
rest: release the session lock before running the handler

syphernl:
Parallel chunk uploads from terasender share one `PHPSESSID`, so every request serialized on the session file lock for as long as the handler ran. On our production instance that dilated p99 chunk latency from roughly 200ms to many seconds while a large upload was in flight.

The REST handlers do not write to the session, so this closes it for writing before dispatching.

Running these changes in production since August.


monkeyiq:
I see some REST code that is calling other code which updates the SESSION. I have created a limited version of this PR in https://github.com/filesender/filesender/pull/2849 which only activates for PUT to file routes. It is also behind a configuration option (with it enabled by default) in case some deployments have issues with it and need to quickly turn it off.

One example of later session use is `getSecurityToken()` called from a somewhat legacy method for whole file upload in a single request in POST for file. If you are using TeraSender then this method is not used so will not be noticed but other legacy configurations might still be using it so I am hesitant to close the session early for all REST.

Thank you for this PR! For a local setup as I have it has a substantial performance increase.

monkeyiq:
I will merge the limited version of this PR from https://github.com/filesender/filesender/pull/2849 and work can be done expanding that for other REST methods as they are verified to not update the SESSION.

PR #2841, opened 2026-09-16
saml: return getLoginURL() unchanged

syphernl:
`getLoginURL()` already returns a complete URL with `ReturnTo` in it. Wrapping it in `Utilities::http_build_query()` appends a second `?` and a duplicate `ReturnTo`. Strict MTAs score the result as spam, so the transfer notification mail that carries the link does not arrive. We hit this with Scaleway TEM.

Running these changes in production since May.


monkeyiq:
The `authentication_source` is passed as part of the URL to visit such as `.../saml/sp/login/default-sp` so that doesn't need to be there. This is inline with the update in this PR. 

I think the `Utilities::http_build_query` might have been added during a sweep of url hanlding and in this case it just makes things slightly incorrect here rather than doing anything good.

Issue #2794, opened 2026-07-28
Link-only mode UI improvements and expiry validation boundary issue


takusng:
## Environment

- FileSender: 3.11
- PHP-FPM
- Apache
- Link-only mode (email transfer disabled)

---

## Issue 1: Link-only mode does not clearly indicate that the upload is in link mode

When FileSender is configured to allow only **"Get a Link"**, the upload page hides the "Get a Link" radio button.

As a result, only the section title ("Transfer method") remains, making it unclear that the current upload is operating in link mode.

### Expected behavior

When only link transfers are enabled, the UI should clearly indicate that the upload is in **Link mode**, similar to how Email mode is presented.

For example:

- Display a "Link mode" heading instead of the generic "Transfer method" label, or
- Otherwise clearly indicate that the current transfer is a link transfer.

This would make the UI more intuitive while keeping the interface simple.

---

### Configuration

```php
$config['default_transfer_days_valid'] = 10;
$config['max_transfer_days_valid'] = 10;
```

Selecting the default expiry date (10 days) results in:

```
Invalid expiry date
```

The transfer cannot be created.

### Workaround

Changing only the maximum value avoids the error:

```php
$config['default_transfer_days_valid'] = 10;
$config['max_transfer_days_valid'] = 11;
```

Uploads then succeed.

---

## Issue 3: DatePicker uses `max_transfer_days_valid` as the selectable limit

The DatePicker is initialized as:

```javascript
filesender.ui.nodes.expires.datepicker({
    minDate: 1,
    maxDate: filesender.config.max_transfer_days_valid
});
```

Because of this, after applying the workaround:

```php
$config['default_transfer_days_valid'] = 10;
$config['max_transfer_days_valid'] = 11;
```

users are allowed to select an 11-day expiry, even though the intended policy is 10 days.

There is currently no way to:

- keep a 10-day expiry policy for users,
- avoid the expiry validation error,
- keep the DatePicker UI consistent.

---

## Suggested improvements

1. Fix the expiry validation so that

```php
$config['default_transfer_days_valid'] = 10;
$config['max_transfer_days_valid'] = 10;
```

works correctly.

2. Decouple the DatePicker limit from the server validation limit.

Instead of

```javascript
maxDate: filesender.config.max_transfer_days_valid
```

consider using

- `default_transfer_days_valid`, or
- the maximum value of `selectable_transfer_days_valid`.

This would allow administrators to configure the UI limit independently from the server-side validation limit.

---

[...]

PR #2844, opened 2026-09-16
upload: send expires as an integer and round the day count down

syphernl:
`Date.getTime()` returns milliseconds, so dividing by 1000 gives a float and the client sends `expires` as a decimal. `RestEndpointGuest` rejects a decimal with `|^[0-9]{1,32}$|` and falls back to `default_transfer_days_valid`, so a voucher quietly gets the default instead of the date that was picked. Both endpoints on this branch accept a dot today, but an epoch is an integer, so this floors it at the source.

The second change is separate and still visible: the day count beside the picker is computed with `ceil` over a diff that still carries the current time of day, so picking 30 days displays 31.

Running in production since May.


monkeyiq (approved):

monkeyiq:
I think this also relates to Issue 2 / 3 of https://github.com/filesender/filesender/issues/2794 and might fix that one as well.

monkeyiq:
Thank you for these PRs!

